PR #6, opened 2016-11-29
Remove LogTrait from AbstractIdentifier

ADmad:
Don't really see the need for it.


markstory:
Looks good.

burzum:
I'm Ok with removing it because it's not used right now. But I put it in there (and then forgot it) because I wanted to discuss if we want to provide logging similar to the database config, so that you can log the login attempts. Good for debugging, good for security.

markstory:
@burzum We can add it back when we have a solid use case :smile:

Issue #2, opened 2016-11-29
RFC - Change file organization 


ADmad:
Not really happy with the way files / namespace are organized. Would much prefer to have them similar to what's shown below:

```
src/
    Identifier/
        IdentifierCollection.php
        IdentifierInterface.php
        AbstractIdentifier.php
        ...
    Adapter/
        AdapterInterface.php
        AbstractAdapter.php
        FormAdapter.php
        ...
    PasswordHasher/
        PasswordHasherFactory.php
        PasswordHasherTrait.php
        AbstractPasswordHasher.php
        ...
    Result.php
    ResultInterface.php
    AuthenticationService.php
```

dereuromark:
Sounds reasonable to me.

dakota:
Sounds good to me

markstory:
Works for me too.

burzum:
The namespaces becomes `Auth\Adapter` then. I'm OK with the others but with do you want to rename the specific "Authenticators" to generic "Adapter". Adapter could be anything. Also, when this is going to go into the core in the future, how would you make the difference to authentication? The last time this was discussed was: `Auth\Authentication` and `Auth\Authorization`.

So if you want to rename the structure as shown in your comment, then we should rename the top level namespace from Auth to Authentication.

ADmad:
The top level `Auth` stands for authentication so `Auth\Authentication is redundant. I am fine with renaming the top level `Auth` to `Authentication` so that we get `Authentication\Adapter`.

ADmad:
@burzum Thanks for taking care of the renames. 

Just I would still prefer using `Adapter` prefix instead of `Authenticator` so `src/Adapter/FormAdapter.php` instead of `src/Adapter/FormAuthenticator.php`.  But if the majority prefers existing prefix then it's okay for me too.

burzum:
I would prefer Authenticator because it's not as generic as Adapter. I would also prefer to keep the folder name Authenticator vs Adapter. So lets people vote on this.

ADmad:
I have few more renaming suggestions :)

- `BasicAuthenticator` => `HttpBasicAuthenticator`
- `DigestAuthenticator` =>  `DigestHttpBasicAuthenticator`
- `FormAuthenticator` => `PostAuthenticator` : Not too sure about the new name but I have always disliked "Form" prefix.

dereuromark:
[...]

PR #9, opened 2016-11-30
#2 Namespace rename

burzum:


dereuromark:
:+1:

burzum:
No idea why GH has a merge conflict, worked fine locally without getting a conflict. Merged i manually and pushed to master.

dereuromark:
Shouldnt the whole plugin then be renamed from

    cakephp/middleware-authentication

to 

    cakephp/authentication

?

I mean Middleware is a specific pattern and implementation detail that does not need to be reflected in the name. How it is done internally should not matter to "what it does" which should be reflected in the name.

burzum:
Yes, I agree

PR #15, opened 2016-11-30
Clarify the need for existing apps to migrate the app properly

dereuromark:
This will otherwise make a few people trip over it ~~once 3.4 is out~~ with 3.0/3.1 or 3.2 apps (or partially upgraded from there to 3.4)

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -9,7 +9,7 @@
 Don't know what a middleware is? [Check the CakePHP documentation](http://book.cakephp.org/3.0/en/controllers/middleware.html) and additionally [read this.](https://philsturgeon.uk/php/2016/05/31/why-care-about-php-middleware/)
 
 This plugin is for CakePHP 3.4+. 
-If your application existed before (<= CakePHP 3.3), please make it is adjusted to leverage middleware as described in the [docs](http://book.cakephp.org/3.0/en/controllers/middleware.html#adding-the-new-http-stack-to-an-existing-application).
+If your application existed before (<= CakePHP 3.3), please make sure it is adjusted to leverage middleware as described in the [docs](http://book.cakephp.org/3.0/en/controllers/middleware.html#adding-the-new-http-stack-to-an-existing-application).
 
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -8,6 +8,9 @@
 
 Don't know what a middleware is? [Check the CakePHP documentation](http://book.cakephp.org/3.0/en/controllers/middleware.html) and additionally [read this.](https://philsturgeon.uk/php/2016/05/31/why-care-about-php-middleware/)
 
+This plugin is for CakePHP 3.4+. 
+If your application existed before (<= CakePHP 3.3), please make it is adjusted to leverage middleware as described in the [docs](http://book.cakephp.org/3.0/en/controllers/middleware.html#adding-the-new-http-stack-to-an-existing-application).
```
ADmad: "..make **sure** it is.."

dereuromark:
Actually, would it be 3.3+? There it was introduced.. thinking about it.

dereuromark:
Or is there any specific 3.4+ usage in the code that requires this constraint?

ADmad:
It should work with 3.3. I haven't had a chance to try this out in an actual app yet :)

ADmad:
I run the test suite again cake 3.3 at it was all green.

@burzum Should we bump down the cake version in composer.json to 3.3 ?

burzum:
It was 3.4 because I didn't expect it to work with 3.3 request object. :)

ADmad:
Okay, we can revisit the cake version constraint once we are happy with everything :)

ADmad:
Hmm while the tests pass with cakephp 3.3, in apps using cake 3.3 the controller gets instance of `Cake\Network\Request`, not the new class which implements `\Psr\Http\Message\ServerRequestInterface` so one won't be able to use `$request->getAttribute()`

burzum:
This is not good. So keep 3.next or 3.4?

dereuromark:
3.4 then it is.

PR #17, opened 2016-11-30
Adding token based authentication

burzum:


src/Authenticator/TokenAuthenticator.php
```diff
@@ -0,0 +1,74 @@
...
+    protected function _getToken(ServerRequestInterface $request)
+    {
+        $header = $this->config('header');
+        if (!empty($header)) {
+            return $request->getHeaderLine($header);
+        }
+
+        $param = $this->config('queryParam');
+        $queryParams = $request->getQueryParams();
+        if (!empty($queryParams[$param])) {
```
markstory: Should you isset() instead of empty()?

src/Identifier/TokenIdentifier.php
```diff
@@ -0,0 +1,95 @@
...
+     * @return false|EntityInterface
+     */
+    public function identify($data)
+    {
+        if (!isset($data['token'])) {
+            return false;
+        }
+
+        $tokenVerification = $this->config('tokenVerification');
+        if (is_callable($tokenVerification)) {
```
markstory: Should you make sure the callable isn't a string?

tests/TestCase/Authenticator/TokenAuthenticatorTest.php
```diff
@@ -0,0 +1,92 @@
...
+        );
+        $response = new Response('php://memory', 200, ['X-testing' => 'Yes']);
+
+        // Test without token
+        $tokenAuth = new TokenAuthenticator($identifiers, [
+            'queryParam' => 'token'
+        ]);
+        $result = $tokenAuth->authenticate($request, $response);
+        $this->assertInstanceOf('\Authentication\Result', $result);
+        $this->assertEquals(Result::FAILURE_OTHER, $result->getCode());
```
markstory: Should these cases be separate tests?

src/Authenticator/TokenAuthenticator.php
burzum: Isset will return true even if it's an empty string or false. Empty seems more useful to me in this case?

> Determine if a variable is set and is not NULL. 

http://php.net/manual/en/function.isset.php

src/Identifier/TokenIdentifier.php
burzum: Check the new commit.

tests/TestCase/Identifier/TokenIdentifierTest.php
```diff
@@ -40,6 +42,30 @@ public function testIdentify()
...
+
+                return false;
+            }
+        ]);
+
+        $result = $identifier->identify(['token' => 'not-larry']);
+        $this->assertFalse($result);
+
+        $result = $identifier->identify(['token' => 'larry']);
+        $this->assertInstanceOf('\Cake\Datasource\EntityInterface', $result);
```
burzum: Feel free to change it? :)

PR #20, opened 2016-12-02
Add underscore prefix to protected properties.

ADmad:
All other classes do that same so it's better to maintain consistency in the plugin code.

### DIFF
--- a/src/Result.php
+++ b/src/Result.php
@@ -57,14 +57,14 @@ class Result implements ResultInterface
      *
      * @var int
      */
-    protected $code;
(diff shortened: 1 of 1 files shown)


markstory:
Another option would be to remove _all_ the prefixes.

ADmad:
@markstory Sure is, but since i am lazy I chose the one with lesser effort :stuck_out_tongue:

burzum:
I prefer the underscores, they're a nice visual guide on what are you're dealing with.

PR #21, opened 2016-12-04
Implement a method to get the successful authenticator

burzum:


src/AuthenticationService.php
```diff
@@ -179,10 +186,24 @@ public function authenticate(ServerRequestInterface $request, ResponseInterface
...
 
+        $this->_successfulAuthenticator = null;
+
         return new Result(null, Result::FAILURE_IDENTITY_NOT_FOUND);
     }
+
+    /**
+     * Gets the successful authenticator instance if one was successful after calling authenticate
+     *
+     * @return \Authentication\Authenticator\\AuthenticateInterface
```
ADmad: Extra `\`, missing `|null`

burzum: Fixed.

ADmad:
This seems fine but since there's no way to access the auth service instance in controller how would one use this method?

This is what I propose. Instead of setting `authenticate` and `identity` attributes to the request, the middleware should set the auth service instance itself to request. Then one can do calls like:

- `$request->getAttribute('authService')->getAuthenticationProvider()`
- `$request->getAttribute('authService')->getResult()`
- `$request->getAttribute('authService')->getIdentity()`

(the last 2 methods would have to be added to class of course).

burzum:
These methods existed before @markstory made the suggestion to change that. ;) We could add them back but I would keep $request->getAttribute('identity').

markstory:
I suggested that we remove the authentication service as a request attribute as we didn't have much reason to have it there in the past. Outside of testing, in what scenarios would you need the authentication provider?

ADmad:
I also realized that while `SessionAuthenticator` reads from session there's no code in the plugin which actually writes the identity to session, am I right?

So do we want to leaving writing / deleting identify to / from session to user?

ADmad:
In my opinion getting rid of storage class has made providing the login / logout functionality difficult.

burzum:
> I also realized that while SessionAuthenticator reads from session there's no code in the plugin which actually writes the identity to session, am I right?

Yes, because any other authenticator would have to write to the session.

> So do we want to leaving writing / deleting identify to / from session to user?

Yes, why not? This is something that would happen in the controller layer and could be still done by a component. I don't think this should be the duty of the authentication service / "layer".

While I partially agree with you about the storage I don't like it at the same time. It just doesn't feel right.

ADmad:
> Yes, why not? This is something that would happen in the controller layer and could be still done by a component. I don't think this should be the duty of the authentication service / "layer".

I don't trust newbies to even handle setting the user entity to session under proper key themselves :stuck_out_tongue: 

> While I partially agree with you about the storage I don't like it at the same time. It just doesn't feel right.

Addition of storage class was a significant improvement to AuthComponent IMO. [ zend-authentication](https://zendframework.github.io/zend-authentication/storage/) also uses storage classes in similar way. So I don't think there's anything technically wrong in using them.

burzum:
> I don't trust newbies to even handle setting the user entity to session under proper key themselves 

Their problem. :) If they can't read the manual it's a lesson they'll have to learn. You can't make code idiot proof.

> Addition of storage class was a significant improvement to AuthComponent IMO. zend-authentication also uses storage classes in similar way. So I don't think there's anything technically wrong in using them.

Well, then lets add them back in ``\Authentication\Storage\`?

markstory:
If we're adding storage back in then should we also remove the `SessionAuthenticator`. It seems redundant if there is also going to be `SessionStorage`. Would it make sense to consider other storage options like a `JwtStorage` in addition to `MemoryStorage`?

I'm also not sure the StorageInterface needs the `redirectUrl` method anymore. We have recently changed AuthComponent to use query strings for redirect targets as it works better for multi-tab scenarios.

ADmad:
> If we're adding storage back in then should we also remove the SessionAuthenticator. It seems redundant if there is also going to be SessionStorage. 

I agree.

> Would it make sense to consider other storage options like a JwtStorage in addition to MemoryStorage?

It would. I do want to look into first class JWT support once things settle down.

> I'm also not sure the StorageInterface needs the redirectUrl method anymore.

Yup `redirectUrl()` would be removed.

burzum:
I've removed `redirectUrl()`.  Where do we want to add the storage objects back? Identity object? Result?

markstory:
> Can we merge this PR here?

Sure. Would you want to further discuss how the Storage objects would work in #25?

burzum:
Yes, makes sense to continue this in #25 :)

PR #23, opened 2016-12-04
Allow middlware constructor to accept service config array.

ADmad:
This avoids having to explicity create an authentication service instance
when setting up the middlware in application.

### DIFF
--- a/src/Middleware/AuthenticationMiddleware.php
+++ b/src/Middleware/AuthenticationMiddleware.php
@@ -30,10 +30,15 @@ class AuthenticationMiddleware
     /**
      * Constructor
      *
-     * @param \Auth\Authentication\AuthenticatorService $authenticationService Authentication service instance.
+     * @param \Auth\Authentication\AuthenticatorService|array $authenticationService
+     *   Authentication service instance or config array used to create service instance.
      */
-    public function __construct(AuthenticationService $authenticationService)
+    public function __construct($authenticationService)
     {
+        if (is_array($authenticationService)) {
+            $authenticationService = new AuthenticationService($authenticationService);
+        }
+
         $this->_authenticationService = $authenticationService;
     }
 
--- a/src/Middleware/AuthenticationMiddleware.php
+++ b/src/Middleware/AuthenticationMiddleware.php
@@ -30,8 +30,8 @@ class AuthenticationMiddleware
     /**
      * Constructor
      *
-     * @param \Auth\Authentication\AuthenticatorService|array $authenticationService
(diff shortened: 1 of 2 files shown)


burzum:
Hmm I think the idea was to have more types as args than magic arguments that can be virtually anything and then check for the entity because of future changes to php that will probably make the language more type strict? Why not add the array as optional 2nd arg?

src/Middleware/AuthenticationMiddleware.php
```diff
@@ -30,10 +30,15 @@ class AuthenticationMiddleware
     /**
      * Constructor
      *
-     * @param \Auth\Authentication\AuthenticatorService $authenticationService Authentication service instance.
+     * @param \Auth\Authentication\AuthenticatorService|array $authenticationService Authentication service
+     *   instance or config array used to create service instance.
      */
-    public function __construct(AuthenticationService $authenticationService)
+    public function __construct($authenticationService)
```
markstory: Should this be a factory method instead? `AuthenicationMiddleware::create($config)` ?

ADmad:
My intention was to simplify this

```php
        $service = new AuthenticationService([
            'identifiers' => [
                'Authentication.Orm' => [
                    'fields' => [
                        'username' => 'email',
                        'password' => 'password'
                    ]
                ]
            ],
            'authenticators' => [
                'Authentication.Form',
                'Authentication.Session'
            ]
        ]);

        $authentication = new AuthenticationMiddleware($service);

        $middleware->add($authentication);
```
to
```php
        $authentication = new AuthenticationMiddleware([
            'identifiers' => [
                'Authentication.Orm' => [
                    'fields' => [
                        'username' => 'email',
                        'password' => 'password'
                    ]
                ]
            ],
            'authenticators' => [
                'Authentication.Form',
                'Authentication.Session'
            ]
        ]);

        $middleware->add($authentication);
```

But if accepting config array as constructor argument is not preferred then this can be closed.

burzum:
You just save one single line in exchange for removing the strict type for the first argument. I don't see much benefit in doing this, IMO it's weakening the implementation and further I think we wanted to get away from the big config arrays and use object instead and fluid interfaces?

ADmad:
Ok i'll undo this and add setter methods.

burzum:
I think we should update the docs because there are already ways of loading them via API:

```php
$service = new AuthenticationService();
$service->identifiers()->load('Authentication.Orm');
$service->loadAuthenticator('Authentication.Form', []);
```

ADmad:
Okay then closing this.

PR #25, opened 2016-12-05
Adding the storage adapters back

burzum:
This is just adding the set of classes back, they'll still need to be integrated with the authenticators or any other place we want them to be.


src/Storage/MemoryStorage.php
```diff
@@ -0,0 +1,78 @@
...
+     */
+    public function delete()
+    {
+        $this->_user = null;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function redirectUrl($url = null)
```
markstory: I thought redirect URLs were in the query string now?

markstory:
What is the motiviation for having storage objects again? I thought the SessionAuthenticator was a way of not needing storage objects?

src/Storage/MemoryStorage.php
burzum: Read through the discussion in #21

burzum:
Read through the discussion in #21 @ADmad wants them back and I don't know a better way either. I'm OK if the developer would have to take care of dealing with the session. It's not part of the authentication IMO.

src/Storage/MemoryStorage.php
burzum: I can remove that if we decide to use them.

markstory:
If we're going to add Storage objects back in how do we plan on dealing the the following situation:

* An application supports both persistent web sessions for a HTML front-end, and stateless HTTP basic authentication for an API
* When logged into the HTML front-end, persistent sessions should be used, while the API should not use persistent storage.

This is the situation that led us to remove the storage in the first place. Our thinking at the time was that persistent storage was an *Authenticator* as once a user has been identified and put into storage, we can 'identify' them in th e future by their session cookie.

With the storage object approach I don't know how we link storage objects and their relevant authenticators yet.

If we were to instead *not* include Storage objects, the session data could be seeded automatically, when a user is identified by having authenticators like `SessionAuthenticator` implement an interface that `AuthenticationService` could use to recognize that they are able to persist data. A possible prototype would be

```php
interface PersistentAuthenticator
{
  public function persistIdentity($user);
  public function clearIdentity();
}
```

Authenticators implementing this interface could encapsulate their storage needs. We would likely also need a service level method to enable persistence so that API authentication doesn't end up creating sessions. Something like:

```php
$authentication->enablePersistence();
```

Might work.

ADmad:
> If we're going to add Storage objects back in how do we plan on dealing the the following situation:
>  An application supports both persistent web sessions for a HTML front-end, and stateless HTTP basic authentication for an API
> When logged into the HTML front-end, persistent sessions should be used, while the API should not use persistent storage.

Same way we do with AutheComponent, load only the authenticators required based on URL. This is where middleware pipelines would have been used.

> when a user is identified by having authenticators like SessionAuthenticator implement an interface that AuthenticationService could use to recognize that they are able to persist data.

We have discussed similar option when trying to fix session start issues for non persistent authenticators for AuthComponent too and we ended up using storage classes. Don't remember why we opted for the latter.

That said having a `PersistentAuthenticator` interface sounds good.

>  We would likely also need a service level method to enable persistence so that API authentication doesn't end up creating sessions. Something like: `$authentication->enablePersistence();`

If identity persistence is going to be handled by authenticator itself why you we need a switch in the service class?  After a successful authentication the service class would just check if authenticator implements `PersistentAuthenticator` and if so call it's `persistIdentity()` method.

For logout purposes we can add `AuthenticationService::clearIdentity()` which would go through the configured adapters and calls `clearIdentity()` for the one's which implement `PersistentAuthenticator`.

markstory:
@ADmad Do you see a need to keep Storage objects if identity persistence is in the Authenticators? If we were to put persistence in the Authenticators, then I would expect that `SessionAuthenticator`, and perhaps a `JwtTokenAuthenticator` would implement the `PersistentAuthenticator` interface. Right now I don't know of any other persistent authenticators that we would need.

I would like to avoid storage objects if possible as its another concept and piece that people have to understand and fit together.

burzum:
@markstory while I agree with you, but mostly because I would like to remove them as well, I disagree with having "yet another concept". We just hide this "concept" in the authenticators if we do what you propose. :)

How is JWT persistent by the way? AFAIR the token is in the header of the request, if you don't send send it along the request you won't get authenticated.

This lib here seems to be nice https://github.com/Spomky-Labs/jose and pretty complete if you look at the comparison table here https://jwt.io/#libraries.

markstory:
JWT could be persistent if you serialized the user data into the token. Whie its not generally done itnis possible. In this situation the JWT could be a source of user data similar to how SessionAuthenticator is.

By abstracting over the separation of storage and authentication adapter we provide one fewer class type to learn about and configure.

ADmad:
> Do you see a need to keep Storage objects if identity persistence is in the Authenticators? 

@markstory There's no need for storage objects if authenticators handle persistence.

burzum:
@markstory I know that. But this is not really persistence for me because you always have to send it from the client side. Any token could be persistence if you send data along with it then. :)

ADmad:
Does the second param of `persistIdentity()` have to be `Result`?  I would directly pass identity. Identity will only be persisted in case of successful authentication anyway so I don't see any other info from `Result` object being useful.

burzum:
I've made the changes you wanted and pushed to a new branch called `persistence`.

I think we should change the constructor of the service to take request and response and the config as third argument. This way we don't need to pass the request to every single method. But this *could* be problematic as well because the request is immutable, right?

ADmad:
@burzum Can you please make a PR for the new branch so it's easy to review. This PR can be closed and we can continue discussion on the new PR.

burzum:
@ADmad see https://github.com/cakephp/authentication/pull/27

markstory:
>But this is not really persistence for me because you always have to send it from the client side. Any token could be persistence if you send data along with it then.

Yes! That's a good thing in my opinion. Sessions are basically pointers to data on the server while a JWT/Cookie authenticator could container a serialized identity allowing stateless authentication for web 'sessions.

markstory:
>I'm thinking about making it a protected property and the 1st arg of the Authenticator constructor. SessionAuthenticator requires access to the session object, which you'll get only from the request for both methods. Your thoughts?

Passing arguments and not retaining state where we don't have to sounds like a benefit to me. If we have the parameters available I would prefer we use arguments over additional object state, especially given that the request object is an immutable and 'current' state is not as easily retained.

ADmad:
Given the immutability of request instance it's safer to pass it as argument

burzum:
Lets continue this in #27

PR #27, opened 2016-12-07
Started working on persistence for the identity

burzum:
I know it's missing doc blocks. But let's get the API done before writing doc blocks that get thrown away 42 times before it's final.

### DIFF
--- a/src/AuthenticationService.php
+++ b/src/AuthenticationService.php
@@ -220,9 +220,10 @@ class AuthenticationService
      * Clears the identity from authenticators that store them and the request
      *
      * @param \Psr\Http\Message\ServerRequestInterface $request The request.
+     * @param \Psr\Http\Message\ResponseInterface The response.
      * @return \Psr\Http\Message\ServerRequestInterface
      */
-    public function clearIdentity(ServerRequestInterface $request, $requestAttribute = 'identity')
+    public function clearIdentity(ServerRequestInterface $request, ResponseInterface $response)
     {
         foreach ($this->_authenticators as $authenticator) {
             if ($authenticator instanceof PersistenceInterface) {
@@ -230,11 +231,7 @@ class AuthenticationService
             }
         }
 
-        if (!empty($requestAttribute)) {
-            return $request->withoutAttribute($requestAttribute);
-        }
-
-        return $request;
+        return $request->withoutAttribute('identity');
     }
 
     /**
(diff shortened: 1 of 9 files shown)


src/Authenticator/SessionAuthenticator.php
```diff
@@ -72,4 +72,22 @@ public function authenticate(ServerRequestInterface $request, ResponseInterface
...
+        $request->getAttribute('session')->write($sessionKey, $identity);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function clearIdentity(ServerRequestInterface $request)
+    {
+        $sessionKey = $this->config('sessionKey');
+        $request->getAttribute('session')->delete($sessionKey);
```
markstory: I like how simple these are :+1:

burzum:
@markstory I've changed the middleware a little. It will set now the service to the request object instead of the result and made the service able to return the result via `getResult()`. This allows us to get the result and log the user out from the requests attribute through the service without adding more proxy methods or introducing more coupling between the identity and the service.

src/AuthenticationService.php
```diff
@@ -220,21 +220,18 @@ public function authenticate(ServerRequestInterface $request, ResponseInterface
...
      * @param \Psr\Http\Message\ServerRequestInterface $request The request.
+     * @param \Psr\Http\Message\ResponseInterface The response.
      * @return \Psr\Http\Message\ServerRequestInterface
      */
-    public function clearIdentity(ServerRequestInterface $request, $requestAttribute = 'identity')
+    public function clearIdentity(ServerRequestInterface $request, ResponseInterface $response)
     {
         foreach ($this->_authenticators as $authenticator) {
             if ($authenticator instanceof PersistenceInterface) {
                 $authenticator->clearIdentity($request);
```
markstory: Should the response be passed here as well? And perhaps be a return value?

burzum: Yes, forgot to restore that after I accidentally did a git reset... Added it back.

But about the return value: What do you want to return? The authenticators could return bool but what should the service return? The only meaningful thing would be an array of authenticator and status of the call Session => true for example to figure out which and if all of them cleared the identity.

markstory: If the authenticators need to remove a cookie, or add a header, they'd need to return a response as the parameter would be immutable.

PR #29, opened 2016-12-12
Passing the response

burzum:


src/AuthenticationService.php
```diff
@@ -227,7 +227,7 @@ public function clearIdentity(ServerRequestInterface $request, ResponseInterface
     {
         foreach ($this->_authenticators as $authenticator) {
             if ($authenticator instanceof PersistenceInterface) {
-                $authenticator->clearIdentity($request);
+                $authenticator->clearIdentity($request, $response);
```
lorenzo: You should assign the response back, since any changes done to it would be lost otherwise

burzum: @lorenzo we need the request as well, because we clear the identity attribute from it. Or will keep this the reference of the passed response?

```php
    public function clearIdentity(ServerRequestInterface $request, ResponseInterface $response)
    {
        foreach ($this->_authenticators as $authenticator) {
                $response = $authenticator->clearIdentity($request, $response);
            }
        }

        return $request->withoutAttribute('identity');
    }
```

lorenzo: If you expect the request to be changed, then you will also need to return it

src/AuthenticationService.php
```diff
@@ -227,11 +227,14 @@ public function clearIdentity(ServerRequestInterface $request, ResponseInterface
     {
```
dereuromark: Here it still seems to be `@return \Psr\Http\Message\ServerRequestInterface` instead of array.

burzum: Fixed and pushed to master. Updated the readme.md as well.

Issue #30, opened 2016-12-13
Authentication component? Needed? Features?


burzum:
This could be done as a convenience way to access the service in the request:

```php
class AuthenticationComponent extends Component {

    public function logout()
    {
        $controller = $this->_registry->getController();
        $authentication = $controller->request->getAttribute('authentication');
        if (!$authentication instanceOf AuthenticationService) {
            // Just return or throw exception?
            return;
        }

        $result = $authentication->clearIdentity(
            $controller->request,
            $controller->response
        );

        $controller->request = $result['request'];
        $controller->response = $result['response'];
    }
}
```

We could get the user as well but we can't write to the persistent authenticators through the service yet. So do we want to implement that at all? Also the component would give us a way to implement events like afterLogin we currently don't have but the old AuthComponent has.

If we implement methods to get and set the user data I would suggest to call them `getUser` and `setUser`.

Question: Is the changed instance of $controller->request and the response going to be available in the view layer as well? Or won't make this the way down to the view?

ADmad:
Yes a convenience component would be nice.

markstory:
> Question: Is the changed instance of $controller->request and the response going to be available in the view layer as well? Or won't make this the way down to the view?

View templates don't currently have access to the response, but they do have access to the request, which comes from the Controller.

markstory:
I think the component methods make sense to make migration easier, and the new auth system easier to use for new developers as well.

burzum:
Closing this in favour of #31

PR #31, opened 2016-12-15
Add AuthenticationComponent

ADmad:
Refs #30

@burzum You can add commits on top of this branch if you haven't already started your own locally :)

### DIFF
--- a/src/Controller/Component/AuthenticationComponent.php
+++ b/src/Controller/Component/AuthenticationComponent.php
@@ -12,6 +12,7 @@
  */
 namespace Authentication\Controller\Component;
 
+use Authentication\AuthenticationServiceInterface;
 use Cake\Controller\Component;
 use Cake\Event\EventDispatcherTrait;
 use Cake\Utility\Hash;
@@ -40,8 +41,8 @@ class AuthenticationComponent extends Component
         $controller = $this->_registry->getController();
         $this->_authentication = $controller->request->getAttribute('authentication');
 
-        if ($this->_authentication === null) {
-            throw new Exception('Authentication service instance not found in request.');
+        if (!$this->_authentication instanceof AuthenticationServiceInterface) {
+            throw new Exception('Authentication service does not implement ' . AuthenticationServiceInterface::class);
         }
 
         $this->eventManager($controller->eventManager());
--- a/src/Controller/Component/AuthenticationComponent.php
+++ b/src/Controller/Component/AuthenticationComponent.php
@@ -1,0 +1,89 @@
+<?php
+/**
+ * CakePHP(tm) : Rapid Development Framework (http://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (http://cakefoundation.org)
+ * Licensed under The MIT License
+ * For full copyright and license information, please see the LICENSE.txt
+ * Redistributions of files must retain the above copyright notice.
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (http://cakefoundation.org)
+ * @link          http://cakephp.org CakePHP(tm) Project
+ * @license       http://www.opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Authentication\Controller\Component;
+
+use Cake\Controller\Component;
+use Cake\Event\EventDispatcherTrait;
+use Exception;
+
+class AuthenticationComponent extends Component
+{
+
+    use EventDispatcherTrait;
+
+    /**
+     * Authentication service instance.
+     *
+     * @var \Authentication\AuthenticationService
+     */
+    protected $_authentication;
+
+    /**
+     * Initialize component.
+     *
+     * @param array $config The config data.
+     * @return void
+     */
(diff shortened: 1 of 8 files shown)


burzum:
@ADmad We should communicate if we started something or not to avoid unnecessary waste of time.

Are you going to add tests?

ADmad:
> We should communicate if we started something or not to avoid unnecessary waste of time.

Normally I would have but here I was totally prepared to thrash this PR if you had already started :)

> Are you going to add tests?

Would appreciate if you could, it might be a while until I can get around to it.

ADmad:
I was wondering how we should handle `setIdentity()`.  Since the identify is inside result should we get a new Result instance to the auth service instance or just update the `identity` attribute in request?

burzum:
You need to do both and we need to implement a proxy method that will pass the identity to all authenticators that implement the persistence interface to get it properly updated.

src/Controller/Component/AuthenticationComponent.php
```diff
@@ -0,0 +1,98 @@
...
+     *
+     * @param array $config The config data.
+     * @return void
+     */
+    public function initialize(array $config)
+    {
+        $controller = $this->_registry->getController();
+        $this->_authentication = $controller->request->getAttribute('authentication');
+
+        if ($this->_authentication === null) {
```
burzum: We should check for the instance not just if it's not null. This also brings up the question if we don't want to implement an interface for the service so people can change the service without breaking stuff.

ADmad: Right, this should be an instance check. Having a `AuthenticationServiceInterface` would be nice.

ADmad:
> You need to do both..

How about we don't set the `identity` attribute anymore in request object?  With `getIdentity()` provided by this component I don't see the need for it.

> ..we need to implement a proxy method that will pass the identity to all authenticators that implement the persistence interface to get it properly updated.

Thats sounds right, in which case we will need to add a `AuthenticationService::setIdentity()` method.

src/Controller/Component/AuthenticationComponent.php
burzum: Done. :)

src/Controller/Component/AuthenticationComponent.php
```diff
@@ -0,0 +1,99 @@
...
+     * @param array $config The config data.
+     * @return void
+     */
+    public function initialize(array $config)
+    {
+        $controller = $this->_registry->getController();
+        $this->_authentication = $controller->request->getAttribute('authentication');
+
+        if (!$this->_authentication instanceof AuthenticationServiceInterface) {
+            throw new Exception('Authentication service does not implement ' . AuthenticationServiceInterface::class);
```
ADmad: It would be better to show a separate message if authentication attribute isn't set in request at all.

src/Controller/Component/AuthenticationComponent.php
```diff
@@ -0,0 +1,99 @@
...
+     */
+    public function getResult()
+    {
+        return $this->_authentication->getResult();
+    }
+
+    /**
+     * Returns the identity used in the authentication attempt.
+     *
+     * @param string $path Path to return from the data.
```
dereuromark: |null then as well

src/Controller/Component/AuthenticationComponent.php
```diff
@@ -0,0 +1,99 @@
...
+
+class AuthenticationComponent extends Component
+{
+
+    use EventDispatcherTrait;
+
+    /**
+     * Authentication service instance.
+     *
+     * @var \Authentication\AuthenticationService
```
dereuromark: This should also be the interface then.

src/Controller/Component/AuthenticationComponent.php
burzum: True, done.

src/Controller/Component/AuthenticationComponent.php
dereuromark: Sry didnt want to Interrupt commits.

src/Controller/Component/AuthenticationComponent.php
```diff
@@ -0,0 +1,118 @@
...
+     * @param string|null $path Path to return from the data.
+     * @return mixed
+     */
+    public function getIdentity($path = null)
+    {
+        $controller = $this->_registry->getController();
+        $identity = $controller->request->getAttribute('identity');
+
+        if (is_string($path)) {
+            return Hash::get($identity, $path);
```
markstory: Nice that Entities implement `ArrayAccess` 😄

src/Controller/Component/AuthenticationComponent.php
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    /**
+     * Log a user out.
+     *
+     * Triggers the `Authentication.logout` event.
+     *
+     * @return void
+     */
+    public function logout()
```
markstory: Should the component also offer things like `loginRedirect`?

PR #33, opened 2016-12-26
Use Cake\Http\Response instead of Zend\Diactoros\Response

ADmad:


dereuromark:
Nice :)

markstory:
Looks good to me.

PR #41, opened 2017-01-10
Add AuthenticationService::loadIdentifier()

### LINKED ISSUE #40: Homogeneous api for identifiers and authenticators
Why don't identifiers and authenticators get loaded on a similar way? I mean, the use case it's pretty much the same: to load a class (optionally with its config) into a stack/set of elements. If you take a look into its load methods and are pretty similar...

Right now we have:

```php
        // Load identifiers
        $service->identifiers()->load('Authentication.Orm', [
            'fields' => [
                'username' => 'email',
                'password' => 'password'
            ]
        ]);

        // Load the authenticators, you want session first
        $service->loadAuthenticator('Authentication.Form');
```

I wonder why don't we have same way on both:

```php
        // Load identifiers
        $service->identifiers()->load('Authentication.Orm', [/*...config...*/]);

        // Load the authenticators, you want session first
        $service->authenticators()->load('Authentication.Session', [/*...config...*/]);
        $service->authenticators()->load('Authentication.Form'); // without config
```

or:

```php
        // Load identifiers
        $service->loadIdentifier('Authentication.Orm', [/*...config...*/]);

        // Load the authenticators, you want session first
        $service->loadAuthenticator('Authentication.Session', [/*...config...*/]);
        $service->loadAuthenticator('Authentication.Form'); // without config
```

I guess it's more a question than a bug, but I just noticed that and wanted to let you know. I think that will be easier to learn/remember an api when things are done in a "standard" way.

ADmad:
Closes #40

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -33,7 +33,7 @@ class Application extends BaseApplication
         $service = new AuthenticationService();
 
         // Load identifiers
-        $service->identifiers()->load('Authentication.Orm', [
+        $service->loadIdentifier('Authentication.Orm', [
             'fields' => [
                 'username' => 'email',
                 'password' => 'password'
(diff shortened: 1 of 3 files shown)


dereuromark:
The readme also should be adjusted then.

ADmad:
The readme updates are already included in the PR.

PR #42, opened 2017-01-15
After identify event

burzum:


burzum:
Has anyone an idea how to get the event properly tested? I've tried for some time but something always went wrong. The mocked eventManager() returned me always null. No matter if I mocked the controller objects eventManager() or the one of the component. If the code is hart to test it's easy to alter in other ways as well I think. So maybe somebody else has an idea how to refactor it or get the code tested?

markstory:
Have you considered a global listener that updates state on the test case that is then asserted.

PR #46, opened 2017-01-20
Pin cakephp libs to beta release instead of dev branch.

### LINKED ISSUE #45: Please change the dependency to a release
If you are releasing betas, please do change the dependencies to make use of beta versions.

https://github.com/cakephp/authentication/blob/1.0.0-beta1/composer.json#L12-L15

```
"cakephp/core": "dev-3.next as 3.4",
"cakephp/event": "dev-3.next as 3.4",
"cakephp/orm": "dev-3.next as 3.4",
"cakephp/utility": "dev-3.next as 3.4",
```

else 

```
{
    "require": {
        "cakephp/authentication":"~1.0"
    },
    "minimum-stability":"beta"
}
```

releases are not useful for it dumps 

```
composer install
Loading composer repositories with package information
Updating dependencies (including require-dev)
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - Installation request for cakephp/authentication ~1.0 -> satisfiable by cakephp/authentication[1.0.0-beta1].
    - cakephp/authentication 1.0.0-beta1 requires cakephp/core dev-3.next as 3.4 -> satisfiable by cakephp/core[dev-3.next] but these conflict with your requirements or minimum-stability.
```

The work around I know is use something like 

```
"cakephp/authentication":"dev-master",
"cakephp/core": "dev-3.next as 3.4",
"cakephp/event": "dev-3.next as 3.4",
"cakephp/orm": "dev-3.next as 3.4",
"cakephp/utility": "dev-3.next as 3.4",
```

but again then these tags are not useful in its own.

ADmad:
Closes #45

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -9,14 +9,14 @@
     ],
     "type": "cakephp-plugin",
     "require": {
-        "cakephp/core": "dev-3.next as 3.4",
-        "cakephp/event": "dev-3.next as 3.4",
-        "cakephp/orm": "dev-3.next as 3.4",
-        "cakephp/utility": "dev-3.next as 3.4",
+        "cakephp/core": "3.4.*@beta",
+        "cakephp/event": "3.4.*@beta",
+        "cakephp/orm": "3.4.*@beta",
+        "cakephp/utility": "3.4.*@beta",
         "psr/http-message": "~1.0"
     },
     "require-dev": {
-        "cakephp/cakephp": "dev-3.next as 3.4",
+        "cakephp/cakephp": "3.4.*@beta",
         "cakephp/cakephp-codesniffer": "^2.1",
         "phpunit/phpunit": "@stable"
     },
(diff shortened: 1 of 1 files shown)


ADmad:
I also want to remove `cakephp/orm` as dependency and move to "suggest".  Non cakephp users might want to use create their own authenticator using another ORM package. CakePHP users won't be affected since their app would include `cakephp/cakephp` and hence they already have our ORM.

Any objections to this?

harikt:
@ADmad I agree with you. There can also be people using only for oauth or jwt tokens. So orm is not needed itself. 

So +1 from my side :-) .

dereuromark:
@ADmad It should also be in require-dev then still, right? As the tests will need it.
But :+1:

harikt:
@dereuromark https://github.com/cakephp/authentication/pull/46/files#diff-b5d0ee8c97c7abd7e3fa29b9a27d1780R19 , cake itself is there . Now what ? :-) .

dereuromark:
Good point then ;)

Issue #48, opened 2017-01-22
Credential challenge headers not sent to client for HttpBasic and HttpDigest authenticators


ADmad:
While the authenticator for http basic and digest authenticator have `unauthenticated()` method for throwing `UnauthorizedException` with the challenge headers this method is never called by the authentication service.

Similarly for other stateless authenticator like `TokenAuthenticator` there's no mechanism to return a response with 401 code. So even if authentication fails the controller action is accessible.

markstory:
Perhaps the fix for this could be related to #47? If instead of using a Cake HTTP exception we could use a new state `Result::FAILURE_HARD_FAILURE/FAILURE_CHALLENGE_REQUIRED` to signal to the `AuthenticationService` that the current authenticator its unauthenticated() method called so it can prepare a `ChallengeException`. The middleware could handle this exception and use the headers/body in the exception to update the response and return it instead of calling the next middlware layer. This would be a pattern similar to how `Router` handles redirect routes.

ADmad:
Yeah fixing this would take care of #47 too. Your suggestion on how to handle this sounds good.

markstory:
I'll try to put together a solution.

markstory:
Closing as #50 is open.

Issue #47, opened 2017-01-22
Remove usage of Cake\Network\Exception\UnauthorizedException


ADmad:
Using `Cake\Network\Exception\UnauthorizedException` means we would have to add `cakephp/cakephp` as dependency which is undesirable. To avoid that the plugin should have it's own `UnauthorizedException`.

burzum:
@ADmad should this exception do something special or do we want to set some data to it? Or just extend \Exception and we're OK?

ADmad:
@burzum Check @markstory's comments on #48.

markstory:
Closing as #50 is open.

PR #50, opened 2017-01-25
Prototype of HTTP authentication challenges

markstory:
Use exceptions and make HTTP authentication failures prevent lower layers of middleware from being invoked until the HTTP authentication passes.

I've not updated digest authentication yet, and the test coverage is a bit sparse, but I'll update that once people are happy with the general approach.

Refs #47
Refs #48


src/Authenticator/HttpBasicAuthenticator.php
```diff
@@ -72,32 +74,28 @@ public function getUser(ServerRequestInterface $request)
...
+     * Create a challenge exception for basic auth challenge.
      *
      * @param \Psr\Http\Message\ServerRequestInterface $request A request object.
-     * @param \Psr\Http\Message\ResponseInterface $response A response object.
-     * @return void
-     * @throws \Cake\Network\Exception\UnauthorizedException
+     * @throws \Authentication\Authenticator\ChallengeException
      */
-    public function unauthenticated(ServerRequestInterface $request, ResponseInterface $response)
+    public function authenticationChallenge(ServerRequestInterface $request)
```
burzum: Is there really no case in which we might need to modify the response as well?

burzum:
Looks OK to me.

ADmad:
I have a concern about the name of exception and interface. Besides http basic and http digest auth for other stateless authenticators also a 401 response code should be returned albeit without any challenge headers. So maybe instead of word "Challenge" something more general can be used?

markstory:
@ADmad I'm open to ideas. I couldn't come up with something better.

ADmad:
How about `StatelessAuthenticatorInterface` (it is a big long) instead of `ChallengerInterface`?  All stateless authenticator would implement it. That would also remove the need for having `AuthenticateInterface::isStateless()` method. We could just check if an authenticator implements `StatelessAuthenticatorInterface`.

Also have a `UnauthenticatedException` in the plugin with stateless authenticators can throw. It could replace `ChallengeException` or `ChallengeException` can extend `UnauthenticatedException`.

markstory:
`UnauthenticatedException` sounds good to me. If the interface uses `Stateless` in the name, should we also use that interface to decide whether or not an authenticator is stateless and drop the `isStateless()` method?

ADmad:
> If the interface uses Stateless in the name, should we also use that interface to decide whether or not an authenticator is stateless and drop the isStateless() method?

We should IMO.

markstory:
@ADmad updated.

ADmad:
Looks good to me.

lorenzo:
Does this mean that only one authentication method can be tried as opposed to running through all methods until one matches?

markstory:
> Does this mean that only one authentication method can be tried as opposed to running through all methods until one matches?

Right now that is the case if you are using a stateless authenticator. One could configure the basic/digest auth as the 'last' authenticator though.

markstory:
Not really sure what is going on with the PHP5.6 builds.

burzum:
> Right now that is the case if you are using a stateless authenticator. One could configure the basic/digest auth as the 'last' authenticator though.

This should be added to the docs or people will run into this trap.

ADmad:
@markstory Looks like some phpunit issue.

markstory:
@burzum Another option is to try and work around it so the configuration order doesn't matter. Personally I think the documentation is the simpler of the two approaches for now.

lorenzo:
Does any of you have had a case  requiring multiple stateless auth mechanisms? I remember in particular that I tend to avoid having basic auth in the stack because if the automatic headers sending, as I had no control over when it was ok to send them.

ADmad:
> Does any of you have had a case requiring multiple stateless auth mechanisms?

I haven't.

markstory:
@lorenzo I have not either. I've used either Basic or Digest Auth but never both.

markstory:
Anything else I should fix up on this? I can update the documentation in the README if people think its necessary.

burzum:
@markstory I think we should update the docs.

markstory:
Docs updated 👍

PR #51, opened 2017-01-26
Jwt token

burzum:


src/Authenticator/JwtAuthenticator.php
```diff
@@ -0,0 +1,138 @@
...
+     * @inheritdoc
+     */
+    public function __construct(IdentifierCollection $identifiers, array $config = [])
+    {
+        parent::__construct($identifiers, $config);
+
+        if (empty($this->_config['salt'])) {
+            if (class_exists('\Cake\Utility\Security')) {
+                $this->getConfig('salt', \Cake\Utility\Security::salt());
+            } else {
```
dereuromark: If we invert and bail early we won't need an else here.

burzum: And it should be set not get. ;)

src/Authenticator/JwtAuthenticator.php
```diff
@@ -0,0 +1,138 @@
...
+use Firebase\JWT\JWT;
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
+use RuntimeException;
+use stdClass;
+
+class JwtAuthenticator extends TokenAuthenticator {
+
+    /**
+     * @inheritDoc
```
ADmad: IIRC apigen requires `{@inheritDoc}`

src/Authenticator/JwtAuthenticator.php
```diff
@@ -0,0 +1,138 @@
...
+     */
+    protected $_defaultConfig = [
+        'header' => 'Authorization',
+        'queryParam' => 'token',
+        'tokenPrefix' => 'bearer',
+        'allowedAlgs' => ['HS256'],
+        'entityClass' => Entity::class,
+        'returnPayload' => true,
+        'key' => null,
+        'salt' => null
```
ADmad: Having a `salt` key is unnecessary. Just having `key` is enough and it user does not provide any value for it then use value of `Security::salt()` as fallback if `Cake\Utility\Security` is available else throw an exception. The fallback check and throwing exception can be done in `_decodeToken()` itself.

src/Authenticator/JwtAuthenticator.php
```diff
@@ -0,0 +1,138 @@
...
+     * @inheritDoc
+     */
+    protected $_defaultConfig = [
+        'header' => 'Authorization',
+        'queryParam' => 'token',
+        'tokenPrefix' => 'bearer',
+        'allowedAlgs' => ['HS256'],
+        'entityClass' => Entity::class,
+        'returnPayload' => true,
+        'key' => null,
```
ADmad: Renaming this to `secret` or `secret_key` would be better. That's the name uses in other implementations I have seen.

src/Authenticator/JwtAuthenticator.php
```diff
@@ -0,0 +1,138 @@
...
+class JwtAuthenticator extends TokenAuthenticator {
+
+    /**
+     * @inheritDoc
+     */
+    protected $_defaultConfig = [
+        'header' => 'Authorization',
+        'queryParam' => 'token',
+        'tokenPrefix' => 'bearer',
+        'allowedAlgs' => ['HS256'],
```
ADmad: `Algs` is a silly short form I have used in my plugin :)  I think we should rename the key to just `algorithm`.

src/Authenticator/JwtAuthenticator.php
ADmad: `JWT::decode()` already throws exception if key is empty so we can even skip that check.

src/Authenticator/JwtAuthenticator.php
```diff
@@ -0,0 +1,138 @@
...
+/**
+ * CakePHP(tm) : Rapid Development Framework (http://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (http://cakefoundation.org)
+ * Licensed under The MIT License
+ * For full copyright and license information, please see the LICENSE.txt
+ * Redistributions of files must retain the above copyright notice.
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (http://cakefoundation.org)
+ * @link          http://cakephp.org CakePHP(tm) Project
+ * @since         4.0.0
```
ADmad: We don't have `@since` in other classes.

src/Authenticator/JwtAuthenticator.php
```diff
@@ -0,0 +1,138 @@
...
+ * @license       http://www.opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Authentication\Authenticator;
+
+use Authentication\Identifier\IdentifierCollection;
+use Authentication\Result;
+use Cake\ORM\Entity;
```
ADmad: This makes Cake's ORM a hard dependency. I don't think that's good for JWT authenticator.

src/Authenticator/JwtAuthenticator.php
```diff
@@ -0,0 +1,138 @@
...
+            if (class_exists('\Cake\Utility\Security')) {
+                $this->getConfig('salt', \Cake\Utility\Security::salt());
+            } else {
+                throw new RuntimeException('You must set the `salt` config key');
+            }
+        }
+    }
+
+    /**
+     * Authenticates the identity contained in a request. Will use the `config.userModel`, and `config.fields`
```
ADmad: Description needs to be updated.

src/Authenticator/JwtAuthenticator.php
```diff
@@ -0,0 +1,138 @@
...
+    /**
+     * Decode JWT token.
+     *
+     * @param string $token JWT token to decode.
+     * @return object|null The JWT's payload as a PHP object, null on failure.
+     */
+    protected function _decodeToken($token)
+    {
+        $config = $this->getConfig();
+        $token = str_ireplace($config['tokenPrefix'] . ' ', '', $token);
```
ADmad: I think stripping out the prefix should be in `TokenAuthenticator::_getToken()` itself. The `tokenPrefix` option could be added to `TokenAuthenticator` itself with default `null`.

src/Authenticator/JwtAuthenticator.php
burzum: Agreed, kept it mostly because you had it there. ;)

src/Authenticator/JwtAuthenticator.php
burzum: OK

src/Authenticator/JwtAuthenticator.php
burzum: I would stick to our exception because it's more specific in what is missing.

src/Authenticator/JwtAuthenticator.php
burzum: I always miss the {} for that.

src/Authenticator/JwtAuthenticator.php
burzum: The ORM is already in the required section of composer. We need at least EntityInterface because this has to be returned by the Result object when getIdentity() is called.

src/Authenticator/JwtAuthenticator.php
markstory: I get this now. Its removing the `Bearer ` prefix. `str_ireplace` will also replace any occurrences *in* the token as well. I don't think we want that.

ADmad: Theoretically it could, but I have never seen a token with space in it.

src/Authenticator/JwtAuthenticator.php
ADmad: > The ORM is already in the required section of composer.

No it's not. Check the discussion in #46.

>  We need at least EntityInterface because this has to be returned by the Result object when getIdentity() is called.

Why is the identity implementing `EntityInterface` a must?  Can't we just have it implement `ArrayAccess` ?  Have Cake's ORM as a hard dependency will dissuade non cake users from using the lib. I would prefer if this plugin was usable outside cakephp too.

src/Authenticator/JwtAuthenticator.php
```diff
@@ -0,0 +1,138 @@
...
+        return new Result($user, Result::SUCCESS);
+    }
+
+    /**
+     * Get payload data.
+     *
+     * @param \Psr\Http\Message\ServerRequestInterface|null $request Request to get authentication information from.
+     * @return object|null Payload object on success, null on failure
+     */
+    public function getPayload(ServerRequestInterface $request = null)
```
ADmad: It's so that one can get the payload after authentication has been completed. JWT can contain permission info too, so authorizers would have need access to the payload. Once payload is already decoded from the request, the request object is not required again.

composer.json
```diff
@@ -18,7 +18,11 @@
     "require-dev": {
         "cakephp/cakephp": "dev-3.next as 3.4",
         "cakephp/cakephp-codesniffer": "^2.1",
-        "phpunit/phpunit": "@stable"
+        "phpunit/phpunit": "@stable",
+        "firebase/php-jwt": "~4.0"
+    },
+    "suggest": {
+        "firebase/php-jwt": "~4.0"
```
ADmad: Yup, the right side should be something like `If you want to use the 'JwtAuthenticator'`.

src/Authenticator/JwtAuthenticator.php
burzum: @ADmad I'm with you, you need to delegate that question to @markstory who wanted entities.

src/Authenticator/JwtAuthenticator.php
markstory: True.

src/Authenticator/JwtAuthenticator.php
markstory: I had originally suggested that Identity data be an `EntityInterface` to reduce the number of interfaces we expose users to, and to streamline working with the ORM or other ORM-ish adapters.

I'm not sure `ArrayAccess` provides enough behavior to be the interface for an identity.

src/Authenticator/JwtAuthenticator.php
burzum: I haven't ever seen a token with words in it, except there is a language that is made of random characters :)

tests/TestCase/Authenticator/JwtAuthenticatorTest.php
```diff
@@ -0,0 +1,120 @@
...
+     * @return void
+     */
+    public function testAuthenticateViaQueryParamToken()
+    {
+        $this->request = ServerRequestFactory::fromGlobals(
+            ['REQUEST_URI' => '/'],
+            ['token' => $this->token]
+        );
+
+        $authenticator = new  JwtAuthenticator($this->identifiers, [
```
burzum: Fixed.

src/Authenticator/JwtAuthenticator.php
burzum: So what are we going to do about this? I haven't changed this yet.

src/Authenticator/JwtAuthenticator.php
burzum: Done.

src/Authenticator/JwtAuthenticator.php
```diff
@@ -0,0 +1,136 @@
...
+use RuntimeException;
+use stdClass;
+
+class JwtAuthenticator extends TokenAuthenticator
+{
+
+    /**
+     * {@inheritDoc}
+     */
+    protected $_defaultConfig = [
```
burzum: @markstory @ADmad how are we going to deal with this?

ADmad: We don't need to change anything. Looks like stickler is using PSR2 standard instead of CakePHP standard.

markstory: Yeah it will do that until you rebase with master.

src/Authenticator/JwtAuthenticator.php
burzum: @markstory @ADmad please see my previous comment.

burzum:
@ADmad I think the authenticator is OK as it is now. Checked it after the short chat we had in the dev channel yesterday but it's already working as discussed.

We don't do any checks for "sub" or other data and just provide the option to return the JWT payload in the result otherwise it gets passed to the identifiers and the result is returned.

ADmad:
@burzum Looks good to me in general. I would prefer not having the Cake ORM as a hard dependency by perhaps making the entity class configurable but that can be discussed later.

Could you just add a test for the JwtAuthenticator using OrmIdentifier to match value of "sub" with "id" and fetch related db record?

src/Authenticator/JwtAuthenticator.php
markstory: `EntityInterface` is in cakephp/datasource, not the ORM :smile:

burzum: @markstory No, that's right. I can't instantiate an entity interface. 😄

The authenticator offers the option to return the JWT token data as it is, without passing it to the identifiers. And the result needs to return an entity class. See the options.

The only problem that remains is how to deal with the "sub" (subject) part of the token. @ADmad  suggested to pass it to an identifier. According to the spec the subject in a token can be a JSON string. I understand it so that this can be in theory any data. Int, string, array... 

I've tried to re-use the OrmIdentifier and Admad suggested the TokenIdentifier. But this requires that the subject is always a single field because then the token identifier could be used to look up the id as if it where a token.

I think we need to find a better way to deal with the data and to transform it into a DB query.

markstory: Does the JWT need to contain a JSON version of the User? If `sub` was a scalar value that could be used to look up a record via the Identifiers then we wouldn't have to couple to the ORM here.

ADmad: > According to the spec the subject in a token can be a JSON string.

@burzum  Which specs are you referring to? This is what the [JWT specs rfc7519](https://tools.ietf.org/html/rfc7519#section-4.1.2) says about "sub":

> The "sub" value is a case-sensitive string containing a StringOrURI  value.

So value a "sub" is always a string. So one can just set "id" or "username" value to sub and then use the `TokenAuthenticator` to return matching user record.

burzum:
@ADmad I think I found a nice solution. Check the latest changes.

ADmad:
The `JwtSubjectIdentifier` looks good to me.

src/Authenticator/TokenAuthenticator.php
```diff
@@ -30,17 +30,65 @@ class TokenAuthenticator extends AbstractAuthenticator
...
+     * @param \Psr\Http\Message\ServerRequestInterface $request The request that contains login information.
+     * @param string $headerLine Header name
+     * @return string|null
+     */
+    protected function getTokenFromHeader(ServerRequestInterface $request, $headerLine)
+    {
+        if (!empty($headerLine)) {
+            $header = $request->getHeaderLine($headerLine);
+            if (!empty($header)) {
+                return $request->getHeaderLine($headerLine);
```
markstory: Why are you calling `getHeaderLine` again? Couldn't you do `return $header` here?

burzum: Fixing this now, guess it was a copy and paste or late night issue. :)

burzum:
@markstory added a middleware test for the JWT auth

src/Authenticator/JwtAuthenticator.php
```diff
@@ -32,7 +31,7 @@ class JwtAuthenticator extends TokenAuthenticator
         'queryParam' => 'token',
         'tokenPrefix' => 'bearer',
         'algorithms' => ['HS256'],
-        'entityClass' => Entity::class,
+        'entityClass' => '\Cake\ORM\Entity',
```
dereuromark: Dont think we need the leading slash then.

burzum: It doesn't hurt either? So whats the standard for this if stickler doesn't complain?

dereuromark: We use it without in core everywhere afaik.

src/Authenticator/JwtAuthenticator.php
```diff
@@ -0,0 +1,135 @@
...
+     */
+    public function __construct(IdentifierCollection $identifiers, array $config = [])
+    {
+        parent::__construct($identifiers, $config);
+
+        if (empty($this->_config['secretKey'])) {
+            if (!class_exists('\Cake\Utility\Security')) {
+                throw new RuntimeException('You must set the `secretKey` config key');
+            }
+            $this->setConfig('salt', \Cake\Utility\Security::salt());
```
ADmad: This should be 'secretKey' not 'salt'.

burzum: Fixed.

burzum:
@ADmad @markstory I'll resolve the merge conflict today or tomorrow and then merge it.

Do we want to tag `1.0.0-beta2` or do we want to go to `1.0.0-rc1`? I think it's feature complete, the Cookie adapter won't make it into 1.0. The only thing remaining that might change the API is the entity vs non entity approach we've discussed.

ADmad:
A lot has changed since beta1 so let's do beta2. Then after some actual testing with an app we can do RC1.

ADmad:
@burzum I took care rebasing the branch to resolve merge conflict.

Issue #49, opened 2017-01-22
public function identify($data)


dereuromark:
This method should not `@return false|EntityInterface` but instead `@return EntityInterface|null`.
It is a cleaner default value in general (despite the fact that this should be "bool", not "false", as false doesnt exist as type).
Also, this would be future compliant with 7.1+ "nullable" strictness, if we ever introduced that.

It is also more consistent with the rest of the framework or the direction it goes to, or where it should go to - either using a NullObject of some sorts or just returning null.
Using more than that as return type alternatives is usually a code smell - and a clear nogo even when returning objects.

markstory:
Seems like a reasonable change to me

dereuromark:
PR open.

PR #52, opened 2017-01-31
Make sure object alternative return value is null instead of bool.

dereuromark:
Closes https://github.com/cakephp/authentication/issues/49

We should run phpstan over it to make sure.

Awesome work so far, guys!


markstory:
There is [one failing test](https://travis-ci.org/cakephp/authentication/jobs/197052761)

dereuromark:
Thx, I also made sure travis doesnt use the global phpunit as that one is buggy.

dereuromark:
Darn I accidently pushed that to master, I reverted the test.
Strange, it didnt fail for me running locally, though.

dereuromark:
OK, I think I figured it out.

burzum:
:+I think it's OK.

PR #57, opened 2017-02-15
Update dependency versions.

ADmad:


composer.json
```diff
@@ -9,13 +9,14 @@
...
-        "cakephp/utility": "3.4.*@beta",
+        "cakephp/core": "~3.4",
+        "cakephp/datasource": "~3.4",
+        "cakephp/event": "~3.4",
+        "cakephp/utility": "~3.4",
         "psr/http-message": "~1.0"
     },
     "require-dev": {
-        "cakephp/cakephp": "3.4.*@beta",
+        "cakephp/cakephp": "~3.4",
```
dereuromark: I would Go with ^.
Projects can still lock it down more, but plugins should probably be a bit more open by default between minors.

ADmad: `~3.4` translates to `>=3.4 <4.0.0` so it's open enough.

dereuromark: Good point. As long as the last digit is not submitted, it will be the same as ^.

PR #58, opened 2017-02-15
Get result from auth attribute

jeremyharris:
The `authentication` attribute contains the AuthenticationService class, not the actual Result class.


markstory:
Thanks!

Issue #59, opened 2017-02-23
Adding a callback identifier?


burzum:
This would allow people to quickly come up with simple identifier solutions. For example I have a simple academic data uploader app and my friend is the only "user" and so I'm just checking against a password coming from the apps config. Could work nice for prototyping as well.

Edit: PR is up. :) https://github.com/cakephp/authentication/pull/60

PR #60, opened 2017-02-23
Adding a callback identifier

burzum:
This implements https://github.com/cakephp/authentication/issues/59

### DIFF
--- a/src/Identifier/CallbackIdentifier.php
+++ b/src/Identifier/CallbackIdentifier.php
@@ -1,0 +1,63 @@
+<?php
+/**
+ * CakePHP(tm) : Rapid Development Framework (http://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (http://cakefoundation.org)
+ * Licensed under The MIT License
+ * For full copyright and license information, please see the LICENSE.txt
+ * Redistributions of files must retain the above copyright notice.
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (http://cakefoundation.org)
+ * @link          http://cakephp.org CakePHP(tm) Project
+ * @license       http://www.opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Authentication\Identifier;
+
+use Cake\Datasource\EntityInterface;
+use Closure;
+use InvalidArgumentException;
+use RuntimeException;
+
+/**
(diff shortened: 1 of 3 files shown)


src/Identifier/CallbackIdentifier.php
```diff
@@ -0,0 +1,63 @@
...
+        if (!$callback instanceof Closure) {
+            throw new InvalidArgumentException(sprintf(
+                'The `callback` option is not a callable but `%s`',
+                gettype($callback)
+            ));
+        }
+
+        $result = $callback($data);
+
+        if (!$result instanceof EntityInterface && $result !== null) {
```
dereuromark: Should be inverted IMO, null check should come first.

burzum: Why?

dereuromark: The null check is way faster than an interface comparison, and the latter is not necessary if we can skip early because of null.

burzum: OK

src/Identifier/CallbackIdentifier.php
```diff
@@ -0,0 +1,63 @@
...
+     * Identify
+     *
+     * @param array $data Authentication credentials
+     * @return \Cake\Datasource\EntityInterface|null
+     */
+    public function identify($data)
+    {
+        $callback = $this->getConfig('callback');
+
+        if (!$callback instanceof Closure) {
```
markstory: We should probably allow other callable types. For example instance methods or static functions.

ADmad:
I pushed updated as per the feedback.

src/Identifier/CallbackIdentifier.php
```diff
@@ -0,0 +1,62 @@
...
+     * Identify
+     *
+     * @param array $data Authentication credentials
+     * @return \Cake\Datasource\EntityInterface|null
+     */
+    public function identify($data)
+    {
+        $callback = $this->getConfig('callback');
+
+        if (!is_callable($callback)) {
```
lorenzo: Can't we make this check in the constructor instead? or in the config setter?

burzum: @lorenzo https://github.com/cakephp/authentication/pull/61

Issue #64, opened 2017-03-01
ResponseInterface for persistIdentity()


steefaan:
It's necessary to have the `$response` available inside `persistIdentity()` of the `PersistenceInterface`, otherwise it's not possible to create a `CookieAuthenticator`.

https://github.com/cakephp/authentication/blob/master/src/Authenticator/PersistenceInterface.php#L27

burzum:
This is probably right and needs further investigation.

However, we'll hold no this until https://github.com/cakephp/cakephp/pull/10328 is done. See https://github.com/cakephp/authentication/issues/26 as well.

At least I don't have any intention in providing a BC compatible cookie authenticator at this moment, this plugin is targeting to be be integrated within the core with CakePHP 3.5. Right now cookie handling is tied to a component but *could* (need to check the code) be done outside of the component as well using the existing code.

PR #65, opened 2017-03-01
#64 Refactoring the PersistenceInterface

burzum:
See #64


dereuromark:
So setEntity() must then also return the array instead of `@return \Psr\Http\Message\ServerRequestInterface` it seems.

markstory:
Looks good to me.

PR #67, opened 2017-03-01
Fix a phpstan error

cleptric:
Fixes
`Authentication\PasswordHasher\WeakPasswordHasher::__construct() does not call parent constructor from Authentication\PasswordHasher\AbstractPasswordHasher.
`


dereuromark:
The order shouldn't matter IMO, or this is a phpstan issue. Sometimes it is needed to do sth before calling the parent :)

cleptric:
`Debugger::checkSecurityKeys` will trigger an error if the salt is `__SALT__`.

PR #70, opened 2017-03-03
Documentation updates

davidyell:
If you don't specify the fields, as they are in the example, the auth will fail as it won't detect the fields in the request.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -11,6 +11,12 @@ Don't know what a middleware is? [Check the CakePHP documentation](http://book.c
 This plugin is for CakePHP 3.4+.
 If your application existed before (<= CakePHP 3.3), please make sure it is adjusted to leverage middleware as described in the [docs](http://book.cakephp.org/3.0/en/controllers/middleware.html#adding-the-new-http-stack-to-an-existing-application).
 
+## What is it?
+This package is an Authentication middleware layer for authenticating requests coming into your application, using various authenticators and indentifiers.
+
+## What it is not
+The pacakge only provides authentication and not authorization. It can be used to find out who someone is, but not what they're allowed to do. 
+
 ## Quick Start
 
 Add the authentication service to the middleware. See the CakePHP [documentation](http://book.cakephp.org/3.0/en/controllers/middleware.html#) on how to use a middleware if you don't know what it is or how to work with it.
@@ -42,7 +48,7 @@ class Application extends BaseApplication
 
         // Load the authenticators, you want session first
         $service->loadAuthenticator('Authentication.Session', [
-            'sessionKey' => 'user'
+            'sessionKey' => 'Auth'
         ]);
         $service->loadAuthenticator('Authentication.Form', [
             'fields' => [
@@ -91,8 +97,15 @@ $service->loadIdentifier('Authentication.Orm', [
 ]);
 
 // Load the authenticators leaving Basic as the last one.
-$service->loadAuthenticator('Authentication.Session');
-$service->loadAuthenticator('Authentication.Form');
+$service->loadAuthenticator('Authentication.Session', [
+    'sessionKey' => 'Auth'
+]);
+$service->loadAuthenticator('Authentication.Form', [
+    'fields' => [
+        'username' => 'email',
+        'password' => 'password'
+    ]
+]);
 $service->loadAuthenticator('Authentication.HttpBasic');
 ```
 
@@ -142,6 +155,7 @@ The debug will show you an array like this:
 
 ### Differences
 
+* No authentication layer is provided, you will need to build this yourself.
 * There is no automatic checking of the session. To get the actual user data from the session you'll have to use the `SessionAuthenticator`. It will check the session if there is data in the configured session key and put it into the identity object.
 * The user data is no longer available through the AuthComponent but is accessible via a request attribute and encapsulated in an identity object: `$request->getAttribute('authentication')->getIdentity();`
 * The logic of the authentication process has been split into authenticators and identifiers. An authenticator will extract the credentials from the request, while identifiers verify the credentials and find the matching user.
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -11,6 +11,12 @@ Don't know what a middleware is? [Check the CakePHP documentation](http://book.c
 This plugin is for CakePHP 3.4+.
 If your application existed before (<= CakePHP 3.3), please make sure it is adjusted to leverage middleware as described in the [docs](http://book.cakephp.org/3.0/en/controllers/middleware.html#adding-the-new-http-stack-to-an-existing-application).
 
+## What is it?
+This package is an Authentication middleware layer for authenticating requests coming into your application, using various authenticators and indentifiers.
+
+## What it is not
+The pacakge only provides authentication and not authorization. It can be used to find out who someone is, but not what they're allowed to do.
```
dereuromark: typo pacakge

readme.md
```diff
@@ -135,6 +155,7 @@ The debug will show you an array like this:
 
 ### Differences
 
+* No authentication layer is provided, you will need to build this yourself.
```
ADmad: s/authentication/authorization

ADmad:
> If you don't specify the fields, as they are in the example, the auth will fail as it won't detect the fields in the request.

Well it would fail if you are not using defaults. The default value of `fields` is `['username' => 'username', 'password' => 'password']` and `SessionAuthenticator` does use `'sessionKey' => 'Auth'` as default.

readme.md
```diff
@@ -41,8 +47,15 @@ class Application extends BaseApplication
...
 
         // Load the authenticators, you want session first
-        $service->loadAuthenticator('Authentication.Session');
-        $service->loadAuthenticator('Authentication.Form');
+        $service->loadAuthenticator('Authentication.Session', [
+            'sessionKey' => 'Auth'
+        ]);
+        $service->loadAuthenticator('Authentication.Form', [
+            'fields' => [
+                'username' => 'email',
```
ADmad: Please change to default '`'username' => 'username'`. You can add a comment or another example with non default config.

markstory: I think this is a good example as it shows that you can customize the fields.

davidyell: It also makes all the examples in the readme consistant.

Perhaps
```php
'fields' => [
    'username' => 'email', // defaults to 'username'
    'password' => 'pswd', // defaults to 'password'
]
```

Then it can show both, or is that too busy?

readme.md
```diff
@@ -11,6 +11,12 @@ Don't know what a middleware is? [Check the CakePHP documentation](http://book.c
 This plugin is for CakePHP 3.4+.
 If your application existed before (<= CakePHP 3.3), please make sure it is adjusted to leverage middleware as described in the [docs](http://book.cakephp.org/3.0/en/controllers/middleware.html#adding-the-new-http-stack-to-an-existing-application).
 
+## What is it?
+This package is an Authentication middleware layer for authenticating requests coming into your application, using various authenticators and indentifiers.
+
+## What it is not
+The package only provides authentication and not authorization. It can be used to find out who someone is, but not what they're allowed to do. 
+
```
burzum: Do we really need this? I mean authentication is obviously not authorization?

readme.md
```diff
@@ -84,8 +97,15 @@ $service->loadIdentifier('Authentication.Orm', [
 ]);
 
 // Load the authenticators leaving Basic as the last one.
-$service->loadAuthenticator('Authentication.Session');
-$service->loadAuthenticator('Authentication.Form');
+$service->loadAuthenticator('Authentication.Session', [
+    'sessionKey' => 'Auth'
```
burzum: You are repeating the default value here. It is not necessary to set this unless you care about what key it is. You should either add that info or remove the config here.

readme.md
burzum: IMO there is no need to mention that. It's obvious and was mentioned before as well.

readme.md
ADmad: It seems many users think this is a drop in replacement for core auth component. So it's better to add this clarification.

readme.md
```diff
@@ -208,11 +229,18 @@ $service->loadIdentifier('Authentication.Orm', [
...
+$service->loadAuthenticator('Authentication.Form', [
+    'fields' => [
+        'username' => 'email',
+        'password' => 'password'
+    ]
+]);
 ```
 
-While this seems to be a little more to write, the benefit is a greater flexibility and better [separation of concerns](https://en.wikipedia.org/wiki/Separation_of_concerns).
+While this seems to be a little more to write, the benefit is a greater flexibility and better [separation of concerns](https://en.wikipedia.org/wiki/Separation_of_concerns). Plus, with this separation, it gives you much greater flexibility in developing your own authorization layer.
```
burzum: Besides this being the third time mentioning authorization (I would remove it here as well), remove at least the `Plus,`.

readme.md
```diff
@@ -208,11 +229,18 @@ $service->loadIdentifier('Authentication.Orm', [
...
```
burzum: Besides this being the third time mentioning authorization (I would remove it here as well), remove at least the `Plus,`.

readme.md
dereuromark: 'password' => 'pwd' 👍

readme.md
davidyell: I don't see any harm in driving home one of the key aspects of the package. Plus if someone scans the readme, or is just looking at one section, repeating yourself could be valuable.

readme.md
davidyell: Agreed, no harm in outlining your intended scope for the package upfront. My intention here was to try and drive home the separation aspect, as that was the thing which tripped me up initially.

readme.md
davidyell: My intention here was to try and impress upon the user that they can develop their own authorisation layer. I didn't want to go into too much detail, so I tried to summarise.

readme.md
davidyell: Yep, I thought it was worth showing the configuration key in the documentation example, so that people know they can customise it. Perhaps it's worth following the other examples, and using a custom value with a comment for the default?

davidyell:
As there is some discussion around how much information should be included about the separation of concerns between authentication and authorization. What is the consensus?

If we can agree on how to outline the scope of the package in a nice descriptive way, then I can make the updates to the pull request.

From my perspective, I do concede that it's probably not worth mentioning throughout the readme, but I would definitely keep the outline of the intended scope, just to be clear to new users. I'm also a big fan of preceding readmes with some explanation about why the package exists and what it's for, before geting into how to use it. That way if someone is looking for a package, they can get a reasonable grasp of what the package does early in the readme.

ADmad:
@davidyell Another PR  #73 has also been opened clarifying that this plugin only does authentication under the "Differences" heading. So repeating the fact over and over is not necessary.

readme.md
burzum: @davidyell if you want to help coming up with a good authorization solution check the CakePHP issues, [there is already a ticket](https://github.com/cakephp/cakephp/issues/9835) for it.

You might want to take a look at [this](https://laravel.com/docs/5.4/authorization), [this](https://github.com/CanCanCommunity/cancancan) and [this](https://en.wikipedia.org/wiki/Authorization). Spending a good amount of planning the architecture and API is the key here. I have this on my agenda but before I'll have to finish the cookies and the cookie authenticator. **After** that I'm going to come back to authorization.

davidyell:
Cool.

PR #72, opened 2017-03-03
Allow OrmIdentifer to fetch record using any one of the specified use…

### LINKED ISSUE #71: Allow OrmIdentifier to identify using any one field from a provided list.
A common use case is allowing users to login using either their username or email. It would be nice if `OrmIdentifier` could cater to this scenario out of the box.  One way would be allow the `'username'` option to take an array like `['username', 'email']` and the code then craft the db query accordingly.

ADmad:
…rname fields.

Closes #71

### DIFF
--- a/src/Identifier/OrmIdentifier.php
+++ b/src/Identifier/OrmIdentifier.php
@@ -106,19 +106,7 @@ class OrmIdentifier extends AbstractIdentifier
         $config = $this->_config;
         $table = TableRegistry::get($config['userModel']);
 
-        if (is_array($config['fields']['username'])) {
-            $conditions = [];
-            foreach ($config['fields']['username'] as $field) {
-                $conditions[$table->aliasField($field)] = $identifier;
-            }
-            $conditions = [
-                'OR' => $conditions
-            ];
-        } else {
-            $conditions = [$table->aliasField($config['fields']['username']) => $identifier];
-        }
-
-        $options = compact('conditions');
+        $options = ['conditions' => $this->_buildConditions($identifier, $table)];
 
         $finder = $config['finder'];
         if (is_array($finder)) {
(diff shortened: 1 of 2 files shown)


burzum:
I was told to not use more than 60 chars in the first line of a commit ;)

tests/TestCase/Identifier/OrmIdentifierTest.php
```diff
@@ -60,6 +60,31 @@ public function testIdentify()
...
+            'username' => 'larry',
+            'password' => 'password'
+        ]);
+        $this->assertInstanceOf('\Cake\Datasource\EntityInterface', $result);
+
+        $result = $identifier->identify([
+            'username' => 3,
+            'password' => 'password'
+        ]);
+        $this->assertInstanceOf('\Cake\Datasource\EntityInterface', $result);
```
burzum: `'\Cake\Datasource\EntityInterface'` vs `EntityInterface::class`?

src/Identifier/OrmIdentifier.php
```diff
@@ -96,17 +98,27 @@ protected function _findUser($username, $password = null)
...
+                $conditions[$table->aliasField($field)] = $identifier;
+            }
+            $conditions = [
+                'OR' => $conditions
+            ];
+        } else {
+            $conditions = [$table->aliasField($config['fields']['username']) => $identifier];
+        }
+
+        $options = compact('conditions');
```
burzum: Can't we put this whole piece of added code into a new method? Something like `buildOptions()`? Makes it more easy to override the method. And will probably make the code rating more happy and going up.

ADmad: We have other  protected methods too using the underscore prefix. So if we were to remove the underscore here we should be consistent and remove it from all methods.

burzum: I would prefer if we keep them, the prefixing is pretty helpful IMO.

So do we want to remove them all or not? If we want to remove them all this should be done probably in another PR.

src/Identifier/OrmIdentifier.php
```diff
@@ -115,9 +115,35 @@ protected function _query($username)
...
+            }
+            $conditions = [
+                'OR' => $conditions
+            ];
+        } else {
+            $conditions = [$table->aliasField($usernameFields) => $identifier];
+        }
+
+        return $conditions;
+    }
```
burzum: 👍

dereuromark:
Looks good.

markstory:
Looks good to me.

PR #76, opened 2017-03-05
Add missing pass of parameter

cleptric:
Add the pass of the `$response` parameter, when calling `persistIdentity()` on a `PersistenceInterface` object.


ADmad:
A test for the fix would be nice.

cleptric:
@ADmad Just realised that :)

cleptric:
Turns out that this was already covered but the test wasn't setup properly, as `$service->loadAuthenticators();` was missing.
We should discuss further changes in the reaming ones in an other PR or issue.

burzum:
Related issue https://github.com/cakephp/authentication/issues/77

PR #78, opened 2017-03-05
More tests for the JWT Adapter & a new exception

burzum:


src/AuthenticationService.php
```diff
@@ -205,11 +206,18 @@ protected function _getAuthenticatorClass($name, $config)
      * @param \Psr\Http\Message\ResponseInterface $response The response.
      * @return \Authentication\ResultInterface A result object. If none of
      * the adapters was a success the last failed result is returned.
+     * @throws RuntimeException Throws a runtime exception when no authenticators are loaded.
```
cleptric: Should be `\RuntimeException`

burzum: Fixed.

PR #79, opened 2017-03-05
Make error messages more consistent.

markstory:
* Use plural forms where they should be used.
* Use backticks instead of " for classnames.


burzum:
Looks good.

burzum:
@markstory this codecov is anoying... Why is it red here!?

markstory:
@burzum no idea, I think it didn't like that coverage didn't improve.

burzum:
@markstory It's pretty annoying because it always looks like it failed and you need to check the CI checks. :(

Issue #77, opened 2017-03-05
AuthenticationComponent::logout


cleptric:
The `AuthenticationComponent::logout()` currently transforms the request and response into an array, which will break afterwards, if you call `$this->redirect()` in a controller.

burzum:
Closing this because we have a PR open for it.

burzum:
No? How does it break? AuthenticationComponent::logout() returns `void|string|array`. Could you please add a test case that demonstrates the issue?

```php
    public function logout()
    {
        $controller = $this->_registry->getController();
        $result = $this->_authentication->clearIdentity(
            $controller->request,
            $controller->response
        );

        $controller->request = $result['request'];
        $controller->response = $result['response'];

        $logoutRedirect = $this->getConfig('logoutRedirect');
        if ($logoutRedirect !== false) {
            return Router::normalize($logoutRedirect);
        }
    }
```

cleptric:
The Problem is the return value of `SessionAuthenticator::clearIdentity()`.
It returns an array with the keys `request` and `response`.
Inside `AuthenticationService:.clearIdentity()`, the whole return value is set as the response, not just the response object inside the return value.

cleptric:
PR is open.

PR #80, opened 2017-03-05
Fix AuthenticationService::clearIdentity

cleptric:
See #77 
The complete return of `$authenticator->clearIdentity($request, $response);` was used as the response, instead of just using the `response` key of the return value.
Also added the missing `$this->service->loadAuthenticators();` to the component test.


markstory:
Looks good to me.

PR #102, opened 2017-03-23
Bump up phpstan level and fix reported errors.

ADmad:


src/Result.php
```diff
@@ -80,7 +80,7 @@ class Result implements ResultInterface
      * @param int $code Error code.
      * @param array $messages Messages.
      */
-    public function __construct($identity, $code, array $messages = [])
+    public function __construct($identity, int $code, array $messages = [])
```
robertpustulka: Are we on  PHP 7?

ADmad: Oops :stuck_out_tongue:

burzum:
Looks OK

PR #103, opened 2017-03-23
PasswordHasherTrait refactoring.

robertpustulka:
I need to mock it in a test :)

### DIFF
--- a/src/Identifier/OrmIdentifier.php
+++ b/src/Identifier/OrmIdentifier.php
@@ -2,6 +2,8 @@
 namespace Authentication\Identifier;
 
 use Authentication\PasswordHasher\DefaultPasswordHasher;
+use Authentication\PasswordHasher\PasswordHasherFactory;
+use Authentication\PasswordHasher\PasswordHasherInterface;
 use Authentication\PasswordHasher\PasswordHasherTrait;
 use Cake\ORM\Locator\LocatorAwareTrait;
 
@@ -50,6 +52,20 @@ class OrmIdentifier extends AbstractIdentifier
     ];
 
     /**
+     * {@inheritDoc}
+     */
+    public function __construct(array $config = [])
+    {
+        parent::__construct($config);
+
+        $passwordHasher = $this->getConfig('passwordHasher');
+        if (!$passwordHasher instanceof PasswordHasherInterface) {
+            $passwordHasher = PasswordHasherFactory::build($passwordHasher);
+        }
+        $this->_passwordHasher = $passwordHasher;
+    }
+
+    /**
      * Identify
      *
      * @param array $data Authentication credentials
@@ -91,7 +107,7 @@ class OrmIdentifier extends AbstractIdentifier
         }
 
         if ($password !== null) {
-            $hasher = $this->passwordHasher();
+            $hasher = $this->getPasswordHasher();
             $hashedPassword = $result->get($this->_config['fields']['password']);
             if (!$hasher->check($password, $hashedPassword)) {
                 return null;
--- a/src/Identifier/OrmIdentifier.php
+++ b/src/Identifier/OrmIdentifier.php
@@ -24,7 +24,7 @@ class OrmIdentifier extends AbstractIdentifier
 
     use LocatorAwareTrait;
     use PasswordHasherTrait {
-        getPasswordHasher as protected traitGetPasswordHasher;
+        getPasswordHasher as protected _getPasswordHasher;
     }
 
     /**
@@ -63,7 +63,7 @@ class OrmIdentifier extends AbstractIdentifier
             if ($passwordHasher !== null) {
                 $passwordHasher = PasswordHasherFactory::build($passwordHasher);
             } else {
-                $passwordHasher = $this->traitGetPasswordHasher();
+                $passwordHasher = $this->_getPasswordHasher();
             }
             $this->_passwordHasher = $passwordHasher;
         }
--- a/src/Identifier/OrmIdentifier.php
+++ b/src/Identifier/OrmIdentifier.php
@@ -24,7 +24,7 @@ class OrmIdentifier extends AbstractIdentifier
 
     use LocatorAwareTrait;
     use PasswordHasherTrait {
-        getPasswordHasher as private traitGetPasswordHasher;
+        getPasswordHasher as protected traitGetPasswordHasher;
     }
 
     /**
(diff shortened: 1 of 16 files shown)


src/PasswordHasher/PasswordHasherTrait.php
```diff
@@ -33,8 +33,11 @@ public function passwordHasher()
         }
 
         $passwordHasher = $this->_config['passwordHasher'];
+        if (!$passwordHasher instanceof PasswordHasherInterface) {
+            $passwordHasher = PasswordHasherFactory::build($passwordHasher);
```
burzum: Isnt the factory taking 2 args? class and config?

The fallback hasher needs config for example. I would enforce passing an object or check if an array was passed that can be used for class and options

And while being there - how about adding a getter and setter and deprecate this method? Or directly remove it. :)

robertpustulka: @burzum One argument: https://github.com/cakephp/authentication/blob/master/src/PasswordHasher/PasswordHasherFactory.php#L34

I'm working on a getter/setter right now :)

ADmad:
I don't really like this early creation of password hasher instance. Password hasher is only required during login so creating it's instance in the constructor when it won't be required for most of the requests is a bit wasteful.

robertpustulka:
Tests are there and password hasher is created only when needed.

src/Identifier/OrmIdentifier.php
```diff
@@ -23,7 +23,9 @@ class OrmIdentifier extends AbstractIdentifier
 {
 
     use LocatorAwareTrait;
-    use PasswordHasherTrait;
+    use PasswordHasherTrait {
+        getPasswordHasher as private traitGetPasswordHasher;
```
dereuromark: Wouldnt protected suffice? This way it could be reused from extension points if needed.

src/Identifier/OrmIdentifier.php
```diff
@@ -46,10 +48,30 @@ class OrmIdentifier extends AbstractIdentifier
...
+     * @return \Authentication\PasswordHasher\PasswordHasherInterface Password hasher instance.
+     */
+    public function getPasswordHasher()
+    {
+        if ($this->_passwordHasher === null) {
+            $passwordHasher = $this->getConfig('passwordHasher');
+            if ($passwordHasher !== null) {
+                $passwordHasher = PasswordHasherFactory::build($passwordHasher);
+            } else {
+                $passwordHasher = $this->traitGetPasswordHasher();
```
dereuromark: Thats one of the reasons I really liked the underscore

    _getPasswordHasher()

;)

But well, thhis way is OK as well I guess.

robertpustulka: I usually do `method as private _method` in such cases, but I'm not sure if this is a proper PSR-way to do that :smiley:

dereuromark: I still don't see why people think `_` violates PSR, it specifically says it is OK to use them if applied with proper reasoning, which CakePHP does :) It is perfectly fine and logical to do so IMO.

burzum: I've been told to remove it as well but I really would prefer to keep it except somebody tells me how I get phpstorm to visually indicate the visibility for me...

PSR my ass. 💩

tests/TestCase/PasswordHasher/PasswordHasherTraitTest.php
```diff
@@ -0,0 +1,55 @@
...
+ * CakePHP(tm) : Rapid Development Framework (http://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (http://cakefoundation.org)
+ *
+ * Licensed under The MIT License
+ * For full copyright and license information, please see the LICENSE.txt
+ * Redistributions of files must retain the above copyright notice.
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (http://cakefoundation.org)
+ * @link          http://cakephp.org CakePHP(tm) Project
+ * @since         3.0.0
```
ADmad: This line should be removed.

dereuromark: Do we really need all that "since tag" stuff? Or could we just say those are not needed and can be sniffer removed since the history is git-versioned anyway... This is a repeating issue afaik.

robertpustulka: I'm removing all of them.

dereuromark: Just asking - lets wait on more feedback before removing maybe ;)

robertpustulka: There are mixed tags 3.0.0 and 2.0.0 - these should be changed to 1.0.0 anyway or removed.

ADmad: We don't need this tags for this plugin, which is why I said the line should be removed.

src/PasswordHasher/PasswordHasherFactory.php
```diff
@@ -49,7 +49,7 @@ public static function build($passwordHasher)
 
         $hasher = new $className($config);
         if (!($hasher instanceof PasswordHasherInterface)) {
-            throw new RuntimeException('Password hasher must extend AbstractPasswordHasher class.');
+            throw new RuntimeException('Password hasher must implement PasswordHasherInterface.');
```
burzum: The exception is untested.

robertpustulka: @burzum Test has been added.

I'm wondering if we should allow `App\Authentication\PasswordHasher` namespace for password hashers? Currently only `App\PasswordHasher` is discovered. In fact the same applies for other components: `Authenticators` and `Identifiers`. I think it would be much cleaner if all these authentication related classes could reside under `App\Authentication` namespace.

burzum: Hashers aren't strictly related to authentication? You'll need them when creating user credentials as well. This can happen via API service, shell, or "normal" form POST by an user.

Are the hashers in any way coupled with the identifiers? I don't think so. So IMHO it's OK if they stay in their own namespace as I don't see them being a "subordinate" of authentication.

Issue #104, opened 2017-03-24
Authentiator and Identifier collection/registry upgrade.


robertpustulka:
Currently loading identifiers slightly differs from loading authenticators.

While `loadAuthenticator()` behaves more CakePHP-way accepting `className` option, `loadIdentifier()` treats a first argument as a class name, and accepts `alias` option. I think this should be refactored to behave the same way as core registries do.

I've been also thinking about some way to add Authenticator and Identifiers instances. Currently these can only be loaded from config. There's also no way to unset an object. Another thing is that identifiers have a collection but authenticators are stored in an array in `AuthenticationService`. Maybe both types of objects should share the same API for managing collections?

Example:
```php
$service->loadAuthenticator('Form');
//or
$service->authenticators()->load('Form');
//or
$authenticator = new FormAuthenticator();
$service->authenticators()->set('Form', $authenticator);
//and then at some point:
$service->authenticators()->unset('Form');
```

## TODO:

- [x] Rename `AuthenticateInterface` to `AuthenticatorInterface` (#105)
- [x] Create common collection for authenticators and identifiers
- [x] Implement subject specific collections
- [x] Add `AuthenticationServiceInterface::loadIdentifier()`

robertpustulka:
I'm also having doubts about `AuthenticationServiceInterface::loadAuthenticator()`. Should authentication services be forced to implement this method? It is needed in default `AuthenticationService` but some implementations wouldn't need it, for example:

```php
class SimpleService implements AuthenticationServiceInterface
{
    public function __construct(AuthenticateInterface $authenticator) {
        ...
    }
}
```

Another thing - is there a reason behind the interface being called `AuthenticateInterface` while all the authenticators use "authenticator" naming convention and reside under `Authenticator` namespace?

dereuromark:
@robertpustulka Sounds like the interface should get a rename here for consistency.

robertpustulka:
Can we use core `ObjectRegistry` for collections? `cakephp/core` is required by this plugin anyway.

burzum:
@robertpustulka can we close this as the PR's have been merged?

robertpustulka:
@burzum There's only a matter of dropping AuthenticationServiceInterface::loadAuthenticator() on todo list. What do you think about it? Should we force this implementation? I think this method is only related to our approach in AuthenticationService. Someone just may want to implement simple service without any authenticators or configuration needed.

burzum:
[...]

PR #107, opened 2017-03-24
Collections refactoring

robertpustulka:
See #104 

I think it was worth implementing :)


src/AbstractCollection.php
```diff
@@ -0,0 +1,73 @@
...
+     */
+    public function isEmpty()
+    {
+        return empty($this->_loaded);
+    }
+
+    /**
+     * Returns iterator.
+     *
+     * @return ArrayIterator
```
dereuromark: Should be FQCN

src/AbstractCollection.php
```diff
@@ -0,0 +1,73 @@
...
+
+use ArrayIterator;
+use Cake\Core\InstanceConfigTrait;
+use Cake\Core\ObjectRegistry;
+use IteratorAggregate;
+
+/**
+ * Description of AbstractCollection
+ *
+ * @author Robert Pustułka <r.pustulka@robotusers.com>
```
ADmad: This too needs to be removed.

robertpustulka:
Is it OK to be merged?

Issue #97, opened 2017-03-23
Config key 'identify' not set in SessionAuthenticator


littleylv:
[https://github.com/cakephp/authentication/blob/master/src/Authenticator/SessionAuthenticator.php#L58](https://github.com/cakephp/authentication/blob/master/src/Authenticator/SessionAuthenticator.php#L58)
Key 'identify' is not set in $_defaultConfig.
Also, in docs [https://github.com/cakephp/authentication/blob/master/docs/Authenticators.md#session](https://github.com/cakephp/authentication/blob/master/docs/Authenticators.md#session)

markstory:
What problems dies this omission create?

burzum:
There is no need to set it. If it's not present the check in L58 will simply not return true and the session is not checked against the DB record.

littleylv:
It won't cause any problems if it's not set.
But shouldn't we set it a default value?
At least mention it in the docs so developers will know SessionAuthenticator has a key 'identify'

burzum:
@littleylv is that OK?

> * **identify**: Set this key with a value of bool `true` to enable checking the session credentials against the identifiers.

ADmad:
It would still be nice to initialize the key in default config for those who prefer reading code rather than docs :)

littleylv:
@burzum As @ADmad  said, I was reading `SessionAuthenticator.php` at that time and discovered the missing config key.  😃

ADmad:
@littleylv Would you like to make a PR adding the kwy to default config?

PR #108, opened 2017-03-24
Add default config 'identify' to SessionAuthenticator

littleylv:
#97


dereuromark:
Wouldnt this go into https://github.com/cakephp/authentication/blob/master/src/Authenticator/AbstractAuthenticator.php#L32-L35 ?

dereuromark:
Or a _defaultConfig attribute at least inside the SessionAuthenticator class?

littleylv:
I didn't see any other authenticator use this key except session

dereuromark:
I see, merging will not work otherwise with the property.

Issue #86, opened 2017-03-06
Add a LDAP identifier?


burzum:
It would a *very* easy to turn [this plugin](https://github.com/QueenCityCodeFactory/LDAP) into an identifier. Let's do it?

I just don't have a LDAP server I could test against it. I have no idea how to test the function calls either. A possibility would be a small OOP wrapper around the modules functions and mock the whole LDAP connection object.

ADmad:
I too don't uave any experience with LDAP. Having adapter for it would be surely nice.

burzum:
I'll ask the author of the plugin if he would like to contribute to this project. :)

burzum:
I've slightly updated my old code I've started 1-2 month ago and pushed it to https://github.com/cakephp/authentication/tree/ldap-identifier

@ceeram and @cleptric use LDAP at work and said they can review and complete the work on this. Thanks guys. 👍

cleptric:
I was able login against a LDAP Server: http://www.forumsys.com/tutorials/integration-how-to/ldap/online-ldap-test-server/
That said, I'm not happy at all with the current implementation.
The handling of the ldap result is pretty fragile. The config array is also getting pretty big.
We should also add a new `OrmTrait`, which can handle the db lookup for the `TokenIdentifier` and the `LdapIdentifier`.

cleptric:
Hmm, on a second thought, I'm not quite sure if we want to lookup anything in the db while using the `LdapIdentifier`. I have to adjust the return of the identify method and create an entity on the fly, then.

burzum:
@cleptric as I understand LDAP it should provide all the info for the user already. You log in using LDAP, done. Sure you could link it with an account inside your local application but that is out of scope of the identifier - IMO. If somebody wants to do that he can link the accounts after login / registration or extend the identifier.

Also I've seen you're overriding $_defaultConfig. Please note that we're not using the MergeVarTrait that would merge the arrays of defined properties. So you're completely overriding the defaults.

cleptric:
@burzum LDAP doesn't provide you with any user data, if you're not searching for it. A ldap_bind is basically all is needed to authenticate a user against a LDAP server (I learned that yesterday 😉). If the bind is successful, you get `true` else `false`. So I dropped all the search code and the result juggling.
You're right about that an ORM lookup isn't necessary. _If_ the `LdapIdentifier` would do an ORM lookup, I just don't wanted to copy the code from the `TokenIdentifier`. This was the idea behind adding a trait.

[...]

PR #109, opened 2017-03-27
Ldap identifier

cleptric:
Implements #86
Does anyone has a better idea how to test this?


src/Identifier/LdapIdentifier.php
```diff
@@ -0,0 +1,186 @@
...
+
+/**
+ * LDAP Identifier
+ *
+ * Identifies authentication credentials using LDAP.
+ *
+ * ```
+ *  new LdapIdentifier([
+ *       'host' => 'ldap.example.com',
+ *       'bindDN' => function($username) {
```
dereuromark: What does  DN mean?

src/Identifier/LdapIdentifier.php
```diff
@@ -0,0 +1,186 @@
...
+ *       ]
+ *  ]);
+ * ```
+ *
+ * @link https://github.com/QueenCityCodeFactory/LDAP
+ */
+class LdapIdentifier extends AbstractIdentifier
+{
+
+    use LdapOopTrait;
```
robertpustulka: Is there a reason there is a trait used?

src/Identifier/LdapIdentifier.php
burzum: It's one of the plenty abbreviations of LDAP. I don't know what exactly it means either.

src/Identifier/LdapIdentifier.php
burzum: It's explained in the docs of the trait. :)

src/Identifier/LdapIdentifier.php
cleptric: `The LDAP API references an LDAP object by its distinguished name (DN). A DN is a sequence of relative distinguished names (RDN) connected by commas.`

tests/TestCase/Identifier/LdapIdentifierTest.php
```diff
@@ -0,0 +1,126 @@
...
+namespace Authentication\Test\TestCase\Identifier;
+
+use Authentication\Identifier\LdapIdentifier as BaseLdapIdentifier;
+use Authentication\Test\TestCase\AuthenticationTestCase as TestCase;
+use Cake\Datasource\EntityInterface;
+
+/**
+ * Overwrite all LdapOopTrait methods to enable tests
+ */
+class LdapIdentifier extends BaseLdapIdentifier
```
robertpustulka: Same as above. It makes little sense implementing an OOP wrapper as a trait and then using this trait in a mock class for tests.
This wrapper could be implemented as a regular class and injected as a dependency to the identifier or build from config array.

src/Identifier/LdapIdentifier.php
robertpustulka: My point is that's not very OOP. We could have a class and reuse it as a dependency.

tests/TestCase/Identifier/LdapIdentifierTest.php
burzum: You'll need an interface besides the class then as well.

src/Identifier/LdapIdentifier.php
burzum: @robertpustulka I've done a concept and pushed it to this branch.

src/Identifier/LdapIdentifier.php
```diff
@@ -0,0 +1,243 @@
...
+     * @throws \RuntimeException
+     * @return void
+     */
+    protected function _buildLdapObject()
+    {
+        if (empty($this->_config['ldapClass'])) {
+            $this->setConfig('ldapClass', Ldap::class);
+        }
+
+        if (is_string($this->_config['ldapClass'])) {
```
robertpustulka: How about using CakePHP convention with `className` config?

For example:
```php
'ldap' => Ldap::class //by default
//or
'ldap' => [
    'className' => Ldap::class,
    ... //other options
]
```

I think that's more intuitive as this pattern is commonly used.

src/Identifier/Ldap.php
```diff
@@ -0,0 +1,109 @@
...
+ */
+class Ldap implements LdapInterface
+{
+
+    /**
+     * LDAP Object
+     *
+     * @var object
+     */
+    protected $connection;
```
ADmad: All other protected member in the plugin have underscore prefix.

.travis.yml
```diff
@@ -25,6 +25,9 @@ matrix:
     - php: 7.1
       env: PHPSTAN=1 DEFAULT=0
 
+before_install:
+  - if [[ $DEFAULT = 1 && $TRAVIS_PHP_VERSION = 7.0 ]]; then echo 'extension = ldap.so' >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini; fi
```
antograssiot: `- if [[ $DEFAULT = 1  ]]; then echo 'extension = ldap.so' >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini; fi` should be better as I guess you also need them for 7.1 and 5.6

src/Identifier/LdapIdentifier.php
burzum: You'll have to implement a locator method then and use App::class() or whatever the name was. I'm in a call right now :)

tests/TestCase/Identifier/LdapIdentifierTest.php
```diff
@@ -82,6 +82,24 @@ public function testIdentify()
...
+     * @expectedExceptionMessage Could not build the LDAP connection object.
+     */
+    public function testWrongLdapObject()
+    {
+        $identifier = new LdapIdentifier([
+            'host' => 'ldap.example.com',
+            'bindDN' => function () {
+                return 'dc=example,dc=com';
+            },
+            'ldapClass' => 1
```
dereuromark: Why a magic number? Wouldn't a string `Invalid\ClassName` be more speaking?

cleptric: A string would be handeld by the `is_string` condition, an unknown class would raise an Error and an anonymous class is only available in php 7+

dereuromark: OK, but how realistic is this test then? Providing an invalid class name as string sounds to me like the only actual testable thing. Otherwise we would be testing all kind of crazyness as arguments everywere :)

src/Identifier/LdapIdentifier.php
burzum: I guess people working with LDAP know these terms, but it wouldn't hurt to add this to the docs / doc block(s)?

src/Identifier/LdapInterface.php
```diff
@@ -0,0 +1,66 @@
...
+ * Copyright (c) Cake Software Foundation, Inc. (http://cakefoundation.org)
+ * Licensed under The MIT License
+ * For full copyright and license information, please see the LICENSE.txt
+ * Redistributions of files must retain the above copyright notice.
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (http://cakefoundation.org)
+ * @link          http://cakephp.org CakePHP(tm) Project
+ * @license       http://www.opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Authentication\Identifier;
```
robertpustulka: I'm not sure this is a right place for LDAP related code.

src/Identifier/LdapIdentifier.php
robertpustulka: Also a getter and setter would be nice to mock it easily in tests.

src/Identifier/LdapInterface.php
burzum: Create a new NS: `namespace Authentication\Identifier\Backend`

robertpustulka: Or maybe `Authentication\Identifier\Adapter`?

src/Identifier/LdapIdentifier.php
markstory: If you're trying to allow dependency injection for testing, why not pass the object|null instead of a class name? Right now if a different implementation can't have any constructor arguments which can be a bit limiting and doesn't play nice with PHPUnit mocks.

cleptric:
I could raise the coverage quite a bit. As the `Ldap` class is just a simple OO wrapper and testing it would require a lot of effort (I'm not completely sure if it's even possible) the missing coverage there should be ok.

burzum:
If nobody else knows how to further increase the test coverage for the OOP wrapper of the LDAP adapter I'll merge this end of next week.

src/Identifier/LdapIdentifier.php
robertpustulka: @cleptric can we do as @markstory suggested? 

Having `ldapClass` now is a bit clumsy in testing and not in line with conventions used throughout the framework, where we accept a class name string or config array with `className` key in builders (see object registries for example). As I don't think we need array support here as implementations of `LdapInterface` may differ, we could allow for objects to be injected in this config. Thus we should rename this option to `ldap` instead of `ldapClass` to better reflect this can accept both types of arguments.

In other worlds we could have:
```php
[
    'ldap' => Ldap::class //by default
    //or
    'ldap' => $ldapAdapter //custom implementation
    ...
]
```

That would probably require tests refactoring in order to use `LdapInterface` mock instead of a custom class.

src/Identifier/LdapIdentifier.php
```diff
@@ -121,42 +122,9 @@ protected function _checkLdapConfig()
...
-     * Constructs the LDAP object and sets it to the property
-     *
-     * @throws \RuntimeException
-     * @return void
-     */
-    protected function _buildLdapObject()
-    {
-        if (empty($this->_config['ldapClass'])) {
-            $this->setConfig('ldapClass', Ldap::class);
+        if (!($this->_config['ldap'] instanceof LdapInterface)) {
```
robertpustulka: Isn't that still a class name?

Instanceof check will fail probably.

burzum:
Let's merge #113 first and then see if this here passes the tests and merge it then.

src/Identifier/LdapIdentifier.php
```diff
@@ -0,0 +1,240 @@
...
+
+    /**
+     * Checks if the php LDAP extension is loaded
+     *
+     * @throws \RuntimeException
+     * @return void
+     */
+    protected function _checkLdapExtension()
+    {
+        if (!extension_loaded('ldap')) {
```
robertpustulka: This check should be done in LDAP adapter. If someone wanted to use their own extension-independent adapter, this would fail.

robertpustulka:
I'm also wondering if `Backend` namespace is ok for that. We decided to have seperate namespace for identifier resolvers/locators/datasource. Maybe we should just rename this to `Ldap`?

- `Identifier\Backend` would become `Identifier\Ldap`
- `Identifier\Backend\LdapInterface` could become `Identifier\Ldap\AdapterInterface`
` `Identifier\Backend\Ldap` could become `Identifier\Ldap\ExtensionAdapter` so it would be clear that this implementation uses php extension.

src/Identifier/LdapIdentifier.php
```diff
@@ -0,0 +1,240 @@
...
+            $config['host'],
+            $config['port']
+        );
+
+        if (isset($config['options']) && is_array($config['options'])) {
+            foreach ($config['options'] as $option => $value) {
+                $this->_ldap->setOption($option, $value);
+            }
+        } else {
+            $this->_ldap->setOption(LDAP_OPT_NETWORK_TIMEOUT, 5);
```
robertpustulka: I'm afraid these constants couple LdapIdentifier to extension-based implementation.

phpunit.xml.dist
```diff
@@ -50,6 +50,9 @@
 		</blacklist>
 		<whitelist processUncoveredFilesFromWhitelist="true">
 			<directory suffix=".php">./src</directory>
+                        <exclude>
```
ADmad: Indentation is off.

src/Identifier/LdapIdentifier.php
cleptric: We can remove this. You can set it via the options array in the config.

robertpustulka: There are more LDAP_* constants used in this class.

robertpustulka:
To be honest I don't like this `LdapInterface` implementation. `LdapIdentifier` and the interface seem to be so tightly coupled to the extension-based implementation that I think there's no point in abstracting the OOP wrapper. Either we simplify the interface (leave maybe `connect()`, `bind()` and `unbind()` methods) and move all of the extension-related code to the adapter (like constants, error handling, extension loaded check etc) or remove the interface and leave `Ldap` as a simple internal wrapper class just so we can test it easily.

src/Identifier/LdapIdentifier.php
```diff
@@ -134,11 +117,11 @@ protected function _buildLdapObject()
         $ldap = $this->_config['ldap'];
 
         if (is_string($ldap)) {
-            $class = App::className($ldap, 'Identifier/Backend');
+            $class = App::className($ldap, 'Identifier/Ldap');
```
robertpustulka: It should have 'Adapter' type so 'Authenticate.Extension' would work.

phpunit.xml.dist
```diff
@@ -50,6 +50,9 @@
 		</blacklist>
 		<whitelist processUncoveredFilesFromWhitelist="true">
 			<directory suffix=".php">./src</directory>
+			<exclude>
+				<file>./src/Identifier/Backend/Ldap.php</file>
```
robertpustulka: This need to be changed to the new ExtensionAdapter class.

src/Identifier/LdapIdentifier.php
```diff
@@ -0,0 +1,216 @@
...
+     * Try to bind the given user to the LDAP server
+     *
+     * @param string $username The username
+     * @param string $password The password
+     * @return \Cake\Datasource\EntityInterface|null
+     */
+    protected function _bindUser($username, $password)
+    {
+        // Turn all LDAP errors into exceptions
+        set_error_handler(
```
robertpustulka: Should we move this error handler to the ExtensionAdapter as it's specific to this implementation? Custom adapters could use regular exceptions.

tests/TestCase/Identifier/LdapIdentifierTest.php
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    /**
+     * testIdentify
+     *
+     * @return void
+     */
+    public function testIdentify()
+    {
+        $ldap = $this->createMock(ExtensionAdapter::class);
```
robertpustulka: An interface should be mocked instead of a class.

tests/TestCase/Identifier/LdapIdentifierTest.php
```diff
@@ -0,0 +1,190 @@
...
+{
+
+    /**
+     * {@inheritDoc}
+     */
+    public function setUp()
+    {
+        parent::setUp();
+
+        $this->skipIf(!extension_loaded('ldap'), 'LDAP extension is not installed');
```
robertpustulka: This test should not depend on an extension?

robertpustulka:
This looks great now @cleptric! :+1:

src/Identifier/LdapIdentifier.php
cleptric: I guess we have to wrap each ldap_* call in a new error handler then, as `set_error_handler` is applied globally?

robertpustulka: Seems like a good idea :+1:

robertpustulka: BTW what's the cost of swapping error handlers? We should probably make sure it wouldn't hurt the performance.

cleptric:
Wow, we have a ✅  build now

Issue #62, opened 2017-02-27
Add authenticator for Oauth2


ADmad:
This would take some effort but wanted to create the issue in case someone wants to take a crack at it.

cleptric:
This is targeting to implement our own oauth 2 server, right?
Maybe we can work sth. out with https://oauth2.thephpleague.com

ADmad:
Not server, client for authentication.

ADmad:
Leveraging league's would definitely help though it would pull in guzzle :P

cleptric:
Ok, this makes things easier then :)
I'll take a look.

dereuromark:
Technically, it would be a suggest (or require dev), so people not using oauth2 (only jwt) wouldnt have to pull in guzzle then :)

ADmad:
Yes it would be suggest.

markstory:
Why would you need guzzle or another library to be an oauth2 client? Is there something more to it beyond setting the Authorization header correctly?

burzum:
@ADmad *someone*? Sounds like you volunteered by opening this issue? 😃

ADmad:
Nope,  I am just finishing, waiting for someone to bite the hook :)

cleptric:
I dived into this a little bit. Not an easy topic, but quite interesting :)
Supporting all the different grant types would result in copying the generic provider of league's oauth2 client, minus guzzle.
Would be a `JWT Bearer` grant supporting authenticator enough or do we want more?

burzum:
@cleptric we use a vendor for JWT as well, I don't see a reason why we can't do it here as well if the leagues lib works fine for what we want to do? Just add it to the suggest section of composer.json and get the authenticator to work. :)

ADmad:
I don't mind using league's oauth2 client either. It would be a "suggest" and not a hard dependency so the fact that it pulls in guzzle is not a problem.  Maybe I'll add a new authenticator to my hybridauth plugin for those who want to avoid guzzle :)

ADmad:
@cleptric Interested in checking out another lib? :smile:

I found a new lib https://github.com/SocialConnect/auth which seems decent. It has proper interfaces which  allow using custom adapter for handling requests and session, thereby avoiding any external dependencies like guzzle.

burzum:
@ADmad why do all these providers require different implementations? That's how it looks like at least to me. I thought Oauth was thought to be some kind of standard so that you *don't* have the fun of dealing with different implementations.

[...]

PR #110, opened 2017-04-02
SocialSignOnAuthenticator authenticator

cleptric:
Implements #62 
I wanted to collect some feedback on the current implementation.
Tests will follow.

### DIFF
--- a/src/Authenticator/OauthAuthenticator.php
+++ b/src/Authenticator/OauthAuthenticator.php
@@ -14,7 +14,9 @@ namespace Authentication\Authenticator;
 
 use Authentication\Identifier\IdentifierCollection;
 use Authentication\Result;
+use Cake\Http\Response;
 use Cake\Http\ResponseEmitter;
+use Cake\Http\ServerRequest;
 use Cake\Routing\Router;
 use Exception;
 use Psr\Http\Message\ResponseInterface;
@@ -132,10 +134,10 @@ class OauthAuthenticator extends AbstractAuthenticator
     /**
      * Checks the requests if it is the configured redirect action
      *
-     * @param \Psr\Http\Message\ServerRequestInterface $request The current request.
+     * @param \Cake\Http\ServerRequest $request The current request.
      * @return bool
      */
-    protected function _isOauthRedirectUrl(ServerRequestInterface $request)
+    protected function _isOauthRedirectUrl(ServerRequest $request)
     {
         $redirectUrl = $this->getConfig('redirectUrl');
         $this->_provider = implode('', $request->getParam('pass'));
@@ -159,10 +161,10 @@ class OauthAuthenticator extends AbstractAuthenticator
     /**
      * Checks the requests if it is the configured login action
      *
-     * @param \Psr\Http\Message\ServerRequestInterface $request The current request.
+     * @param \Cake\Http\ServerRequest $request The current request.
      * @return bool
      */
-    protected function _isLoginUrl(ServerRequestInterface $request)
+    protected function _isLoginUrl(ServerRequest $request)
     {
         $loginUrl = $this->getConfig('loginUrl');
         $this->_provider = $request->getQuery('provider');
@@ -221,10 +223,10 @@ class OauthAuthenticator extends AbstractAuthenticator
     /**
      * Redirect the user to the provider
      *
-     * @param \Psr\Http\Message\ResponseInterface $response Response object.
+     * @param \Cake\Http\Response $response Response object.
      * @return void
      */
-    protected function _redirect(ResponseInterface $response)
+    protected function _redirect(Response $response)
     {
         $provider = $this->_authService->getProvider($this->_provider);
 
--- a/src/Authenticator/OauthAuthenticator.php
+++ b/src/Authenticator/OauthAuthenticator.php
@@ -23,8 +23,6 @@ use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
 use RuntimeException;
 use SocialConnect\Auth\Service;
-use SocialConnect\Common\Http\Client\Curl;
-use SocialConnect\Provider\Session\Session;
 
 /**
  * Oauth Authenticator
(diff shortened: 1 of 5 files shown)


src/Authenticator/OauthAuthenticator.php
```diff
@@ -0,0 +1,243 @@
...
+            throw new RuntimeException('You must pass the `redirectUrl` option.');
+        }
+        if (empty($this->getConfig('loginUrl'))) {
+            throw new RuntimeException('You must pass the `loginUrl` option.');
+        }
+        if (empty($this->getConfig('authService'))) {
+            throw new RuntimeException('You must pass the `authService` option.');
+        }
+        if (!$this->getConfig('authService') instanceof Service) {
+            throw new RuntimeException('Option `authService` must be an insatce of \SocialConnect\Auth\Service.');
```
dereuromark: typo "instance"

tests/TestCase/Authenticator/OauthAuthenticatorTest.php
```diff
@@ -134,6 +134,182 @@ public function testAuthenticate()
     }
 
     /**
+     * testWorngUrls
+     *
+     * @return void
+     */
+    public function testWorngUrls()
```
dereuromark: typo wrong

tests/TestCase/Authenticator/OauthAuthenticatorTest.php
```diff
@@ -0,0 +1,470 @@
...
+            'redirectUrl' => [
+                'controller' => 'Users',
+                'action' => 'callback'
+            ],
+            'authService' => $authService
+        ]);
+
+        $result = $oauth->authenticate($request, $response);
+
+        $this->assertInstanceOf('\Authentication\Result', $result);
```
burzum: Can we use `Authentication::class` instead? Same in other places assertInstanceOf() is used?

src/Authenticator/OauthAuthenticator.php
```diff
@@ -0,0 +1,243 @@
...
+    protected function _isOauthRedirectUrl(ServerRequest $request)
+    {
+        $redirectUrl = $this->getConfig('redirectUrl');
+        $this->_provider = implode('', $request->getParam('pass'));
+
+        if (empty($this->_provider) || !array_key_exists($this->_provider, $this->_authService->getConfig()['provider'])) {
+            return false;
+        }
+
+        if (!empty($redirectUrl)) {
```
dereuromark: We could return early here to avoid nesting.

src/Authenticator/OauthAuthenticator.php
```diff
@@ -0,0 +1,243 @@
...
+    protected function _isLoginUrl(ServerRequest $request)
+    {
+        $loginUrl = $this->getConfig('loginUrl');
+        $this->_provider = $request->getQuery('provider');
+
+        if (empty($this->_provider) || !array_key_exists($this->_provider, $this->_authService->getConfig()['provider'])) {
+            return false;
+        }
+
+        if (!empty($loginUrl)) {
```
dereuromark: same here

src/Authenticator/OauthAuthenticator.php
```diff
@@ -0,0 +1,243 @@
...
+        $provider = $this->_authService->getProvider($this->_provider);
+
+        $response = $response
+            ->withStatus(302)
+            ->withLocation($provider->makeAuthUrl());
+
+        $emitter = new ResponseEmitter();
+        $emitter->emit($response);
+
+        exit;
```
ADmad: This is bad, the request shouldn't be short circuited this way with an `exit`.  A proper response instance should be returned to handle redirection, though some structural changes might be needed in the plugin to handle that.

src/Authenticator/OauthAuthenticator.php
```diff
@@ -0,0 +1,243 @@
...
+use Psr\Http\Message\ServerRequestInterface;
+use RuntimeException;
+use SocialConnect\Auth\Service;
+
+/**
+ * Oauth Authenticator
+ *
+ * Authenticates an identity based on a response from an Oauth provider.
+ */
+class OauthAuthenticator extends AbstractAuthenticator
```
cleptric: SocialNetworksAuthenticator?

src/Authenticator/OauthAuthenticator.php
cleptric: This would require to check the result of `authenticate` and redirect in the `AuthenticationMiddleware` if a response object is returned?
Don't know if this is a good idea either 🤔
I'm totally with you that this is a "dirty" fix, but I haven't found any other PSR7 conform solution to redirect early.

dereuromark: In the least this should be a protected _exit() etc, so you can mock it in tests :)
But a non exit way would sure be a bit cleaner.

src/Authenticator/SocialSignOnAuthenticator.php
```diff
@@ -0,0 +1,237 @@
...
+        $provider = $this->_authService->getProvider($this->_provider);
+
+        $response = $response
+            ->withStatus(302)
+            ->withLocation($provider->makeAuthUrl());
+
+        $emitter = new ResponseEmitter();
+        $emitter->emit($response);
+
+        exit;
```
markstory: This is not a good idea.

cleptric:
I have to close this PR. The current implementation has some flaws which I'm not sure will be fixable without ending up with a bad implementation of Oauth/Oauth2.
Big part of this decision was the following quote from the article linked below:
>OAuth 2.0 is not an authentication protocol.
https://oauth.net/articles/authentication/

I also talked with @markstory about this, too.

burzum:
@cleptric we should add a section in the docs why this is not going to include OAuth and explain the reasons.

robertpustulka:
@cleptric @burzum what about [OpenID Connect](http://openid.net/connect/) then which is said to be an authentication build on OAuth2?

Issue #112, opened 2017-04-03
Datasource backend proposal.


robertpustulka:
In #109 "backend" idea was introduced. I was wondering if we could transform this idea into kind of a pattern.

Currently both `OrmIdentifier` and `TokenIdentifier` (and identifiers that extend it) depend on ORM but do it differently. OrmIdentifier is an identifier build especially for identifying users by password using CakePHP ORM. TokenIdentifier can be extended and use other methods of token verification. Implementations differ but the idea is the same: query the ORM (or any other datasource) with given conditions (username/password or token) and return an identity (EntityInterface in this case).

I was thinking we could enhance this and make it easier to integrate with other ORMs.

My idea is to add `Authentication\Identifier\Backend\DatasourceInterface`:

```php
interface DatasourceInterface {
    //return value should be ArrayAccess instead of cake EntityInterface
    public function find(array $conditions): ArrayAccess;
}
```

And implement it for our ORM:
```php
//draft implementation
class OrmBackend implements DatasourceInterface {
    public function find(array $conditions) {
        return $this->table->find()->where($conditions)->first();
    }
}
```

Now both `OrmIdentifier` and `TokenIdentifier` could use this in order to query an identity from database (or any other datasource). 

For example:
```php
//TokenIdentifier
$identity = $this->backend->find([$tokenField => $token]);
...
//OrmIdentifier
$identity = $this->backend->find([$usernameField => $username]);
//password check etc
...
```

Developers could implement adapters (*backends*) for custom ORMs and still use those identifier classes.

I think `OrmIdentifier` should be then renamed to `PasswordIdentifier` to better reflect what this does since it would no longer be ORM dependent.

dereuromark:
+1 for PasswordIdentifier

cleptric:
Keep in mind that the `OauthAuthenticator` could also do an ORM lookup without a password as we don't have one in place in this case.

burzum:
> I was thinking we could enhance this and make it easier to integrate with other ORMs.

Honestly, except showing off how well it is abstracted I have doubts that anyone ever is using this with another ORM. SF and Laravel guys will stick to their stack and tools.

For any other person I feel this lib still has to many dependencies:

```
        "cakephp/core": "~3.4",
        "cakephp/datasource": "~3.4",
        "cakephp/event": "~3.4",
        "cakephp/utility": "~3.4",
```

[...]

PR #113, opened 2017-04-03
Identifier resolvers implementation

robertpustulka:
See #112

It's still WIP, but I wanted to show progress and get some feedback.

#### Todo
- [x] Implement resolvers.
- [x] Update TokenIdentifier.
- [x] Update classes that extend TokenIdentifier.
- [x] Refactor OrmIdentifier into PasswordIdentifier.
- [x] Add tests for ResolverAwareTrait.
- [x] Add tests for OrmResolver 
- [x] Remove Orm specific asserions from other tests.
- [x] Change return types from EntityInterface into ArrayAccess :question: 
- [x] Update docs.


src/Identifier/Backend/DatasourceAwareTrait.php
```diff
@@ -0,0 +1,98 @@
...
+
+        $class = App::className($config['className'], 'Identifier/Backend', 'Datasource');
+        if ($class === false) {
+            $message = sprintf('Datasource class %s does not exist.', $config['className']);
+            throw new InvalidArgumentException($message);
+        }
+        $instance = new $class($config);
+
+        if (!$instance instanceof DatasourceInterface) {
+            $message = sprintf('Datasource must implement %s.', DatasourceInterface::class);
```
burzum: Aren't we usually wrapping the %s into ` ?

src/Identifier/Backend/DatasourceAwareTrait.php
```diff
@@ -0,0 +1,98 @@
...
+                'className' => $config
+            ];
+        }
+
+        if (!isset($config['className'])) {
+            $message = 'Option `className` is not present.';
+            throw new InvalidArgumentException($message);
+        }
+
+        $class = App::className($config['className'], 'Identifier/Backend', 'Datasource');
```
burzum: As discussed before, wouldn't this make the app class a hard requirement? We would like to try make it as less hard coupled to Cake as possible - if possible. So check if the class exists and if not go without that part?

src/Identifier/TokenIdentifier.php
```diff
@@ -12,34 +12,32 @@
...
     ];
 
     /**
-     * Identify
+     * Identify method.
      *
-     * @param array $data Authentication credentials
-     * @return \Cake\Datasource\EntityInterface|null
+     * @param array $data Authentication credentials.
+     * @return \ArrayAccess|null
```
burzum: Any reason you've already changed this without a decision on if we use \ArrayAccess or \EntityInterface?

src/Identifier/Backend/DatasourceAwareTrait.php
robertpustulka: @burzum we're depending on `InstanceConfigTrait` and `ObjectRegistry` from core package so the `App` class would be present anyway. Let's maybe hold with that until this cakephp/core dependency issue is resolved.

src/Identifier/TokenIdentifier.php
robertpustulka: @burzum right. 
`DatasourceInterface` returns `ArrayAccess` in my case because we can't force users to return EntityInterface in their implementations I think. So I changed the typehint  here. I could revert this though but let's hear what others have to say about it.

src/Identifier/Backend/DatasourceAwareTrait.php
burzum: OK

src/Identifier/PasswordIdentifier.php
```diff
@@ -130,51 +142,21 @@ protected function _findUser($identifier, $password = null)
...
-        if (is_array($usernameFields)) {
-            $conditions = [];
-            foreach ($usernameFields as $field) {
-                $conditions[$table->aliasField($field)] = $identifier;
-            }
+        $fields = $this->getConfig('fields.username');
+        foreach ((array)$fields as $field) {
             $conditions = [
-                'OR' => $conditions
+                $field => $identifier
```
ADmad: I don't like this. Removing the `OR` means for multi column usage it now uses two queries to find matching records.

robertpustulka: @ADmad we can't support `OR` syntax for every ORM available.

Although we could change method definition of `DatasourceInterface`:
```php
public function find(string|array $field, string|array $value);
```
For our ORM that could be implemented something like that:
```php
$conditions = [
    'OR' => [
        'field_1 in' => $values
        ...
    ]
];

Comment (ADmad) [inline src/Identifier/PasswordIdentifier.php]: > we can't support OR syntax for every ORM available.

Right, we can't, so the API should be changed. Your suggestion sounds good.

File: src/Identifier/Backend/DatasourceAwareTrait.php (hunk 2 above)
Comment (markstory) [inline src/Identifier/Backend/DatasourceAwareTrait.php]: Don't forget that people will assume that CakePHP plugins integrate well with the rest of the framework/application stack, and classname resolution is something we do in CakePHP.

File: src/Identifier/PasswordIdentifier.php (hunk 4 above)
Comment (markstory) [inline src/Identifier/PasswordIdentifier.php]: What other ORM's are there support for so far?

Comment (robertpustulka) [inline src/Identifier/PasswordIdentifier.php]: @markstory none. Should we also do this? I'm not sure, that's something a developer should implement for their need. We just support our ORM.

Comment (markstory) [inline src/Identifier/PasswordIdentifier.php]: Couldn't we continue assuming that we can support the `OR` syntax until its proven we can't? At least each IdentityResolver can adapt the `OR` syntax into the queries necessary for that ORM backend.

File: src/Identifier/Backend/DatasourceAwareTrait.php (hunk 2 above)
Comment (ADmad) [inline src/Identifier/Backend/DatasourceAwareTrait.php]: `cakephp/core` is a pretty smallish lib, so depending on it to ensure convenience for Cake users isn't a problem.

File: src/Identifier/Resolver/ResolverInterface.php [hunk 5]
```diff
@@ -10,16 +10,20 @@
...
     /**
-     * Returns identity with given conditions from a datasource.
+     * Returns identity with given conditions.
      *
      * @param array $conditions Find conditions.
+     * @param string $type Condition type. Can be `AND` or `OR`.
      * @return \ArrayAccess|null
      */
-    public function find(array $conditions);
+    public function find(array $conditions, $type = self::TYPE_AND);
```
markstory: Do we have any scenarios where someone would need a mixture of and/or?

dereuromark: One would be E-Mail or Password and the other with some active flag right?

robertpustulka: @dereuromark you can use your custom `active` finder for that. Currently we have no such situations coming from Identifier classes. One uses single `token => value` condition, another `username/email/etc => value` condition with OR.

inoas: ... like?: is_admin OR (is_validated AND is_active)

robertpustulka: @inoas these kinds of conditions will not come from identifiers. They pass conditions based on data retrieved from authenticators.

robertpustulka:
It is ready to be reviewed I think.

src/Identifier/Resolver/ResolverAwareTrait.php
```diff
@@ -0,0 +1,98 @@
...
+ * @license       http://www.opensource.org/licenses/mit-license.php MIT License
+ */
+
+namespace Authentication\Identifier\Resolver;
+
+use Cake\Core\App;
+use InvalidArgumentException;
+use RuntimeException;
+
+trait ResolverAwareTrait
```
burzum: Can we name this `IdentityResolverAwareTrait`? Right now its very generic. The name gives no clue what it will resolve.

src/Identifier/Resolver/ResolverInterface.php
```diff
@@ -0,0 +1,29 @@
...
+ * For full copyright and license information, please see the LICENSE.txt
+ * Redistributions of files must retain the above copyright notice.
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (http://cakefoundation.org)
+ * @link          http://cakephp.org CakePHP(tm) Project
+ * @license       http://www.opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Authentication\Identifier\Resolver;
+
+interface ResolverInterface
```
burzum: IdentityResolverInterface?

src/Identifier/TokenIdentifier.php
burzum: @markstory your word on this?

src/Identifier/Resolver/ResolverAwareTrait.php
robertpustulka: @burzum It's under `Identifier` namespace so it should be obvious what it's used for. 

In ORM we have a "generic" `LocatorAwareInterface` and `LocatorInterface` but it's under `Cake\ORM` so the scope is somewhat clear.

src/Identifier/Resolver/ResolverInterface.php
dereuromark: I agree with @robertpustulka that ResolverInterface etc should be enough

Otherwise

    class OrmResolver implements ResolverInterface

should also be

    class OrmIdentityResolver implements IdentityResolverInterface

to be consistent here. And that is a bit too much maybe.

robertpustulka:
Let's merge it if that's OK and no one has any objections, so we can continue with #109 and reduce the number of dependnecies as discussed in #112

burzum:
@robertpustulka I would like to wait for @markstory feedback on the Entity vs ArrayObject.

robertpustulka:
@burzum that has been discussed [here](https://github.com/cakephp/authentication/issues/112#issuecomment-292270057)

PR #118, opened 2017-05-03
Add array type hint to IdentifierInterface::identify

cleptric:
`array_intersect_key ` will trigger a warning if the `$data` argument is not an array.

### DIFF
--- a/src/Identifier/IdentifierInterface.php
+++ b/src/Identifier/IdentifierInterface.php
@@ -18,10 +18,10 @@ interface IdentifierInterface
     /**
      * Identifies an user or service by the passed credentials
      *
-     * @param mixed $credentials Authentication credentials
+     * @param array $credentials Authentication credentials
      * @return \ArrayAccess|null
      */
-    public function identify($credentials);
+    public function identify(array $credentials);
 
     /**
      * Gets a list of errors happened in the identification process
--- a/src/Identifier/CallbackIdentifier.php
+++ b/src/Identifier/CallbackIdentifier.php
@@ -64,7 +64,7 @@ class CallbackIdentifier extends AbstractIdentifier
      * @param array $data Authentication credentials
      * @return \ArrayAccess|null
      */
-    public function identify($data)
+    public function identify(array $data)
     {
         $callback = $this->getConfig('callback');
 
--- a/src/Identifier/LdapIdentifier.php
+++ b/src/Identifier/LdapIdentifier.php
@@ -133,7 +133,7 @@ class LdapIdentifier extends AbstractIdentifier
      * @param array $data Authentication credentials
      * @return \ArrayAccess|null
      */
-    public function identify($data)
+    public function identify(array $data)
     {
         $this->_connectLdap();
         $fields = $this->getConfig('fields');
--- a/src/Identifier/PasswordIdentifier.php
+++ b/src/Identifier/PasswordIdentifier.php
@@ -89,7 +89,7 @@ class PasswordIdentifier extends AbstractIdentifier
      * @param array $data Authentication credentials.
      * @return \ArrayAccess|null
      */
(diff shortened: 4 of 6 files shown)


dereuromark:
If it is allowed to be null, shouldnt that be reflected in the doc block?

    @param array|null $data Authentication credentials.

markstory:
Any tests we can add?

cleptric:
@dereuromark The`IdentifierInterface` enforces a none optional parameter. This should just prevent the warning when passing a string.

dereuromark:
I didnt mean nullable, but null as value. So the doc block should still say it if you internally check for null.
Otherwise you should either ignore or throw a meaningful exception :)
But since it is clearly OK here, the doc block should be fixed to clearly specify the null as value.

dereuromark:
The alternative would be to fix the calling code to array cast - this way there would be no null value input, always array, and the code stays clean.

cleptric:
@dereuromark Yes, this was a misunderstanding on my side about the `null` value :)
Well, I don't know what would be the best thing to do here when looking into the other identifiers.
Casting the data passed to `$this->identifiers()->identify($userData)` might be bad if we implement an Identifier that accepts a string.
I'm working on the `CookieAuthenticator` right now, which could pass a string, array or null.

dereuromark:
IMO it should still continue to do what it always did as per doc block: array only.
Cast the null to array would solve the root issue here.

robertpustulka:
If we really want to have arrays only shouldn't we add a hard typehint to the method declaration in the interface?

cleptric:
I looked through the authenticators and all which are calling `identify()` are passing arrays. So I'm fine with adding the typehint to the Interface.

robertpustulka:
@cleptric go for it :)

dereuromark:
This should go into 3.5 then, as technically this will be a BC break for all custom authenticators.
Instead for master you should fix the incoming data to cast it beforehand as I mentioned earlier.

cleptric:
@dereuromark Should we already thing about BC as this plugin hasn't been released yet?

robertpustulka:
@dereuromark I'm afraid there are no custom authenticators yet :)

src/Identifier/IdentifierInterface.php
```diff
@@ -18,10 +18,10 @@
     /**
      * Identifies an user or service by the passed credentials
      *
-     * @param mixed $credentials Authentication credentials
-     * @return \ArrayAccess|null
+     * @param array $credentials Authentication credentials
+     * @return mixed
```
robertpustulka: Should be `\ArrayAccess|null`. You have probably overwritten changes to the master.
Try `git rebase upstream/master` so the changes are reflected in the correct order.

src/Identifier/IdentifierInterface.php
```diff
@@ -18,10 +18,10 @@
     /**
      * Identifies an user or service by the passed credentials
      *
-     * @param mixed $credentials Authentication credentials
```
robertpustulka: Should be \ArrayAccess|null. You have probably overwritten changes to the master.
Try git rebase upstream/master so the changes are reflected in the correct order.

src/Identifier/IdentifierInterface.php
robertpustulka: Should be `\ArrayAccess|null`. You have probably overwritten changes to the master.
Try `git rebase upstream/master` so the changes are reflected in the correct order.


*sorry for the another edit, too much work today*

cleptric: I did `git pull --rebase origin master`. `git rebase origin/master` is telling me `Current branch fix-password-identifier is up to date.`

cleptric:
I squashed and applied the missing changes from @robertpustulka

Issue #26, opened 2016-12-07
Adding a cookie authenticator


burzum:
I know this exists https://github.com/Xety/Cake3-CookieAuth/blob/master/src/Auth/CookieAuthenticate.php but the code is really trivial to do and maintain with our identifiers. :) So what about shipping the plugin with a cookie authenticator out of the box?

**Edit**: We'll implement it, just have to wait for the Cookie changes that we need to get done before in the request and response objects.

dereuromark:
Sounds like a good idea to me!

ADmad:
Sure, once the API is finalized.

burzum:
We'll also need to work on the cookie handling because right now we're stuck on the component, which is not available in the middleware stack... I've already discussed the issue with Mark Story. We'll address it and add the cookie authenticator when this has been done.

burzum:
There is a branch open for this feature if somebody wants to help:

https://github.com/cakephp/authentication/tree/cookie-authenticator

cleptric:
Closing as we have an open PR.

PR #119, opened 2017-05-05
Cookie authenticator

burzum:
Refs #26


src/Authenticator/CookieAuthenticator.php
```diff
@@ -0,0 +1,121 @@
...
+        return new Result($user, Result::SUCCESS);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function persistIdentity(ServerRequestInterface $request, ResponseInterface $response, $identity)
+    {
+        $cookie = $this->getConfig('cookie');
+        $cookie['value'] = $identity;
```
ADmad: Identity here would be user entity isn't it?  I don't think it's a good idea to put the whole user record in the cookie.

markstory: What about just the primary key? We would need to reload the User from the datastore on each request, but it seems safer than trying to put the user in a possibly unverified/unprotected cookie.

burzum: What about giving the user the choice what he wants to store in the cookie?

ADmad: Shouldn't persistIdentity() set identity to session instead of cookie? Setting cookie here means the cookie gets written whenever a user is authenticated. You usually write "remember me" cookie only when user explicitly selects that feature.

burzum: @ADmad good question. But I think we need both:

1) Form / POST authenticates user, needs to store identity in session + cookie (optional)
2) Read from cookie, write to session

I don't remember how exactly it works right now, but I don't think we have something in the plugin that would cover this automatically. So you would have to manually persist the cookie in session and cookie. However, even if you do it manually you need to call the persistIdentity() method and store it somewhere.

Not sure how to solve this the best right now.

src/Authenticator/CookieAuthenticator.php
```diff
@@ -0,0 +1,121 @@
...
+            throw new RuntimeException('You must use the CakePHP `3.next` branch or wait until CakePHP version 3.5 is released to use the CookieAuthenticator');
+        }
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function authenticate(ServerRequestInterface $request, ResponseInterface $response)
+    {
+        $userData = $request->getCookie($this->getConfig('cookie.name'));
```
robertpustulka: `getCookie()` method isn't a part of `ServerRequestInterface`. There's `getCookieParams()` instead.

src/Authenticator/CookieAuthenticator.php
```diff
@@ -0,0 +1,121 @@
...
+     * {@inheritDoc}
+     */
+    public function persistIdentity(ServerRequestInterface $request, ResponseInterface $response, $identity)
+    {
+        $cookie = $this->getConfig('cookie');
+        $cookie['value'] = $identity;
+
+        return [
+            'request' => $request,
+            'response' => $response->withCookie(
```
robertpustulka: There's no `withCookie()` method in PSR `ResponseInterface`.

We should do some check for that method or return a response `withHeader()`.

src/Authenticator/CookieAuthenticator.php
```diff
@@ -0,0 +1,121 @@
...
+     * @return array Returns an array containing the request and response object
+     */
+    public function clearIdentity(ServerRequestInterface $request, ResponseInterface $response)
+    {
+        $cookie = $this->getConfig('cookie');
+        $cookie['expire'] = strtotime('-1 year');
+
+        return [
+            'request' => $request,
+            'response' => $response->withCookie(
```
robertpustulka: Same thing with `withCookie()` method.

src/Authenticator/CookieAuthenticator.php
ADmad: Right, `CookieAuthenticator::persistIdentity()` could just write the identity to session. 

Most common use case is user selects "remember me" checkbox on login form, so could we perhaps make `FormAuthenticator` write the cookie if checkbox is checked?

burzum: @ADmad we need to make sure the cookie implementation is consistent across both then. But yes, we could define a default field to look for, maybe `rememer_me` and if its present and not `0` then write to the cookie as well.

It would be nice if we already can get access to the other loaded authenticators, we can use call the persistsIdentity() method on the other authenticator. :)

ADmad: We could add a `AuthenticationService::getAuthenticator()` method.

burzum: Don't we want to have a `hasAuthenticator()` then as well?

ADmad: I would just throw an exception.

burzum: @ADmad we can already use `AuthenticationService::authenticators()->get()` :) It's calling the AuthenticatorCollection and it extends ObjectRegistry which provides get(). But it returns null instead of throwing an exception. We could go with that or override the method.

ADmad: Indeed!  So we could do without wrapper/convenience methods too.

robertpustulka: How would you access the `AuthenticationService`?

ADmad: @robertpustulka The `AuthenticationMiddleware` sets `AuthenticationService` instance as [request attribute](https://github.com/cakephp/authentication/blob/master/src/Middleware/AuthenticationMiddleware.php#L69).

robertpustulka: How can we make sure that session data does not expire quicker than a `remember_me` cookie?
This authenticator-mixed approach seems a bit hacky to me...

Let's summarize, what we want to achieve:
- have a `remember_me` checkbox in a login form
- persist the identity if a checkbox is checked until the user logs them out manually

That does not mean we have to do this using cookies. We could as well utilize cache or session for that.

src/Authenticator/CookieAuthenticator.php
burzum: @robertpustulka all of this has already been discussed, it just needs to be implemented. :)

> 12:26] 
burzum Hi Mark, any idea how to resolve these two? https://travis-ci.org/cakephp/authentication/jobs/229100440
[12:26]
slackbot APP markstory has snoozed notifications. Send one anyway? Send notification
Only visible to you
[12:26] 
burzum The cause is that the methods use the Interface and the interface doesn't implement these methods.
[12:26] 
But our ServerRequest object does
[12:27] 
If we want to keep this generic the solution would be to check if the instance is a cake ServerRequest and otherwise use whatever the default it
[18:42] 
markstory I would remove the typehint or use the correct one.
[19:09] 
burzum The main question is this: Do we want to make the cookie authenticator to work without cakephp as well?
[19:10] 
If we rely on cakes ServerRequest and Response we'll introduce a dependency we don't have yet
[22:19] 
markstory I am ok with that. Cookies with the base objects is pretty awkward
[23:14] 
burzum So this means you want what I suggest, support for both OR only cake support? :slightly_smiling_face:

robertpustulka: @burzum but our cookies already have `toHeaderValue()` or sth method. We could just send a header and keep PSR interfaces.

src/Authenticator/CookieAuthenticator.php
burzum: @ADmad I've pushed a basic implementation but I think there is some room for abstraction, maybe creating a trait from this. The code is very similar. So we could make it a trait and allow calling any other authenticator by just checking if the array is empty, if not run over it, check if each authenticator exists and implements persistence interface and then call it's persistence method.

src/Authenticator/FormAuthenticator.php
```diff
@@ -87,10 +87,37 @@ public function authenticate(ServerRequestInterface $request, ResponseInterface
...
+        if ($cookieAuthenticator === null) {
+            return;
+        }
+
+        $cookieAuthenticator
+            ->authenticators()
+            ->get('Cookie');
+
+        if ($cookieAuthenticator instanceof CookieAuthenticator) {
+            return $cookieAuthenticator->persistIdentity($request, $response, $user);
```
robertpustulka: I don't think that resolves the issue. The identity is being stored in a cookie as well. We wanted to avoid that if I understand correctly.

And this breaks the whole SoC idea IMO.

src/Authenticator/CookieAuthenticator.php
burzum: @robertpustulka I've just got an answer from @markstory:

> 23:26] 
markstory Supporting cookies with the base objects is awkward enough that I would make cookie auth cake only. Isn't session storage effectively cake only as well?
[23:27] 
burzum Good question

robertpustulka: @burzum the issue here isn't the server request/response implementation but cookie implementation. We already have means to keep compatibility with PSR interfaces. We can get cookie value from `getCookieParams()` and set a cookie to the response with `withAddedHeader()`. The value for the header can be retrieved from `Cookie::toHeaderValue()`.

No instanceof check and limiting the implementation to cake `ServerRequest` needed.

src/Authenticator/CookieAuthenticator.php
robertpustulka: If I understand correctly we want to implement the "remember me" feature and not to store the whole identity in a cookie. The cookie here should only be used to indicate if a current user wants to be logged in forever in the current browser. The session could expire eventually so I don't think that storing the identity in a session is OK.

src/Authenticator/CookieAuthenticator.php
robertpustulka: @burzum OK, I'll handle this

src/Authenticator/CookieAuthenticator.php
markstory: Using `withHeader()` on the `Set-Cookie` header means appending/replacing the serialized cookie to the header value if it the header exists, or adding a new header value.

src/Authenticator/CookieAuthenticator.php
markstory: I'm on-board if you want to make this authenticator work with the PSR7 interfaces. I've found Cookies to be a bit tedious to work with at the header level. If we do end up staying with the CakePHP specific methods, the typehint should be corrected/removed.

src/Authenticator/CookieAuthenticator.php
robertpustulka: I meant withAddedHeader()

src/Authenticator/CookieAuthenticator.php
burzum: @markstory does the cookie encryption middleware work as well with the PSR7 interface? If this authenticator gets data it needs to be able to somehow decrypt it if it is encrypted.

robertpustulka:
I've been thinking about this authenticator and I'm not sure we have a clarity on what this authenticator should do.

I think there should be some kind of "remember me in this browser" authenticator. 
As I mentioned in the comments above I believe this authenticator should:
- support a `remember_me` checkbox in a login form
- persist the identity if a checkbox is checked until the user logs them out manually.

So in the `authenticate()` method we should probably check for `remember_me` field in request body and set a cookie with the checkbox value. Then comes authenticating itself. I think there are two ways:
1) Store user credentials in a cookie. We've been thinking about the primary key, but in that case we would limit this implementation to datastores that had some kind of primary key. And we would need to implement another identifier for checking against the PK. Another way is to use standard username/password credentials. Note that the cookie should be encrypted in both cases.
2) Support cache. We could persist the identity in the cache and retrieve it from there. The cookie could contain some partial cache key. Using cache instead of session would ensure that the identity is stored "forever". The cookie does not need the encryption as long as we use some hash for the cache key.

In both cases identiy persistence (store user credentials or cache key in a cookie) should happen only if there's `remember_me` field checked.

If you ask me I think that the option 2) is safer (you don't keep sensitive information in a browser) but I'm not sure about depending on cache storage.

src/Authenticator/CookieAuthenticator.php
```diff
@@ -0,0 +1,130 @@
...
+
+    /**
+     * {@inheritDoc}
+     */
+    public function persistIdentity(ServerRequestInterface $request, ResponseInterface $response, $identity)
+    {
+        $data = $this->getConfig('cookie');
+        $data['value'] = $identity;
+
+        $cookie = new Cookie(
```
robertpustulka: What if we had static `Cookie::createFromArray()` on the core?

ADmad:
For reference here's how symfony generate the remember cookie https://github.com/symfony/security/blob/master/Http/RememberMe/TokenBasedRememberMeServices.php

src/Authenticator/CookieAuthenticator.php
```diff
@@ -0,0 +1,130 @@
...
+            $data['expire'],
+            $data['path'],
+            $data['domain'],
+            $data['secure'],
+            $data['httpOnly']
+        );
+
+        return [
+            'request' => $request,
+            'response' => $response->withAddedHeader('Set-Cookie', $cookie->toHeaderValue())
```
ADmad: We still need to address the fact that the cookie needs to be written only if `remember_me` checkbox is checked on login form. Once #121 is merged `persistIdentity()` will be called for all configured authenticator, so just checking that `remember_me` is set to 1 in request body should suffice isn't it?

robertpustulka: @ADmad yes. I didn't touch storing the identity in a cookie yet. I'm waiting for a decision how should we proceed. Symfony token-based approach is also interesting. We could leverage our token identifier for that.

ADmad: @robertpustulka That symfony file I linked is basically generating a JWT :) It uses HMAC with SHA256 which is the same what JWT does if you use `HS256` algorithm.

burzum: @ADmad we could do the very same just as they do it. I've seen they have an interface for "RememberMe", not sure if something like that is *really* needed.

robertpustulka: The token based approach would require developers to have some token field in a table just for "remember me" feature. Is that OK with you?

burzum: It should be optional IMO.

robertpustulka: @burzum so are we talking about two different implementations of remember me authenticator?

burzum: @robertpustulka that or can't we simply make it a configurable option? Honestly, I've never seen an app in the past using a token inside a cookie the way we want it now. I like the idea but I think we should provide both ways. We could promote the token inside the cookie in the docs as more secure and that people should prefer it but we shouldn't enforce it. Certain people will anyways always find a way to break stuff.

robertpustulka: @burzum OK, so we could store encrypted username/password credentials or a token in a cookie as an option? This could work.

burzum: Yes. Just leave it up to the user what he wants to store there. Complete user data, just credentials, just a token. Encrypted or not encrypted, their cup of tea. We should always promote best practice and most secure options but not limit users. For example what happens if you need to read the worst case: Non encrypted cookie data that contains the whole user data from some legacy app.

ADmad: > The token based approach would require developers to have some token field in a table just for "remember me" feature.

JWT don't require to be saved to db. So if were to use JWT then we generate JWT and save in cookie. The JWT payload would contain id or username. For authentication you just decrypt token, get id/username from payload, get matching user from db and persist to session.

robertpustulka: @ADmad that would require to add specific identifier for that. Allowing other authenticators to look for an identity using only id or username doesn't sound secure.

ADmad: JWT is an industry standard. What i suggested above is the same what JWT identifier does.

robertpustulka: @ADmad right, I was looking at `TokenIdentifier`. Seems like JWT is looking for an id.
So we should suggest using JWT identifier for remember me feature. Seems OK to me.

So this cookie/remember me authenticator should extend JWT and look for a token in a cookie instead of a header? Also a token should be persisted in a cookie if `remember_me` is checked?

burzum: @ADmad ping, any input on this from your side please?

src/Authenticator/CookieAuthenticator.php
markstory: Once the request is passed on by the EncrtptedCookieMiddleware, all cookie values are decrypted, so the authenticator should not see encrypted data if used properly.

src/Authenticator/CookieAuthenticator.php
ADmad: What @robertpustulka suggested sounds good to me.

burzum: @ADmad @robertpustulka I'm OK with making an encrypted token value the default but I don't think we need to extend JWT, we can simply use the single line call for that. But we should give the user the option to not only use a JWT token inside the cookie. Anyone interested in doing this?

robertpustulka: @burzum we should follow SR principle and have only one implementation per authenticator. We could provide JWT based implementation and some other one.

BTW do we need to encrypt the JWT token before it is sent with the cookie? Isn't that already encrypted?

burzum: @robertpustulka then lets have two cookie authenticators? JwtCookie and Cookie?

AFAIK JWT tokens are always encrypted, it's part of it.

markstory: JWT are not always encrypted. The tokens I have seen use sha256 checksums instead of encrypted payloads.

burzum: @markstory thats why I'm asking if we really need to couple it with the JWT lib instead of simply having a simple function for that. Encryption of the cookie should happen anyway in the cookie middleware?

markstory: > Encryption of the cookie should happen anyway in the cookie middleware?

Yes.

src/Authenticator/CookieAuthenticator.php
burzum: @robertpustulka @markstory I've made the tests work and it looks like what you, Robert, wanted a few comments above is already done? So if you agree I would merge it now.

src/Authenticator/CookieAuthenticator.php
ADmad: JWT supports multiple encryption algorithms. We can use the the popular `HS256` algorithm for encryption.

I have always wanted this plugin to be usable outside Cake too, so relying on the middleware to handle encryption / decryption would  prevent that.

src/Authenticator/CookieAuthenticator.php
robertpustulka: @burzum I'm not sure this authenticator does what it ws intended to do. Currently it does not provide "remember me" feature. It looks like it always stores an entire identity in a cookie and tries to identify a user using the stored identity.

src/Authenticator/CookieAuthenticator.php
burzum: @ADmad we should make it configurable then. If somebody wants to use the encryption middleware you would need to disable encryption because its done externally.

robertpustulka: @burzum I think that it would be better to include this authenticator in the next version of the plugin than to include it prematurely now. It looks like we still haven't settled on what resposibility this authenticator should have and how it should be implemented.

There's no core cookie authenticate in the current component-based implementation, so there would be nothing missing from 3.5 if we deprecated the component in favor of this plugin.

burzum: Agreed. So lets do an RC1?

robertpustulka: OK, let's do this :+1:

ADmad: > we should make it configurable then. If somebody wants to use the encryption middleware you would need to disable encryption because its done externally. I've checked the JWT lib, we could extract the part for the signing and put it into a trait and use it in our cookie authenticator? Or even better, improve our already existing https://github.com/cakephp/cakephp/blob/3.next/src/Utility/CookieCryptTrait.php with it.

Sounds good to me.

robertpustulka: @burzum are we doing RC1 of this?

burzum: @robertpustulka I think the last decision was to do a RC1 and finish this cookie stuff after 3.5 is released and after the plugin is stable.

burzum: @robertpustulka I've added the remember me check as you requested here https://github.com/cakephp/authentication/pull/119#issuecomment-299810868 in my last commit.

src/Authenticator/CookieAuthenticator.php
```diff
@@ -0,0 +1,131 @@
...
+
+        if (!is_array($data) || empty($data[$field])) {
+            return [
+                'request' => $request,
+                'response' => $response
+            ];
+        }
+
+        $data = $this->getConfig('cookie');
+        $data['value'] = (array)$identity;
```
robertpustulka: Unfortunetally array casting won't work for `Entity` instances. We need some other way of storing the identity data in a cookie. We could serialize the object or extract only some data needed for identification.

We can leverage `VerificationIdentifier` (if #151 gets merged) and store only the `id` field needed for reverification.

docs/Authenticators.md
```diff
@@ -69,6 +69,26 @@ Configuration options:
 * **nonce**: Default is `uniqid(''),`
 * **opaque**: Default is `null`
 
+## Cookie Authenticator aka "Remember Me"
+
+The Cookie Authenticator allows you to implement the "remember me" feature for your login forms.
+
+Just make sure your login form has a field field that matches the field name that is configured in this authenticator.
```
markstory: Double 'field' here.

src/Authenticator/CookieAuthenticator.php
```diff
@@ -0,0 +1,131 @@
...
+        $this->_defaultConfig['rememberMeField'] = 'remember_me';
+        $this->_defaultConfig['cookie'] = [
+            'name' => 'CookieAuth',
+            'expire' => null,
+            'path' => '/',
+            'domain' => '',
+            'secure' => false,
+            'httpOnly' => false,
+            'value' => ''
+        ];
```
markstory: Couldn't this default config be defined in the class?

burzum: @markstory done.

tests/TestCase/Authenticator/CookieAuthenticatorTest.php
```diff
@@ -0,0 +1,215 @@
...
+        ]);
+
+        $request = ServerRequestFactory::fromGlobals(
+            ['REQUEST_URI' => '/testpath'],
+            null,
+            null,
+            [
+                'CookieAuth' => [
+                    'username' => 'mariano',
+                    'password' => 'password'
```
markstory: Requiring or expecting passwords in cookies seems like an insecure practice. Can we come up with another approach?

burzum: @markstory @robertpustulka any better idea?

dereuromark: Dont other existing solutions use tokens?

markstory: One idea I proposed in slack today was storing `$username . ':' . sha256($username . $passwordHash . $salt)` as the 'token'. This would allow you to find the user again by username, and verify the identity of the token without exposing the password/password hash, and without requiring additional columns/storage.  This token would also be invalidated when the username, or password hash were updated making it compatible with common password reset flows.

robertpustulka:
I updated the authenticator with a token based authentication according to @markstory suggestion.

tests/TestCase/Authenticator/CookieAuthenticatorTest.php
```diff
@@ -135,6 +125,34 @@ public function testCredentialsNotPresent()
...
+        $identifiers = new IdentifierCollection([
+            'Authentication.Password'
+        ]);
+
+        $request = ServerRequestFactory::fromGlobals(
+            ['REQUEST_URI' => '/testpath'],
+            null,
+            null,
+            [
+                'CookieAuth' => 'mariano:$2y$10$1bE1SgasKoz9WmEvUfuZLeYa6pQgxUIJ5LAoS/asdasdsadasd'
```
markstory: Should we have a test for the valid password hash with a mismatched username on the left side of the token?

src/Authenticator/CookieAuthenticator.php
```diff
@@ -97,18 +82,25 @@ public function authenticate(ServerRequestInterface $request, ResponseInterface
                 'Login credentials not found'
             ]);
         }
-        $userData = $cookies[$cookieName];
-        if ($this->getConfig('encrypt')) {
-            $userData = $this->_decrypt($userData, $this->getConfig('encryptionType'));
-        }
+        $token = $cookies[$cookieName];
+        list($username, $tokenHash) = explode(':', $token);
```
markstory: The token could be mangled and not contain 2 parts. We should probably check that there is a `:` in the token and fail authentication early if it is not present.

robertpustulka: Right.

Should we encode the username as it may contain `:` character?

markstory: Could use json_encode().

src/Authenticator/CookieAuthenticator.php
```diff
@@ -0,0 +1,208 @@
...
+    {
+        $cookies = $request->getCookieParams();
+        $cookieName = $this->getConfig('cookie.name');
+        if (!isset($cookies[$cookieName])) {
+            return new Result(null, Result::FAILURE_CREDENTIALS_NOT_FOUND, [
+                'Login credentials not found'
+            ]);
+        }
+        $token = $cookies[$cookieName];
+        $token  = explode(':', $token);
```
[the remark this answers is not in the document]
burzum: Fuck you PSR spaces

burzum:
@robertpustulka hmm the json representation of `dd(json_encode('mari:ano'));` is `'"mari:ano"'` If we use base64 an username is not allowed to contain `/`. So there will be always some kind of a caveat.

robertpustulka:
What if we wrapped the username and a token in an array and json_encode that?

```php
$token = json_encode([$username, $hash]);
list($username, $token) = json_decode($token);
```

src/Authenticator/CookieAuthenticator.php
```diff
@@ -0,0 +1,208 @@
...
+        if (count($token) !== 2) {
+            return new Result(null, Result::FAILURE_CREDENTIAL_INVALID, [
+                'Cookie token is invalid.'
+            ]);
+        }
+
+        list($username, $tokenHash) = $token;
+
+        $credentials = [
+            'username' => json_decode($username)
```
markstory: If we json_decode after the split I don't think we'll get around `:` in usernames. I think we'd need to json_encode/decode the entire token to be safe. Something like

```php
$token = json_encode([$username, $hash]);
```

And when checking the token something like:

```php
$value = json_decode($cookies[$cookeName], true);
if ($value === null || count($value) !== 2) {
  // return error
}
```

robertpustulka:
I think we probably need the URL check similar to the one implemented in FormAuthenticator. Currently the CookieAuthennticator will look for a `remember_me` post data key in all requests.

robertpustulka:
Sorry, master rebase didn't go as well as I expected :/

burzum:
@robertpustulka I would like to move the login URL check into an object and inject it. After reading this http://rosstuck.com/how-i-use-traits I think the statement of the article is true:

> If you’re running into it very often, you probably need to step back and reexamine your style of programming. There’s a good chance you’ve got tons of objects waiting to be extracted.

I think we should take the opportunity for Cake4 and refactor traits were we can into separate objects for the reasons named in the article, they make sense.

robertpustulka:
I’m OK with that. How about extracting url check config into a seperate array and building url checker object with that?

```
$service->loadAuthenticator('Form', [
    'url' => [
        //config here
    ]
]);

**Comment by robertpustulka** [comment]
I’d handle that in another PR though ;)

**Comment by burzum** [comment]
Meh, read this after I've done it... I've just refactored the Trait into a class... @robertpustulka Could you please review the change?

File: src/Authenticator/LoginUrlCheckerInterface.php [hunk 16]
```diff
@@ -0,0 +1,31 @@
...
+ * @license       http://www.opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Authentication\Authenticator;
+
+use Psr\Http\Message\ServerRequestInterface;
+
+/**
+ * LoginUrlCheckerInterface
+ */
+interface LoginUrlCheckerInterface
```
robertpustulka: Do we need an interface for an internal class? There's a very little chance that somebody would replace that checker with their own implementation. Wouldn't just a callable/invokable class be sufficent?

burzum: Yes, there should be an interface IMHO. I haven't yet added a type check on this and actually, I might use my own checker. The interface provides a clear implementation. The invokeable stuff feels so javascriptish as well. I don't like the in-transparency of them.

I haven't yet added a type check that will check for the interface intentionally, I hope we might get some kind of more abstract DI than the locators that are always provided via traits. If / when this happens we want to have an interface.

I've added the trait back, but it doesn't do any logic now, it's just there to share the logic (DRY) between the two authenticators and acts as "glue" between the object and the classes, it's basically a getter/locator. I've also refactored some code in FormAuthenticator::authenticate() :)

robertpustulka: Right, on the second thought the interface is OK. You could implement the cake router checker with that :+1:

burzum: I can do that later, but before doing more work here, can we just merge this PR? 😄 

I need to leave for some time and take care of my horse any way.

robertpustulka: I have just one more concern regarding configuring the checker. Since different checkers might require different config, maybe we should extract this into an array and build the checker with this config?

burzum: I don't think the constructor belongs into the interface, you can already pass the config as third arg. Nor would I create a dependency on the ConfigTrait.

src/Authenticator/CookieAuthenticator.php
```diff
@@ -0,0 +1,213 @@
...
+            return [
+                'request' => $request,
+                'response' => $response
+            ];
+        }
+
+        $data = $this->getConfig('cookie');
+        $value = $this->_createToken($identity);
+
+        $cookie = new \Cake\Http\Cookie\Cookie(
```
dereuromark: Those FQCN should be moved to use statement ideally. Inline is usually more a code smell as it hides those things.

src/Authenticator/LoginUrlCheckerInterface.php
robertpustulka: OK, so maybe let's wrap the config into an array and pass as the second argument to the checker.

Since the checker is another object, I don't like how it's config is mixed with authenticator config.

I have something like this in mind:
```php
$service->loadAuthenticator('Form', [
    'loginUrl' => '/some-url',
    'loginUrlChecker' => [
        'className' => CustomChecker::class,
        'customOption' => true
    ]
]);

//internally:
$checker->check('/some-url', [
    'customOption' => true
]);

File: src/Authenticator/CookieAuthenticator.php (hunk 17 above)
Comment (robertpustulka) [reply src/Authenticator/CookieAuthenticator.php]: I wasn't sure how should we handle the class names of "optional" packages.

Comment (robertpustulka) [reply src/Authenticator/CookieAuthenticator.php]: Hmm Symfony uses the `use` statements for potentially missing classes as well.
https://github.com/symfony/dom-crawler/blob/master/Crawler.php#L698

Comment (dereuromark) [reply src/Authenticator/CookieAuthenticator.php]: Those can also be there IMO, if not found those will be ignored :)

Comment (burzum) [reply src/Authenticator/CookieAuthenticator.php]: I wasn't aware that they'll be ignored and to lazy to test it. ☺️

File: src/Authenticator/LoginUrlCheckerInterface.php (hunk 16 above)
Comment (burzum) [reply src/Authenticator/LoginUrlCheckerInterface.php]: I'm OK with passing the array, I don't like yet another locator. Also can we keep the check() taking the request as first arg, so in total the three we already have? Otherwise you would have to put the constructor in the interface.

Comment (robertpustulka) [reply src/Authenticator/LoginUrlCheckerInterface.php]: Current API is ok. I just want to have the ability to pass custom options to the checker. Config array seems like a good solution.

Sorry for the wrong example, of course there should be the request as a first argument.

Comment (burzum) [reply src/Authenticator/LoginUrlCheckerInterface.php]: @robertpustulka no problem. I've changed the config but I'm not happy with it either. A 2nd level array config opens always room for undefined index errors.

File: src/Authenticator/CookieAuthenticator.php (hunk 17 above)
Comment (burzum) [reply src/Authenticator/CookieAuthenticator.php]: Fixed.

File: src/Authenticator/LoginUrlCheckerInterface.php (hunk 16 above)
Comment (robertpustulka) [reply src/Authenticator/LoginUrlCheckerInterface.php]: You can merge options with defaults to ensure that indexes are present

Comment (burzum) [reply src/Authenticator/LoginUrlCheckerInterface.php]: Which means adding the config trait again. :( Feel free to change the code, I'm leaving again, now for my Wing Chun training.

Comment (robertpustulka) [reply src/Authenticator/LoginUrlCheckerInterface.php]: You don’t need that.

Just `$options += $defaults` should do the job.

Comment (burzum) [reply src/Authenticator/LoginUrlCheckerInterface.php]: I'll add that now and we can merge it after that is done?

**Comment by robertpustulka** [comment]
I refactored the code around url checker to make it easier to reuse.

File: src/Authenticator/FormAuthenticator.php [hunk 18]
```diff
@@ -80,17 +80,10 @@ protected function _getData(ServerRequestInterface $request)
...
-            $url = (string)$request->getUri();
-        } else {
-            $url = $request->getUri()->getPath();
-        }
-
         $errors = [
             sprintf(
                 'Login URL `%s` did not match `%s`.',
-                $url,
+                (string)$request->getUri(),
```
burzum: I thought about doing exactly the same, but then the test failed, you just changed the tests. I'm not sure if this is really what we want?

robertpustulka: We can't rely on fact that *every* checker should have full URL check option.

src/UrlChecker/UrlCheckerTrait.php
```diff
@@ -54,7 +55,8 @@ protected function _getLoginUrlChecker()
             $options['className'] = DefaultUrlChecker::class;
         }
 
-        $checker = new $options['className']();
+        $className = App::className($options['className'], 'UrlChecker', 'UrlChecker');
```
burzum: Since we've built the plugin with being useable in other apps / frameworks, will this line work OK outside the context of a CakePHP application? Is it even then aware of the correct paths? I'm not sure, never done it, that's why I'm asking.

robertpustulka: The `cakephp/core` package is required so this will work outside the framework. The support for `Plugin.Class` notation is something we use throughout this plugin. Note that authenticators and identifiers are loaded this way: `$service->loadAuthenticator('Authentication.Form');` etc

robertpustulka:
@burzum I think it's ready to go. Anything else?

robertpustulka:
:tada:

PR #121, opened 2017-05-07
Ensure identity is persisted to session upon successful form auth.

ADmad:


src/Authenticator/TokenAuthenticator.php
```diff
@@ -118,4 +119,14 @@ public function authenticate(ServerRequestInterface $request, ResponseInterface
...
     }
+
+    /**
+     * No-op method.
+     *
+     * @param \Psr\Http\Message\ServerRequestInterface $request A request object.
+     * @return void
+     */
+    public function unauthorizedChallenge(ServerRequestInterface $request)
+    {
```
markstory: Shouldn't this method do something?

src/AuthenticationService.php
```diff
@@ -169,12 +169,8 @@ public function authenticate(ServerRequestInterface $request, ResponseInterface
         foreach ($this->authenticators() as $authenticator) {
             $result = $authenticator->authenticate($request, $response);
             if ($result->isValid()) {
-                if ($authenticator instanceof PersistenceInterface) {
```
markstory: Do we still use `PersisterInterface` anywhere else?

src/Authenticator/TokenAuthenticator.php
ADmad: We could throw a UnauthorizedException. I added it because the PersistenceInterface requires implementing this method. I added PersistenceInterface to TokenAuthenticator so that identity is not persisted to session when authentication is done use TokenAuthenticator / TokenAuthenticator.

src/AuthenticationService.php
ADmad: Currently only SessionAuthenticator implements PersistenceInterface. Probably the WIP CookieAuthenticator will implement it too.

src/Authenticator/TokenAuthenticator.php
burzum: I don't think adding the PersistenceInterface is a good idea because a token is not persistent. Its somewhat missleading IMHO.

markstory: `PersistenceInterface` doesn't require this method, `StatelessInterface` does. Is TokenAuth stateless?

src/AuthenticationService.php
robertpustulka: Yes, Cookie/RememberMe Authenticator will need some identity persistence.

src/Authenticator/TokenAuthenticator.php
robertpustulka: Token authenticator probably should be stateless. It would be used in APIs mostly.
Or we could have another `StatelessTokenAuthenticator` so a developer could decide what they want.

BTW is JWT authentication always stateless?

ADmad: I mistyped, I added `StatelessInterface` not `PersistenceInterface` interface for `TokenAuthenticator` :)

I have only ever seen JWT and custom Token authenticator used in stateless manner so having them implement `StatelessInterface` seems reasonable to me.

robertpustulka: @ADmad so what about this exception?

robertpustulka:
@burzum I don't think that was complete. There's noop method: https://github.com/cakephp/authentication/blob/d30eca1ad00e75f9b11abeebb0577d1249fd8b07/src/Authenticator/TokenAuthenticator.php#L129

ADmad:
@robertpustulka The problem with throwing an exception is you won't be able to do any kind of access control. Let's say you want some routes to be publicly accessible, you won't be able to. That said middleware pipelines would probably solve that problem. So i can thow an exception from the no-op exception it's others agree.

burzum:
@robertpustulka meh, my fault, you're right. :(

PR #122, opened 2017-05-08
Update identity typehints to ArrayAccess|null.

robertpustulka:
I've also split the `getIdentity()` method in the component so we can have stricter typehinting.


dereuromark:
Nice job!

PR #124, opened 2017-05-08
Fix typehints.

robertpustulka:


.travis.yml
```diff
@@ -43,7 +43,7 @@ script:
   - if [[ $DEFAULT = 1 && $TRAVIS_PHP_VERSION != 7.0 ]]; then vendor/bin/phpunit; fi
 
   - if [[ $PHPCS = 1 ]]; then vendor/bin/phpcs -n -p --extensions=php --standard=vendor/cakephp/cakephp-codesniffer/CakePHP ./src ./tests; fi
-  - if [[ $PHPSTAN = 1 ]]; then vendor/bin/phpstan analyse -l 4 src; fi
+  - if [[ $PHPSTAN = 1 ]]; then vendor/bin/phpstan analyse -l 5 src; fi
```
ADmad: Woah! You turned the dial all the way :laughing:

robertpustulka: And it's all green! Locally at least :)

ADmad: I had held back because the readme said level 5 is experimental. But what the heck :)

robertpustulka: We will probably lower it back to 4 when it becomes a PITA, but for now it's green!!

ADmad: Yup.

src/Authenticator/AuthenticatorCollection.php
```diff
@@ -65,8 +65,8 @@ protected function _create($className, $alias, $config)
     /**
      * Resolves authenticator class name.
      *
-     * @param string $class Class name to be resolbed.
-     * @return string
+     * @param string $class Class name to be resolved.
+     * @return string|bool
```
dereuromark: We should not leak the bool then IMO.

robertpustulka: @dereuromark good point. We can change this in the 4.0 version of the core too.

src/Authenticator/AuthenticatorCollection.php
```diff
@@ -65,12 +65,12 @@ protected function _create($className, $alias, $config)
...
      *
-     * @param string $class Class name to be resolbed.
-     * @return string
+     * @param string $class Class name to be resolved.
+     * @return string|null
      */
     protected function _resolveClassName($class)
     {
-        return App::className($class, 'Authenticator', 'Authenticator');
+        return (string)App::className($class, 'Authenticator', 'Authenticator') ?: null;
```
ADmad: Is the string casting necessary? I believe even without it `null` will be returned if `App::className()` returns `false`.

robertpustulka: phpstan says:
> Method Authentication\Authenticator\AuthenticatorCollection::_resolveClassName() should return string|null but returns bool|string|null.

So I've added the type hint.

It's already becoming a PITA :(

josegonzalez: You could check if the response is null and return empty string if it is.

dereuromark: The string cast is not necessary.

ADmad: @robertpustulka Technically phpstan is right since bool means it could also be `true` though our method only return `false`. I would suggest reverting back to level 4 instead of adding unnecessary casting :)

dereuromark: Let's keep the code as simple as possible. with level 4 that seems easy.

ADmad: I doubt it causes any noticeable degradation in performance but I still don't like adding it. We could stay on level 5 and ignore this particular error instead.

josegonzalez: Rather than let you all wax philosophically about which is better or worse, I just added the type check :P

src/Authenticator/AuthenticatorCollection.php
```diff
@@ -65,12 +65,17 @@ protected function _create($className, $alias, $config)
...
-     * @param string $class Class name to be resolbed.
-     * @return string
+     * @param string $class Class name to be resolved.
+     * @return string|null
      */
     protected function _resolveClassName($class)
     {
-        return App::className($class, 'Authenticator', 'Authenticator');
+        $className = App::className($class, 'Authenticator', 'Authenticator') ?: null;
+        if (is_bool($class)) {
```
ADmad: This is equally bad as the casting imo, quite unnecesary :stuck_out_tongue:.  Stop trying to please phpstan :smile:.

josegonzalez: :thisisfine:

dereuromark: Seriouly guys, keeping level to 4 and all is good and simple :)

PR #126, opened 2017-05-09
Update Quick-start-and-introduction.md

cleptric:
cakephp/app's `Application.php` uses the `$middleware` variable name instead of `$middlewareQueue `.


dereuromark:
It should then also be in sync with the docs: https://book.cakephp.org/3.0/en/controllers/middleware.html#using-middleware
It uses $middlewareStack there which is even more correct - as a queue is a term of asyn/deferred execution.

ADmad:
Merging to keep in sync in app skeleton code. If / when it's changed in app skeleton we can change here again too.

PR #129, opened 2017-06-09
Identity object

burzum:


src/Identity.php
```diff
@@ -0,0 +1,140 @@
...
+     */
+    public function get($field)
+    {
+        $map = $this->getConfig('fieldMap');
+        if (isset($map[$field])) {
+            $field = $map[$field];
+        }
+
+        if (isset($this->data[$field])) {
+           return $this->data[$field];
```
burzum: GOD DAMN SPACES... 💩

src/AuthenticationService.php
```diff
@@ -255,4 +256,34 @@ public function getResult()
...
+     * @return \Authentication\IdentityInterface
+     */
+    public function buildIdentity($identityData)
+    {
+        $class = $this->getConfig('identityClass');
+        if (is_callable($class)) {
+            return $class($identityData);
+        }
+
+        return new $class($identityData);
```
markstory: This is a nice approach. Accepting either a class string or a callback factory function is nice :+1:

src/IdentityInterface.php
```diff
@@ -0,0 +1,34 @@
...
+ */
+interface IdentityInterface extends ArrayAccess
+{
+    /**
+     * Constructor
+     *
+     * @param array $identityData Identity data
+     * @param array $config Config options
+     */
+    public function __construct($identityData, array $config = []);
```
markstory: Constructors in the interface can be a bit limiting. I wouldn't include this personally. If people want to vary the constructor they won't be able to.

src/Identity.php
```diff
@@ -0,0 +1,154 @@
...
+ * @license       http://www.opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Authentication;
+
+use Cake\Core\InstanceConfigTrait;
+
+/**
+ * Identity object
+ */
+class Identity implements IdentityInterface
```
ADmad: Shouldn't it implement ArrayAccess too?

burzum: Strange, I thought I did this, I wrote it down. Not sure why it's missing. Going to add it back.

burzum: @ADmad please check the IdentityInterface:

```interface IdentityInterface extends ArrayAccess```

Comment (ADmad) [inline src/Identity.php]: Hehe.. I saw the offsetGet/Set and looked at class declaration only.

**Comment by dereuromark** [comment]
There are quite a few classes and interfaces in the src root dir. I wonder if those maybe could be put into an own namespace, as well.

**Comment by burzum** [comment]
@dereuromark die only two files i would move are the result(interface). Let's move them to the authenticator folder, because this is the only NS they're used in?

**Comment by dereuromark** [comment]
Actually, couldnt AuthenticationService.php and its interface by in its own sub namespace

    /src/Service

as `Authentication\Service\`?

**Comment by burzum** [comment]
Don't think so? If you want to change that as well please do it in another PR :)

File: src/IdentityInterface.php [hunk 5]
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ * Identity interface
+ */
+interface IdentityInterface extends ArrayAccess
+{
+    /**
+     * @return string|int|null
+     */
+    public function getIdentifier();
```
markstory: :+1: looks good.

PR #130, opened 2017-06-11
Fix up docs.

markstory:
Fixed some grammar errors and re-ordered the sections as you'd likely create an Identity before you could use it.

### DIFF
--- a/docs/Identity-Object.md
+++ b/docs/Identity-Object.md
@@ -2,7 +2,7 @@
 
 The identity object is returned by the service and made available in the request. The object provides a method `getIdentifier()` that can be called to get the id of the current log in identity.
 
-The reason this object exists is to provide an interface that makes it easy to get access to the identity's across various implementations/sources.
+The reason this object exists is to provide an interface that makes it easy to get access to the identity's id across various implementations/sources.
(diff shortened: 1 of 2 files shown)


docs/Identity-Object.md
```diff
@@ -2,7 +2,7 @@
 
 The identity object is returned by the service and made available in the request. The object provides a method `getIdentifier()` that can be called to get the id of the current log in identity.
 
-The reason this object exists is to provide an interface that makes it easy to get access to the identities id no matter what it is.
+The reason this object exists is to provide an interface that makes it easy to get access to the identity's across various implementations/sources.
```
ADmad: Should be `identity's id`

burzum:
@markstory release another beta after these doc changes are merged?

I would like to go RC but the cookie stuff is missing. To make progress here is my suggestion: Lets go for the CakePHP Http interface for now, create an issue as reminder to add pure PSR7 support to it later, so it won't be forgotten.

docs/Identity-Object.md
markstory: Fixed now. Thanks.

PR #131, opened 2017-07-26
Bcrypt cost changing should require rehashing.

markstory:
Changing the cost option in bcrypt hashing should trigger a rehash.
Porting changes from cakephp/cakephp#10951


cleptric:
Thanks 🙂

Issue #132, opened 2017-08-08
[RFC] Identity abstraction


robertpustulka:
IIRC we decided that `ArrayAccess` for identity abstraction would suffice. It would allow for using `EntityInterface` instances as identities.

I noticed that there's still `IdentityInterface` and `Identity` class in use. It looks like a wrapper for a `ArrayAccess` an actual identity returned from a successfull authenticator result.

I'm wondering if that's even necessary given the fact, that the actual identity passed to the request parameter will be an `IdentityInterface` instance rather than an entity or whatever returned from an identity resolver. The whole idea of having flexible interface (`ArrayAccess`) for identities falls through. The `IdentityInterface` doesn't even have a method to access the original identity.

My idea is to get rid of `IdentityInterface` and `Identity` at all and use actual identities as request parameters. The interface and the class look reduntant to me.

markstory:
The Identity class implements `ArrayAccess`. Having a generic way to get an identities 'id' is the value that class provides. If we were to use 'bare' entities, we would not be able to reliably find the id field as it could have arbitrary names based on the developer's schema.

robertpustulka:
Is that needed somewhere to get the identity `id`?

If that's a convenience for the userland code I think that the developer is aware of the schema their Users model is using and there's no real benefit of having such wrappers.

It's actually limiting, because the identity no longer implements the EntityInterface if ORM resover was used.

markstory:
> Is that needed somewhere to get the identity id?

I think it is. Wouldn't we need a consistent way to get the id when making application code that is agnostic of the identifier/resolver?

> It's actually limiting, because the identity no longer implements the EntityInterface if ORM resover was used.

If your application uses ldap & local ORM identifiers, you wouldn't be able to rely on an EntityInterface either.

robertpustulka:
OK, but the method to get the original data from the identity is required IMO.

markstory:
Won't [the ArrayAccess](https://github.com/cakephp/authentication/blob/master/src/IdentityInterface.php#L20) extension allow for that?

robertpustulka:
That won't allow me to access custom methods defined in my `User` entity. I hit a wall while developing my app. I was certain i have an instance of my `User` class in a request attribute.

I remember that the decision to make identities `ArrayAccess` was made on ordee to allow people to use their entity classes optionally as one of the complaints about `AuthComponent` was the use of raw arrays. Having array in a form of an object wrapper does not resolve the issue.
[...]

PR #133, opened 2017-08-09
Some improvements around identity.

robertpustulka:
Refs #132

### DIFF
--- a/src/Identity.php
+++ b/src/Identity.php
@@ -45,20 +45,20 @@ class Identity implements IdentityInterface
     /**
      * Constructor
      *
-     * @param array|\ArrayAccess $identityData Identity data
+     * @param array|\ArrayAccess $data Identity data
      * @param array $config Config options
      * @throws InvalidArgumentException When invalid identity data is passed.
      */
-    public function __construct($identityData, array $config = [])
+    public function __construct($data, array $config = [])
     {
-        if (!is_array($identityData) && !$identityData instanceof ArrayAccess) {
-            $type = is_object($identityData) ? get_class($identityData) : gettype($identityData);
-            $message = sprintf('Array data must be an `array` or implement `ArrayAccess` interface, `%s` given.', $type);
+        if (!is_array($data) && !$data instanceof ArrayAccess) {
+            $type = is_object($data) ? get_class($data) : gettype($data);
+            $message = sprintf('Identity data must be an `array` or implement `ArrayAccess` interface, `%s` given.', $type);
             throw new InvalidArgumentException($message);
         }
 
         $this->setConfig($config);
-        $this->data = $identityData;
+        $this->data = $data;
     }
 
     /**
--- a/src/Identity.php
+++ b/src/Identity.php
@@ -14,6 +14,7 @@ namespace Authentication;
 
 use ArrayAccess;
 use Cake\Core\InstanceConfigTrait;
+use InvalidArgumentException;
 
 /**
  * Identity object
@@ -37,18 +38,25 @@ class Identity implements IdentityInterface
     /**
      * Identity data
      *
-     * @var \ArrayAccess
+     * @var array|\ArrayAccess
      */
     protected $data;
 
     /**
      * Constructor
      *
-     * @param \ArrayAccess $data Identity data.
-     * @param array $config Config options.
+     * @param array|\ArrayAccess $data Identity data
+     * @param array $config Config options
+     * @throws InvalidArgumentException When invalid identity data is passed.
      */
-    public function __construct(ArrayAccess $data, array $config = [])
+    public function __construct($data, array $config = [])
     {
+        if (!is_array($data) && !$data instanceof ArrayAccess) {
+            $type = is_object($data) ? get_class($data) : gettype($data);
+            $message = sprintf('Identity data must be an `array` or implement `ArrayAccess` interface, `%s` given.', $type);
+            throw new InvalidArgumentException($message);
+        }
+
         $this->setConfig($config);
         $this->data = $data;
     }
--- a/src/Identity.php
+++ b/src/Identity.php
@@ -12,7 +12,9 @@
  */
 namespace Authentication;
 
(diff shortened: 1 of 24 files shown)


robertpustulka:
No checks from travis?

ADmad:
Strange.. I pushed a commit to master and travis build was triggered.

ADmad:
The good 'ole close and reopen PR trick!

robertpustulka:
Force push also retriggered the build. But shouldn't there be 2 of them?

ADmad:
Nope, you made the `identity` branch on your fork.

robertpustulka:
I found another bug, where the AuthenticationMiddleware would set the identity from a result (raw data) not the IdentityInterface object.

Should I fix it here or open another PR?

robertpustulka:
I discovered that the identity object and identity data concepts were mixed up around the code.
That required updating tests to assure the `IdentityInterface` instances are attached to the request attribute.

`ResultInterface::getIdentity()` actually returns the identity data retrieved from identifiers, not the `IdentityInterface` object. The object is created by the `AuthenticationService`. The middleware attached the raw data to the request not the actual identity object. In order to fix this I had to expose `AuthenticationServiceInterface::getIdentifier()` method and use that to attach the request attribute.


I believe the name of method `ResultInterface::getIdentity()` is confusing. It does not get the identity, it gets the identity data. Should we rename the method to `getData` or `getIdentityData` to make the distinction clearer?

markstory:
>  Should we rename the method to getData or getIdentityData to make the distinction clearer?

Good find on that inconsistency. `getData()` make sense to me.

src/Identity.php
```diff
@@ -36,18 +38,25 @@ class Identity implements IdentityInterface
...
-     * @param array $identityData Identity data
+     * @param array|\ArrayAccess $identityData Identity data
      * @param array $config Config options
+     * @throws InvalidArgumentException When invalid identity data is passed.
      */
     public function __construct($identityData, array $config = [])
     {
+        if (!is_array($identityData) && !$identityData instanceof ArrayAccess) {
+            $type = is_object($identityData) ? get_class($identityData) : gettype($identityData);
+            $message = sprintf('Array data must be an `array` or implement `ArrayAccess` interface, `%s` given.', $type);
```
markstory: Should this be `identityData must be an array...`

robertpustulka:
Identity data can only be an instance of `ArrayAccess` so there's no need to support arrays.

`Identity::toArray()` has been refactored to `getOriginalData()`. It's not easy to convert every implementation of `ArrayAccess` into an array.
Checking for existance of `toArray()` method and typecasting otherwise seems like a bad idea. I decided that having a getter for the original data object is enough.

src/Identity.php
```diff
@@ -134,11 +135,11 @@ public function offsetUnset($offset)
...
 
     /**
-     * Turns the object into an array
+     * Gets the original data object.
      *
-     * @return array
+     * @return \ArrayAccess
      */
-    public function toArray()
+    public function getOriginalData()
```
markstory: :+1: to this. It is a fair bit safer to use.

ADmad:
This makes me think what's the benefit of having identity data be an `ArrayAccess` instance?

Included identifiers return Cake' `Entity` instance so we can just use their properties directly and for external providers too you are rarely going to get objects which implement array access. It's lot more likely to get simple object of type stdclass with properties. E.g [1](https://github.com/SocialConnect/common/blob/master/src/Entity/User.php), [2](https://github.com/hybridauth/hybridauth/blob/master/src/User/Profile.php).

robertpustulka:
We force the identity data to be `ArrayAccess` in order to easily retrieve the properties in `Identity` class, including the property for `getIdentifier()` method. You're right that this requirement may over complicate things in certain scenarios (for example when working with external libs). Someone might need to write a wrapper implementing the `ArrayAccess` interface for their internal identity classes so the `Identity` class can access the `id` property. The shorter way would be to return the `IdentityInterface` instance from the custom resolver, but still there's a requirement of `ArrayAccess`.

I think we could drop the requirement of `ArrayAccess` from `IdentityInterface` and expose a `getOriginal()` method within this interface. That may simplify things. 

We would need to remove any `ArrayAcces` type hints from where the identity data is concerned. The return data for `IdentifierInterface::identify()` and `ResultInterface::getData()` would need to be `mixed` as well.

The default `Identity` class could still require an `ArrayAccess` in it's constructor to easily blend in with CakePHP ORM. Custom resolvers could return `IdentifierInstance` so the default `Identity` class does not get involved. Or the developer could provide their own identity class to the authentication service.

markstory:
Having `IdentityInterface` extend `ArrayAccess` lets us provide more features to users of the plugin without being an onerous or difficult to implement interface.

robertpustulka:
@markstory OK right. So how about identity data and the requirement of `ArrayAccess`? Should we be a bit more liberal about it and allow regular arrays as well? We could allow any kind of data as well, but I'm not sure how  the data would be converted to the `Identity` in that case...

markstory:
>  So how about identity data and the requirement of ArrayAccess? Should we be a bit more liberal about it and allow regular arrays as well?

I think the identity data being an array or ArrayAccess will be fine.

src/Identity.php
```diff
@@ -37,18 +38,25 @@ class Identity implements IdentityInterface
...
-     * @param array $config Config options.
+     * @param array|\ArrayAccess $data Identity data
+     * @param array $config Config options
+     * @throws InvalidArgumentException When invalid identity data is passed.
      */
-    public function __construct(ArrayAccess $data, array $config = [])
+    public function __construct($data, array $config = [])
     {
+        if (!is_array($data) && !$data instanceof ArrayAccess) {
+            $type = is_object($data) ? get_class($data) : gettype($data);
```
ADmad: We should have a function in core itself for this :)

robertpustulka:
Is there anything to be done here?

PR #134, opened 2017-08-09
Run tests on PHP 7.2 also.

ADmad:


ADmad:
@markstory Can you please do the magic related to session/init_set() you did for core to get the build passing on 7.2? :smile:

markstory:
@ADmad That magic will be in the next release of 3.4.

PR #135, opened 2017-08-25
Use constant time comparision

markstory:
Prevent timing attacks by using a constant time comparison.

Port of cakephp/cakephp#11096

### DIFF
--- a/src/Authenticator/HttpDigestAuthenticator.php
+++ b/src/Authenticator/HttpDigestAuthenticator.php
@@ -100,7 +100,7 @@ class HttpDigestAuthenticator extends HttpBasicAuthenticator
         }
 
         $hash = $this->generateResponseHash($digest, $password, $server['ORIGINAL_REQUEST_METHOD']);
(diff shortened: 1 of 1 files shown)


src/Authenticator/HttpDigestAuthenticator.php
```diff
@@ -100,7 +100,7 @@ public function authenticate(ServerRequestInterface $request, ResponseInterface
         }
 
         $hash = $this->generateResponseHash($digest, $password, $server['ORIGINAL_REQUEST_METHOD']);
-        if ($digest['response'] === $hash) {
+        if (hash_equals($digest['response'], $hash)) {
```
ceeram: arguments should be swapped

ceeram:
```Note:
It is important to provide the user-supplied string as the second parameter, rather than the first.```

http://php.net/manual/en/function.hash-equals.php

markstory: Thanks I missed that.

PR #136, opened 2017-08-27
Updating DigestAuthenticator to match CakePHP core updates.

markstory:
Shore up some of the implementation gaps we have around nonces. Nonces should expire and should be generated by the server. I've followed the Symfony approach to generating nonces which results in short lived (5 minutes by default) nonces that are regenerated each time the client receives a 401.

By validating and requiring nonce rotation we can avoid replay attacks with our Digest authentication implementation. This replicates changes made in cakephp/cakephp#9668

Adopt constant time string comparisons to prevent timing attacks.


markstory:
This also replicates changes made in cakephp/cakephp#11103

Issue #137, opened 2017-09-19
FormAuthenticator::_checkLoginUrl() fails when accessing alternative route


burzum:
For testing I've been accessing the direct controller/action URL `http://cake3.world-architects.com/en/users/login` *but* we have a route for that as well `/en/login`.

But the FormAuthenticator compares *only* a string URL and not if the passed array really matches the resolved controller action but instead only the first route that machtes:

```
\vendor\cakephp\authentication\src\Authenticator\FormAuthenticator.php (line 71)
'/en/users/login'
\vendor\cakephp\authentication\src\Authenticator\FormAuthenticator.php (line 72)
'/en/login'
```

I'm not sure if there is a better way than to compare the arrays to resolve this?

markstory:
I think you would have to compare array values as well.

burzum:
@markstory exactly my thought but then it is required that you run the routing middleware before authentication. But this brings up a problem if we want to do authorization against routes. authorization requires the identity.

markstory:
Couldn't the 'loginAction' be defined as a string URL if people have multiple possible routes a login page can be reached at?

burzum:
We would always require a string then and turn the URI from the request also into an array and then compare them. Or am I missing something? If not I'll do the change.

markstory:
That sounds like it should work to me.

burzum:
@markstory does that look OK?

```php
    protected function _checkLoginUrl(ServerRequestInterface $request)
    {
        $loginUrl = $this->getConfig('loginUrl');

        if (!empty($loginUrl)) {
            $requestUrl = Router::parseRequest($request);

            if (is_string($loginUrl)) {
                $loginUrl = Router::parseRequest((new ServerRequest([
                    'uri' => $loginUrl
                ])));
                $this->setConfig('loginUrl', $loginUrl);
            }

            $keysToCompare = array_keys($loginUrl);

            foreach ($keysToCompare as $key) {
                if (!isset($requestUrl[$key])
                    || $requestUrl[$key] !== $loginUrl[$key]
                ) {
                    return false;
                }
            }

        }

        return true;
    }
```

markstory:
You could use `Hash::diff` or `array_key_diff()` to check the difference between the two URL arrays.

burzum:
@markstory I've pushed the code to the branch `bug/check-login-url` there is an issue with the routing it seems, I'm not sure if we simply want to match against the URL it has generated in this case. Would you mind to look at the code?

[...]

PR #146, opened 2017-11-06
Attempt to fix the login url comparison

burzum:
Refs https://github.com/cakephp/authentication/issues/137

### DIFF
--- a/src/Authenticator/FormAuthenticator.php
+++ b/src/Authenticator/FormAuthenticator.php
@@ -12,6 +12,7 @@
  */
 namespace Authentication\Authenticator;
 
+use Cake\Http\ServerRequest;
 use Cake\Routing\Router;
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
@@ -61,11 +62,15 @@ class FormAuthenticator extends AbstractAuthenticator
     public function authenticate(ServerRequestInterface $request, ResponseInterface $response)
     {
         if (!$this->_checkLoginUrl($request)) {
+            $loginUrl = $this->getConfig('loginUrl');
+            if (is_array($loginUrl)) {
+                $loginUrl = Router::url($loginUrl, true);
+            }
             $errors = [
                 sprintf(
                     'Login URL %s did not match %s',
                     $request->getUri()->getPath(),
-                    $this->getConfig('loginUrl')
+                    $loginUrl
                 )
             ];
 
@@ -100,12 +105,26 @@ class FormAuthenticator extends AbstractAuthenticator
         $loginUrl = $this->getConfig('loginUrl');
 
         if (!empty($loginUrl)) {
-            if (is_array($loginUrl)) {
-                $loginUrl = Router::url(['_base' => false] + $loginUrl);
+            $requestUrl = Router::parseRequest($request);
+            unset($request['_matchedRoute']);
+
+            if (is_string($loginUrl)) {
+                $loginUrl = Router::parseRequest((new ServerRequest([
+                    'uri' => $loginUrl
+                ])));
+
+                unset($loginUrl['_matchedRoute']);
                 $this->setConfig('loginUrl', $loginUrl);
             }
 
-            return strcasecmp($request->getUri()->getPath(), $loginUrl) === 0;
+            $keysToCompare = array_keys($loginUrl);
+            foreach ($keysToCompare as $key) {
+                if (!array_key_exists($key, $requestUrl)
+                    || $requestUrl[$key] !== $loginUrl[$key]
+                ) {
+                    return false;
+                }
+            }
         }
 
         return true;
--- a/src/Authenticator/FormAuthenticator.php
+++ b/src/Authenticator/FormAuthenticator.php
@@ -13,7 +13,7 @@
 namespace Authentication\Authenticator;
 
 use Cake\Http\ServerRequest;
-use Cake\Routing\Router;
+use InvalidArgumentException;
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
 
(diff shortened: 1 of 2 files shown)


markstory:
Given that we can only confidently have string URLs to compare with loginAction, should we accept a list of URLs instead of a routing array? Something like 

```php
['loginAction' => ['/users/login', '/en/login']]
```

We could also support syntax like `/*`  for URL patterns with trailing arguments. This is a bit of a departure from AuthComponent, but helps decouple this from Routing.

burzum:
@markstory could we support at least regex as well then?

We have 18 languages and your suggestion will enforce that somebody doesn't forgot to add a new language there when needed.

I could change the code in a way it supports the router but only when it's present, so this package won't have a hard dependency on it.

dereuromark:
Can we have a few test cases for this scenario maybe?

burzum:
@dereuromark before I waste time and it gets rejected *after* I wrote a test I would prefer to get an opinion if this is going to be OK?

https://gist.github.com/burzum/efe49d06fc928dec4955ad2291c7be81

src/Authenticator/FormAuthenticator.php
```diff
@@ -100,12 +105,26 @@ protected function _checkLoginUrl(ServerRequestInterface $request)
...
 
         if (!empty($loginUrl)) {
-            if (is_array($loginUrl)) {
-                $loginUrl = Router::url(['_base' => false] + $loginUrl);
+            $requestUrl = Router::parseRequest($request);
+            unset($request['_matchedRoute']);
+
+            if (is_string($loginUrl)) {
+                $loginUrl = Router::parseRequest((new ServerRequest([
+                    'uri' => $loginUrl
```
markstory: I think you want `url` not `uri`.

src/Authenticator/FormAuthenticator.php
```diff
@@ -100,12 +105,26 @@ protected function _checkLoginUrl(ServerRequestInterface $request)
         $loginUrl = $this->getConfig('loginUrl');
 
         if (!empty($loginUrl)) {
-            if (is_array($loginUrl)) {
-                $loginUrl = Router::url(['_base' => false] + $loginUrl);
+            $requestUrl = Router::parseRequest($request);
+            unset($request['_matchedRoute']);
+
+            if (is_string($loginUrl)) {
+                $loginUrl = Router::parseRequest((new ServerRequest([
```
markstory: Extra `(` here.

robertpustulka:
Regex seems ok as well the explicit array support. Relying on router would probably mean that the authentication middleware must be always added after the routing middleware so the routes are build before the check made by the authenticator.

src/Authenticator/FormAuthenticator.php
```diff
@@ -64,8 +64,9 @@ public function authenticate(ServerRequestInterface $request, ResponseInterface
         if (!$this->_checkLoginUrl($request)) {
             $loginUrl = $this->getConfig('loginUrl');
             if (is_array($loginUrl)) {
-                $loginUrl = Router::url($loginUrl, true);
+                $loginUrl = \Cake\Routing\Router::url($loginUrl, true);
             }
+            dd($loginUrl);
```
dereuromark: dd

src/Authenticator/FormAuthenticator.php
```diff
@@ -12,7 +12,8 @@
  */
 namespace Authentication\Authenticator;
 
-use Cake\Routing\Router;
+use Cake\Http\ServerRequest;
```
robertpustulka: Looks like `cakephp/cakephp` dependency is required for this class to be loaded.

I'm still not convinced that this should rely on HTTP and/or Routing packages. Could we just support string urls as @markstory suggested? `loginUrl` option could accept an explicit url string (like `/login`), an array of urls (`['/login', '/users/login']`) or a regex (`%^/[a-z]+/login$%`).

burzum: @robertpustulka would you mind doing the change then? I've explicitly asked before I started on the change but didn't get any clear yes or no on this... So please change it as you like until it fits what you want.

burzum:
@markstory @robertpustulka any idea why the URL is wrong in the test? The routing setup seems to be right?

src/Authenticator/FormAuthenticator.php
robertpustulka: @burzum Sorry, I may have not been clear on this here: https://github.com/cakephp/authentication/pull/146#issuecomment-346813351

I’ll take a look on this tomorrow.

src/Authenticator/FormAuthenticator.php
```diff
@@ -100,14 +105,76 @@ protected function _checkLoginUrl(ServerRequestInterface $request)
...
         if (!empty($loginUrl)) {
-            if (is_array($loginUrl)) {
-                $loginUrl = Router::url(['_base' => false] + $loginUrl);
-                $this->setConfig('loginUrl', $loginUrl);
+            if (is_string($loginUrl) && $this->_compareStringUrl($loginUrl, $request->getUri()->getPath())) {
+                return true;
             }
 
-            return strcasecmp($request->getUri()->getPath(), $loginUrl) === 0;
+            return $this->_compareArrayUrl($loginUrl, $request);
```
markstory: Do we need array URL support? I'd like to believe that simple string URLs and regex patterns should cover the bulk of use cases.

src/Authenticator/FormAuthenticator.php
```diff
@@ -100,14 +105,76 @@ protected function _checkLoginUrl(ServerRequestInterface $request)
...
+    {
+        if (!is_string($loginUrl)) {
+            throw new InvalidArgumentException();
+        }
+
+        if ($loginUrl === $requestUrl) {
+            return true;
+        }
+
+        $regex = $this->getConfig('loginUrlRegex');
```
markstory: Having a separate config option is a good idea :+1: Better than using `preg:` prefix as we've done in other places in the past.

src/Authenticator/FormAuthenticator.php
robertpustulka: > Do we need array URL support?

I don't think we do. 

I can handle string URL approach if you @burzum and @markstory don't mind.

PR #149, opened 2017-11-29
Login URL comparison fix alternative

robertpustulka:
Alternative for #146, refs #137

Features added:

- multiple URL candidate support
- regex support
- full base URI comparison support (eg. to enable login only on subdomain)


src/Authenticator/FormAuthenticator.php
```diff
@@ -97,17 +121,32 @@ public function authenticate(ServerRequestInterface $request, ResponseInterface
...
+        } else {
+            $url = $request->getUri()->getPath();
         }
 
-        return true;
+        if ($this->getConfig('useRegex')) {
+            $check = 'preg_match';
+        } else {
+            $check = function ($loginUrl, $url) {
+                return strcasecmp($loginUrl, $url) === 0;
```
ADmad: I would suggest doing a case sensitive comparison, since technically URLs with different casing are unique.

src/Authenticator/FormAuthenticator.php
```diff
@@ -97,17 +121,32 @@ public function authenticate(ServerRequestInterface $request, ResponseInterface
...
-            }
+        if (empty($loginUrls)) {
+            return true;
+        }
 
-            return strcasecmp($request->getUri()->getPath(), $loginUrl) === 0;
+        if ($this->getConfig('checkFullUri')) {
+            $url = (string)$request->getUri();
+        } else {
+            $url = $request->getUri()->getPath();
```
markstory: Should this block become a separate method?

burzum:
@markstory it's OK but I bet on that somebody will try to use a Cake style array for an URL for the login and open a ticket for it. ;)

dereuromark:
Wait, did we completely use array URLS here now for the login URL? Or only for the regexp comparison etc (the case mentioned here)? The latter hopefully. This could then also be mentioned in the doc block above, that in that case a string URL is required maybe?

Issue #150, opened 2017-11-29
SessionAuthenticator identify option does not work.


robertpustulka:
I set up authentication as below:
```php
    public function authentication(AuthenticationServiceInterface $authentication)
    {
        $authentication->loadIdentifier('Authentication.Password');

        $authentication->loadAuthenticator('Authentication.Session', [
            'identify' => true
        ]);
        $authentication->loadAuthenticator('Authentication.Form');

        return $authentication;
    }
```

Identity verification won't work. `SessionAuthenticator` assumes that the data persisted in session provides a plain text password: https://github.com/cakephp/authentication/blob/master/src/Authenticator/SessionAuthenticator.php#L60-L63

We need an alternative approach to reidentify the identity since we can't expect that stored identity data will provide plain text password for identifiers. 

This will be needed for the `CookieAuthenticator` as well.

robertpustulka:
I was wondering if we could have a simple `IdIdentifier` for identity reverification?

This way persistence authenticators could check if an identity with given id still exist in a database (isn't banned or something).

robertpustulka:
Resolved as a part of #153

PR #151, opened 2017-11-29
Add VerificationIdentifier

robertpustulka:
This is a identifier which can be used by authenticators that implement `PersistenceInterface` (Session, Cookie) in order to verify that the persisted identity is still valid.

Fix attempt for #150

### DIFF
--- a/src/Identifier/VerificationIdentifier.php
+++ b/src/Identifier/VerificationIdentifier.php
@@ -1,0 +1,54 @@
+<?php
+/**
+ * CakePHP(tm) : Rapid Development Framework (http://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (http://cakefoundation.org)
+ * Licensed under The MIT License
+ * For full copyright and license information, please see the LICENSE.txt
+ * Redistributions of files must retain the above copyright notice.
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (http://cakefoundation.org)
+ * @link          http://cakephp.org CakePHP(tm) Project
+ * @license       http://www.opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Authentication\Identifier;
+
+use Authentication\Identifier\Resolver\ResolverAwareTrait;
+
+/**
+ * Identity verification identifier
+ */
+class VerificationIdentifier extends AbstractIdentifier
+{
+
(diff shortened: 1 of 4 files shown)


robertpustulka:
With this change in order for `identify` option to work the `Verification` identifier needs to be loaded or a token based identification could be leveraged if implemented.

For instance:
```php
$service->loadAuthenticator('Authentication.Session', [
    'identify' => true
]);
$service->loadIdentifier('Authentication.Verification`);
```

Or:
```php
$service->loadAuthenticator('Authentication.Token');
$service->loadAuthenticator('Authentication.Session', [
    'fields' => ['token' => 'api_key'],
    'identify' => true
]);
    'tokenField' => 'api_key'
]);
```

src/Identifier/VerificationIdentifier.php
```diff
@@ -0,0 +1,54 @@
...
+ * @license       http://www.opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Authentication\Identifier;
+
+use Authentication\Identifier\Resolver\ResolverAwareTrait;
+
+/**
+ * Identity verification identifier
+ */
+class VerificationIdentifier extends AbstractIdentifier
```
markstory: How would this work in practice with other identifiers? If I made an identifier collection with this and PasswordIdentifier, couldn't I become any user by 'logging in' with just an id?

robertpustulka: Only authenticators have access to identifiers so if an authenticator pass the user provided id with the `id` data key to the identifiers it could log in as any user. But currently there’s no such authenticator. FormAuthenticator uses `username/password` data keys and Token operate on `token` key.

markstory: Ok. But right now FormAuthenticator passes the entire request body into the identifiers. https://github.com/cakephp/authentication/blob/0ecd87e1cdd22719c30c43b3327b63814de8e3dc/src/Authenticator/FormAuthenticator.php#L82-L83

Which is why I'm concerned.

robertpustulka: You are right, this should be fixed.

robertpustulka: PR is open #153

robertpustulka:
The same can be achieved using `PasswordIdentifier` and password-less identification. Closing this then.

PR #153, opened 2017-11-30
Form, JWT, Password security improvements.

robertpustulka:
I made some improvements around form/password authentication security.

Formerly the entire body data was passed to `identify()` method. This allowed malicious users pass additional data and possibly authenticate using identifiers that didn't require password. Now only the explicitly configured fields are passed to the identifier check.

`PasswordIdentifier` was fixed to skip password check only if a password key is missing from data. The password check is always performed even if an identity has not been found by resolvers.

I also fixed some incosistencies around username/password data keys.
`FormAuthenticator` will look for configured fields in body data and pass them to identifiers using field keys. 

For example:
```php
//authentication
$service->loadAuthenticator('Form', [
  'fields' => [
    'username' => 'user', 
    'password' => 'secret'
  ]
]);
$service->loadIdentifier('Password', [
  'fields' => [
    'username' => ['user_name', 'user_email'], 
    'password' => 'user_password'
  ]
]);

//body
[
  'user' => 'mail@example.com',
  'secret' => 'mypass',
  'other' => 'value'
];

//data passed to identifiers:
[
  'username' => 'mail@example.com',
  'password' => 'mypass'
]

//resolver find conditions
['OR' => [
  'user_name' => 'mail@example.com',
  'user_email' => 'mail@example.com'
]];
```
The "reserved" pair of identifer data keys helps to keep a stricter API, decouple `FormAuthenticator` from `PasswordIdentifier` and avoid key collisions with other authenticators/identifiers.


robertpustulka:
Since this PR allows for a more secure password-less identification I updated the `SessionAuthenticator` to leverage this feature with a possible fix for #150.

src/Authenticator/FormAuthenticator.php
```diff
@@ -28,25 +28,28 @@ class FormAuthenticator extends AbstractAuthenticator
      * Checks the fields to ensure they are supplied.
      *
      * @param \Psr\Http\Message\ServerRequestInterface $request The request that contains login information.
-     * @param array $fields The fields to be checked.
-     * @return bool False if the fields have not been supplied. True if they exist.
+     * @return array|null False if the fields have not been supplied. True if they exist.
```
markstory: The types and explanation don't match anymore.

src/Identifier/PasswordIdentifier.php
```diff
@@ -88,61 +88,58 @@ public function getPasswordHasher()
...
-            $hashedPassword = $result[$passwordField];
-            if (!$hasher->check($password, $hashedPassword)) {
-                return null;
-            }
-
-            $this->_needsPasswordRehash = $hasher->needsRehash($hashedPassword);
-            unset($result[$passwordField]);
+        $passwordField = $this->getConfig('fields.password');
+        $hasher = $this->getPasswordHasher();
+        $hashedPassword = $identity[$passwordField];
```
markstory: We would still need a `passwordField` configuration option if we were to make an 'interface' as I had proposed. :cry:

robertpustulka:
@markstory That’s what I’m trying to achieve here.
The interface for authenticator-identifier communication. Configuration would be required only if form fields or db fields are not standard username/password.

I left the option for FormAuthenticator „data key” configuration only to make it more flexible to work with other kinds of identifiers (that would expect a different pair or number of keys)

markstory:
@robertpustulka If we have `username`, `password` and `token` as the interface to `identifier::identify()` do we need fields configuration there as well?

robertpustulka:
@markstory yes, because the `username` can be matched to `email` or `user_name` or anything else in a db.

Let's say we have a following login form:
```html
<input type=text name=user />
<input type=password name=secret />
```

And a following DB table:
```php
$table
  ->addColumn('email', 'string')
  ->addColumn('pass', 'string')
```

We need to configure `FormAuthenticator` to look for `user` and `secret` form fields so it would pass the correct values to identifiers using "interfaced" `username` and `password` keys.

Then we need to configure `PasswordIdentifiers` as well to match `username` value to `email` column and `password` value to `pass` column.

```php
//form fields config
$service->addAuthenticator('Form', [
  'fields' => ['username' => 'user', 'password' => 'secret']
]);

//db fields config
$service->addIdentifier('Password', [
  'fields' => ['username' => 'email', 'password' => 'pass']
]);
```

robertpustulka:
@markstory would it make sense to create constants for those "interface" keys?

Like `IdentifierInterface::CREDENTIAL_USERNAME` etc.

markstory:
@robertpustulka Thank you for explaining why we need the configuration in both places. I was being dense and forgot about needing to map the interface fields onto the correct columns when creating conditions. 😊 

Constants would help catch typos/mistakes earlier. I'm :+1: for having them.

src/Authenticator/SessionAuthenticator.php
```diff
@@ -57,10 +58,11 @@ public function authenticate(ServerRequestInterface $request, ResponseInterface
...
         if ($this->getConfig('identify') === true) {
-            $user = $this->identifiers()->identify([
-                'username' => $user[$this->getConfig('fields')['username']],
-                'password' => $user[$this->getConfig('fields')['password']]
-            ]);
+            $credentials = [];
+            foreach ($this->getConfig('fields') as $key => $field) {
+                $credentials[$key] = $user[$field];
+            }
+            $user = $this->identifiers()->identify($credentials);
```
markstory: Identifying users from the session on username only looks good now that FormAuthenticator whitelists its fields :+1:

robertpustulka:
@markstory "data keys" now use interface constants.

robertpustulka:
I found out that JWT authenticator passed the entire payload to the identifiers as well. Fixed that to make sure that only `sub` key is passed.

src/Identifier/PasswordIdentifier.php
```diff
@@ -88,63 +88,61 @@ public function getPasswordHasher()
...
-            unset($result[$passwordField]);
+        $hasher = $this->getPasswordHasher();
+        $hashedPassword = $identity[$passwordField];
+        if (!$hasher->check($password, $hashedPassword)) {
+            return false;
         }
 
-        return $result;
+        $this->_needsPasswordRehash = $hasher->needsRehash($hashedPassword);
+        unset($identity[$passwordField]);
```
markstory: Why is this unset() necessary?

robertpustulka: @markstory I'm not sure. It's been there before. 

@burzum any idea?

burzum: I guess to empty the password field in case of a form? No idea other than that. Just remove it and see if a test breaks. Is the rehashing tested btw?

robertpustulka: Some tests fail as they expect that the password field is missing. I'm still not sure why it's being unset. I'll remove this, OK?

burzum: OK

markstory:
Looks good to me.

Issue #156, opened 2017-12-12
Composer PHPUnit incompatible with framework


sdustinh:
The plugins PHPUnit is pinned to `<6.0` which is incosistent with the [framework](https://github.com/cakephp/cakephp/blob/master/composer.json#L44) which makes this plugin hard to install in applications that support `>=6.0`. I'm not sure if this was planned, just thought I'd throw an issue in since the auth component [deprecations](https://github.com/cakephp/cakephp/wiki/3.6-Roadmap#http-middleware-improvements) are en route.

robertpustulka:
Good catch. The PR is open.

PR #157, opened 2017-12-12
Composer.json updates

robertpustulka:
A number of changes to composer.json. Keep things in line with the core composer.json file.

Refs #156

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -38,7 +38,7 @@
             "AuthTest\\": "tests/test_app/Plugin/AuthTest/src/",
             "Authentication\\Test\\": "tests/",
             "Authentication\\Test\\App\\": "tests/test_app/App/",
-            "Cake\\Test\\": "./vendor/cakephp/cakephp/tests/",
+            "Cake\\Test\\": "vendor/cakephp/cakephp/tests/",
             "TestApp\\": "tests/test_app/TestApp/",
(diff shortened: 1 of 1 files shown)


dereuromark:
Can we also have trailing slashes on the psr parts? "src" => "src/" etc?
That would be in sync with core and the rest of the eco system then at some point.

composer.json
```diff
@@ -29,29 +30,39 @@
...
-            "AuthTest\\": "tests/test_app/Plugin/AuthTest/src",
-            "Authentication\\Test\\": "tests",
-            "Authentication\\Test\\App\\": "tests/test_app/App",
-            "Cake\\Test\\": "./vendor/cakephp/cakephp/tests",
-            "TestApp\\": "tests/test_app/TestApp",
-            "TestPlugin\\": "tests/test_app/Plugin/TestPlugin/src"
+            "AuthTest\\": "tests/test_app/Plugin/AuthTest/src/",
+            "Authentication\\Test\\": "tests/",
+            "Authentication\\Test\\App\\": "tests/test_app/App/",
+            "Cake\\Test\\": "./vendor/cakephp/cakephp/tests/",
```
dereuromark: PS: `vendor/`, no need for the beginnig ./ in this only line.

PR #158, opened 2017-12-12
Update to phpstan 0.9

ADmad:


robertpustulka:
```
 ------ ----------------------------------------------------------------- 
  Line   src/Identity.php                                                
 ------ ----------------------------------------------------------------- 
  152    Method Authentication\Identity::getOriginalData() should return ArrayAccess but returns array|ArrayAccess.                       
 ------ ----------------------------------------------------------------- 
```

Too bad PHP doesn't have some kind `array_access` pseudo type similar to `iterable`.

Issue #155, opened 2017-12-10
Feedback: Status Codes


inoas:
About the Status Codes:

```
    1: Success (Result::SUCCESS)
    0: Failure (Result::FAILURE)
    -1: Identify not found (Result::FAILURE_IDENTITY_NOT_FOUND)
    -2: Credentials invalid (Result::FAILURE_CREDENTIAL_INVALID)
    -3: Any other kind of failure (Result::FAILURE_OTHER)
    -4: Credentials not found (Result::FAILURE_CREDENTIALS_NOT_FOUND)
```

The numbers are weird ...
- There are no holes to add more inbetween or have them grouped
- Working with negative status codes makes stuff harder to work with as you cannot easily use strpos
- They do not follow posix-style or http-style (why not one of them?)
- They are ordered strange... identity not found > credentials not found > credentials invalid > any other failure or so would make more sense.
- This shows PHP misses atoms :(.

markstory:
> Working with negative status codes makes stuff harder to work with as you cannot easily use strpos

Would you prefer to see string error codes? And `'ok`' for success?

> This shows PHP misses atoms :(.

Sure but we have to make due with what we have.

ADmad:
What's the usecase that would require using string function like strpos() on result code?

inoas:
I would indeed prefer strings that match the constant name, aka `Result::SUCCESS == 'SUCCESS'` - This would also make things unambiguous and easy to debug, IMHO.

@Admad if you group things into error categories, like HTTP does, you can determine the type of error from the first "char" of the error int code.

markstory:
> if you group things into error categories, like HTTP does, you can determine the type of error from the first "char" of the error int code.

We could do something similar with our error codes too. Just need more than 1 digit.

burzum:
> I would indeed prefer strings that match the constant name, aka Result::SUCCESS == 'SUCCESS' - This would also make things unambiguous and easy to debug, IMHO (and would be closest to atoms)

Each time I spot a string comparison like that I comment on that issue in the PR and demand to use a constant. Why would ever have the case you demonstrate? We have constants because they're, well, guess, constant values and have a clear name. The thing you try to show here would be actually:

```php
if ($result->getCode() !== Result::SUCCESS) { /* handle error in detail */}
```

This reads very nice and leaves no room for typos, not in a string nor in an integer. If you make a type in the constant an error will be thrown.

dereuromark:
[...]

PR #159, opened 2017-12-12
Use string based Result:: constants

inoas:
Fixes https://github.com/cakephp/authentication/issues/155

### DIFF
--- a/docs/Quick-start-and-introduction.md
+++ b/docs/Quick-start-and-introduction.md
@@ -146,7 +146,6 @@ if ($result->isValid()) {
 The result sets objects status returned from `getStatus()` will match one of these these constants in the Result object:
 
 * `ResultInterface::SUCCESS`, when successful.
-* `ResultInterface::FAILURE`, on any general failure.
 * `ResultInterface::FAILURE_IDENTITY_NOT_FOUND`, when identity could not be found.
 * `ResultInterface::FAILURE_CREDENTIALS_INVALID`, when credentials are invalid.
 * `ResultInterface::FAILURE_CREDENTIALS_MISSING`, when credentials are missing in the request.
--- a/docs/Quick-start-and-introduction.md
+++ b/docs/Quick-start-and-introduction.md
@@ -149,7 +149,7 @@ The result sets objects code returned from `getCode()` will match one of these c
  * `'FAILURE'`, value of `Result::FAILURE`, on any general failure.
  * `'FAILURE_IDENTITY_NOT_FOUND'`, value of `Result::FAILURE_IDENTITY_NOT_FOUND`, when identify could not be found.
  * `'FAILURE_CREDENTIALS_INVALID'`, value of `Result::FAILURE_CREDENTIALS_INVALID`, when credentials are invalid.
- * `'FAILURE_CREDENTIALS_NOT_FOUND_IN_REQUEST_IN_REQUEST'`, value of `Result::FAILURE_CREDENTIALS_NOT_FOUND_IN_REQUEST`, when credentials are not found in request.
+ * `'FAILURE_CREDENTIALS_NOT_IN_REQUEST_IN_REQUEST'`, value of `Result::FAILURE_CREDENTIALS_NOT_IN_REQUEST`, when credentials are not found in request.
  * `'FAILURE_OTHER'`, value of `Result::FAILURE_OTHER`, on any other kind of failure.
 
 The error array returned by `getErrors()` contains *additional* information coming from the specific system against which the authentication attempt was made. For example LDAP or OAuth would put errors specific to their implementation in here for easier logging and debugging the cause. But most of the included authenticators don't put anything in here.
--- a/docs/Quick-start-and-introduction.md
+++ b/docs/Quick-start-and-introduction.md
@@ -147,7 +147,7 @@ The result sets objects status returned from `getStatus()` will match one of the
 
 * `ResultInterface::SUCCESS`, when successful.
 * `ResultInterface::FAILURE`, on any general failure.
-* `ResultInterface::FAILURE_IDENTITY_NOT_FOUND`, when identify could not be found.
+* `ResultInterface::FAILURE_IDENTITY_NOT_FOUND`, when identity could not be found.
 * `ResultInterface::FAILURE_CREDENTIALS_INVALID`, when credentials are invalid.
 * `ResultInterface::FAILURE_CREDENTIALS_NOT_IN_REQUEST`, when credentials are missing in the request.
 * `ResultInterface::FAILURE_OTHER`, on any other kind of failure.
--- a/docs/Quick-start-and-introduction.md
+++ b/docs/Quick-start-and-introduction.md
@@ -138,19 +138,19 @@ $result = $request->getAttribute('authentication')->getResult();
 if ($result->isValid()) {
     $user = $request->getAttribute('identity');
 } else {
-    $this->log($result->getCode());
+    $this->log($result->getStatus());
     $this->log($result->getErrors());
 }
 ```
 
-The result sets objects code returned from `getCode()` will match one of these codes. The codes are implemented as class constants in the Result object.
+The result sets objects status returned from `getStatus()` will match one of these these constants in the Result object:
 
- * `ResultInterface::SUCCESS`, when successful.
- * `ResultInterface::FAILURE`, on any general failure.
- * `ResultInterface::FAILURE_IDENTITY_NOT_FOUND`, when identify could not be found.
- * `ResultInterface::FAILURE_CREDENTIALS_INVALID`, when credentials are invalid.
- * `ResultInterface::FAILURE_CREDENTIALS_NOT_IN_REQUEST`, when credentials are missing in the request.
- * `ResultInterface::FAILURE_OTHER`, on any other kind of failure.
+* `ResultInterface::SUCCESS`, when successful.
+* `ResultInterface::FAILURE`, on any general failure.
+* `ResultInterface::FAILURE_IDENTITY_NOT_FOUND`, when identify could not be found.
+* `ResultInterface::FAILURE_CREDENTIALS_INVALID`, when credentials are invalid.
+* `ResultInterface::FAILURE_CREDENTIALS_NOT_IN_REQUEST`, when credentials are missing in the request.
+* `ResultInterface::FAILURE_OTHER`, on any other kind of failure.
 
 The error array returned by `getErrors()` contains *additional* information coming from the specific system against which the authentication attempt was made. For example LDAP or OAuth would put errors specific to their implementation in here for easier logging and debugging the cause. But most of the included authenticators don't put anything in here.
 
--- a/docs/Quick-start-and-introduction.md
+++ b/docs/Quick-start-and-introduction.md
@@ -149,7 +149,7 @@ The result sets objects status returned from `getStatus()` will match one of the
 * `ResultInterface::FAILURE`, on any general failure.
 * `ResultInterface::FAILURE_IDENTITY_NOT_FOUND`, when identity could not be found.
 * `ResultInterface::FAILURE_CREDENTIALS_INVALID`, when credentials are invalid.
-* `ResultInterface::FAILURE_CREDENTIALS_NOT_IN_REQUEST`, when credentials are missing in the request.
+* `ResultInterface::FAILURE_CREDENTIALS_MISSING`, when credentials are missing in the request.
 * `ResultInterface::FAILURE_OTHER`, on any other kind of failure.
 
 The error array returned by `getErrors()` contains *additional* information coming from the specific system against which the authentication attempt was made. For example LDAP or OAuth would put errors specific to their implementation in here for easier logging and debugging the cause. But most of the included authenticators don't put anything in here.
--- a/docs/Quick-start-and-introduction.md
+++ b/docs/Quick-start-and-introduction.md
@@ -145,12 +145,12 @@ if ($result->isValid()) {
 
 The result sets objects code returned from `getCode()` will match one of these codes. The codes are implemented as class constants in the Result object.
 
- * `'SUCCESS'`, value of `Result::SUCCESS`, when successful.
- * `'FAILURE'`, value of `Result::FAILURE`, on any general failure.
- * `'FAILURE_IDENTITY_NOT_FOUND'`, value of `Result::FAILURE_IDENTITY_NOT_FOUND`, when identify could not be found.
- * `'FAILURE_CREDENTIALS_INVALID'`, value of `Result::FAILURE_CREDENTIALS_INVALID`, when credentials are invalid.
- * `'FAILURE_CREDENTIALS_NOT_IN_REQUEST'`, value of `Result::FAILURE_CREDENTIALS_NOT_IN_REQUEST`, when credentials are not found in request.
- * `'FAILURE_OTHER'`, value of `Result::FAILURE_OTHER`, on any other kind of failure.
+ * `ResultInterface::SUCCESS`, when successful.
+ * `ResultInterface::FAILURE`, on any general failure.
+ * `ResultInterface::FAILURE_IDENTITY_NOT_FOUND`, when identify could not be found.
+ * `ResultInterface::FAILURE_CREDENTIALS_INVALID`, when credentials are invalid.
+ * `ResultInterface::FAILURE_CREDENTIALS_NOT_IN_REQUEST`, when credentials are missing in the request.
+ * `ResultInterface::FAILURE_OTHER`, on any other kind of failure.
 
 The error array returned by `getErrors()` contains *additional* information coming from the specific system against which the authentication attempt was made. For example LDAP or OAuth would put errors specific to their implementation in here for easier logging and debugging the cause. But most of the included authenticators don't put anything in here.
 
--- a/docs/Quick-start-and-introduction.md
+++ b/docs/Quick-start-and-introduction.md
@@ -145,12 +145,12 @@ if ($result->isValid()) {
 
 The result sets objects code returned from `getCode()` will match one of these codes. The codes are implemented as class constants in the Result object.
 
- * **1**: Success (Result::SUCCESS)
- * **0**: Failure (Result::FAILURE)
- * **-1**: Identify not found (Result::FAILURE_IDENTITY_NOT_FOUND)
- * **-2**: Credentials invalid (Result::FAILURE_CREDENTIAL_INVALID)
- * **-3**: Any other kind of failure (Result::FAILURE_OTHER)
- * **-4**: Credentials not found (Result::FAILURE_CREDENTIALS_NOT_FOUND)
+ * **`'SUCCESS'`** aka `Result::SUCCESS`, when successful.
+ * **`'FAILURE'`** aka `Result::FAILURE`, on any general failure.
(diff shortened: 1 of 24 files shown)


inoas:
### Note

* I have also took the liberty to change one constant name to reflect better what the status code means/does and make singular/plural consistent.
* Unit tests pass on my side (PHPUnit 6.5.4).

### Questions

- Not sure we need `Result::FAILURE`, see: https://github.com/cakephp/authentication/issues/155#issuecomment-351170979

src/Authenticator/Result.php
```diff
@@ -74,13 +74,13 @@ public function __construct($data, $code, array $messages = [])
      */
     public function isValid()
     {
-        return $this->_code > 0;
+        return strpos($this->_code, 'SUCCESS') === 0;
```
ADmad: `return $this->_code === ResultInterface::SUCCESS;`

src/Authenticator/ResultInterface.php
```diff
@@ -17,32 +17,32 @@
...
      */
-    const FAILURE_CREDENTIAL_INVALID = -2;
+    const FAILURE_CREDENTIALS_INVALID = 'FAILURE_CREDENTIALS_INVALID';
 
     /**
-     * Failure due to other circumstances.
+     * The authentication credentials were not found in the request.
      */
-    const FAILURE_OTHER = -3;
+    const FAILURE_CREDENTIALS_NOT_FOUND_IN_REQUEST = 'FAILURE_CREDENTIALS_NOT_FOUND_IN_REQUEST';
```
ADmad: This constant name is too long, `_IN_REQUEST` should be dropped.

src/Authenticator/Result.php
inoas: If you are certain that there is only one kind of SUCESS (not a range) in future.
I will adapt it for now as for now there is not.

src/Authenticator/ResultInterface.php
inoas: would `FAILURE_CREDENTIALS_NOT_IN_REQUEST` instead of `FAILURE_CREDENTIALS_NOT_FOUND` be ok?

docs/Quick-start-and-introduction.md
```diff
@@ -149,7 +149,7 @@ The result sets objects code returned from `getCode()` will match one of these c
  * `'FAILURE'`, value of `Result::FAILURE`, on any general failure.
  * `'FAILURE_IDENTITY_NOT_FOUND'`, value of `Result::FAILURE_IDENTITY_NOT_FOUND`, when identify could not be found.
  * `'FAILURE_CREDENTIALS_INVALID'`, value of `Result::FAILURE_CREDENTIALS_INVALID`, when credentials are invalid.
- * `'FAILURE_CREDENTIALS_NOT_FOUND_IN_REQUEST_IN_REQUEST'`, value of `Result::FAILURE_CREDENTIALS_NOT_FOUND_IN_REQUEST`, when credentials are not found in request.
+ * `'FAILURE_CREDENTIALS_NOT_IN_REQUEST_IN_REQUEST'`, value of `Result::FAILURE_CREDENTIALS_NOT_IN_REQUEST`, when credentials are not found in request.
```
dereuromark: ..._IN_REQUEST_IN_REQUEST?

inoas: bugs in my brain... eh find and replace ;-)? (fixed already)

src/Authenticator/ResultInterface.php
```diff
@@ -17,32 +17,32 @@
...
      */
-    const FAILURE_OTHER = -3;
+    const FAILURE_CREDENTIALS_NOT_IN_REQUEST = 'FAILURE_CREDENTIALS_NOT_IN_REQUEST';
 
     /**
-     * The authentication credentials were not found in the request.
+     * Failure due to other circumstances.
      */
-    const FAILURE_CREDENTIALS_NOT_FOUND = -4;
+    const FAILURE_OTHER = 'FAILURE_OTHER';
```
ADmad: Either of `FAILURE` or `FAILURE_OTHER` could be dropped. Either one would suffice for any non-specific failures.

inoas: Is that correct @burzum - then I will remove FAILURE_OTHER and replace all instances in the code base and tests with FAILURE.

inoas: or should we keep FAILURE_OTHER... maybe better?

ADmad: I would prefer keeping `FAILURE_OTHER`, let's see what others have to say.

src/Authenticator/ResultInterface.php
```diff
@@ -54,7 +54,7 @@ public function isValid();
     /**
      * Get the result code for this authentication attempt.
      *
-     * @return int
+     * @return string
      */
     public function getCode();
```
robertpustulka: Should we change the method name to reflect this change? This is no longer a “code”, more like “status”?

src/Authenticator/ResultInterface.php
```diff
@@ -17,32 +17,32 @@
...
      */
-    const FAILURE_CREDENTIAL_INVALID = -2;
+    const FAILURE_CREDENTIALS_INVALID = 'FAILURE_CREDENTIALS_INVALID';
 
     /**
-     * Failure due to other circumstances.
+     * The authentication credentials were not found in the request.
      */
-    const FAILURE_OTHER = -3;
+    const FAILURE_CREDENTIALS_NOT_IN_REQUEST = 'FAILURE_CREDENTIALS_NOT_IN_REQUEST';
```
robertpustulka: What’s the reason behind that change?

docs/Quick-start-and-introduction.md
```diff
@@ -145,12 +145,12 @@ if ($result->isValid()) {
 
 The result sets objects code returned from `getCode()` will match one of these codes. The codes are implemented as class constants in the Result object.
 
- * **1**: Success (Result::SUCCESS)
- * **0**: Failure (Result::FAILURE)
- * **-1**: Identify not found (Result::FAILURE_IDENTITY_NOT_FOUND)
- * **-2**: Credentials invalid (Result::FAILURE_CREDENTIAL_INVALID)
- * **-3**: Any other kind of failure (Result::FAILURE_OTHER)
- * **-4**: Credentials not found (Result::FAILURE_CREDENTIALS_NOT_FOUND)
+ * `'SUCCESS'`, value of `Result::SUCCESS`, when successful.
```
robertpustulka: Maybe let’s point to the interface instead? `ResultInterface::SUCCESS` etc...

src/Authenticator/ResultInterface.php
dereuromark: getStatus() or getStatusCode()?
It kind of is still a code, just not an int but string one.

ADmad: getStatus() feels more appropriate.

src/Authenticator/ResultInterface.php
burzum: Just pick one of both, I would favor FAILURE because it's shorter and has no better or more specific meaning than FAILURE_OTHER - IMO.

docs/Quick-start-and-introduction.md
```diff
@@ -145,12 +145,12 @@ if ($result->isValid()) {
...
+ * `'FAILURE'`, value of `Result::FAILURE`, on any general failure.
+ * `'FAILURE_IDENTITY_NOT_FOUND'`, value of `Result::FAILURE_IDENTITY_NOT_FOUND`, when identify could not be found.
+ * `'FAILURE_CREDENTIALS_INVALID'`, value of `Result::FAILURE_CREDENTIALS_INVALID`, when credentials are invalid.
+ * `'FAILURE_CREDENTIALS_NOT_IN_REQUEST'`, value of `Result::FAILURE_CREDENTIALS_NOT_IN_REQUEST`, when credentials are not found in request.
```
markstory: Isn't this effectively the same as the identity not being found?

src/Authenticator/ResultInterface.php
bcrowe: I think the `_IN_REQUEST` bit is the frivolous bit. `FAILURE_CREDENTIALS_NOT_FOUND` does the trick, as @ADmad suggested,

docs/Quick-start-and-introduction.md
robertpustulka: `FAILURE_IDENTITY_NOT_FOUND` is returned when an identity with given credentials hasn't been found by any identifier. `FAILURE_CREDENTIALS_NOT_FOUND` is returned when for example a username hasn't been found in post data, which usually means skipping the autenthicator and trying the next one.

burzum: Can't we remove the FAILURE_ prefix as well? It's obvious that there is a problem if the identity was not found, same for the others.

src/Authenticator/ResultInterface.php
inoas: I did move around the lines, I did not replace those. Take a look at the full diff.

docs/Quick-start-and-introduction.md
inoas: @markstory right because I was asking me this question I changed the constant name after reading a bit into the code base.

=> IDENTITY_NOT_FOUND => user does not exist
=> CREDENTIALS_NOT_IN_REQUEST => user credentials are missing in the request

burzum: Just a remark, I'ts named 'IDENTITY' and not 'USER' for a reason: Not every actor in a system is an user. Could be a service or system as well that asks for authentication.

inoas:
@robertpustulka

A quick glimse: I agree
Can we do that in a stand-alone-PR?

src/Authenticator/ResultInterface.php
robertpustulka: @inoas I'm talking about `FAILURE_CREDENTIALS_NOT_FOUND` => `FAILURE_CREDENTIALS_NOT_IN_REQUEST`

docs/Quick-start-and-introduction.md
```diff
@@ -145,12 +145,12 @@ if ($result->isValid()) {
...
 
- * **1**: Success (Result::SUCCESS)
- * **0**: Failure (Result::FAILURE)
- * **-1**: Identify not found (Result::FAILURE_IDENTITY_NOT_FOUND)
- * **-2**: Credentials invalid (Result::FAILURE_CREDENTIAL_INVALID)
- * **-3**: Any other kind of failure (Result::FAILURE_OTHER)
- * **-4**: Credentials not found (Result::FAILURE_CREDENTIALS_NOT_FOUND)
+ * `ResultInterface::SUCCESS`, when successful.
+ * `ResultInterface::FAILURE`, on any general failure.
+ * `ResultInterface::FAILURE_IDENTITY_NOT_FOUND`, when identify could not be found.
```
robertpustulka: Typo, should be "when identity could not be found"

src/Authenticator/Result.php
```diff
@@ -44,25 +44,28 @@ class Result implements ResultInterface
...
      */
-    public function __construct($data, $code, array $messages = [])
+    public function __construct($data, $status, array $messages = [])
     {
-        if (empty($data) && $code === self::SUCCESS) {
+        if ($status === self::SUCCESS && empty($data)) {
             throw new InvalidArgumentException('Identity data can not be empty with status success.');
         }
-        if ($data !== null && !is_array($data) && !$data instanceof ArrayAccess) {
+        if ($data !== null && !is_array($data) && !($data instanceof ArrayAccess)) {
```
inoas: This is better, right?

src/Authenticator/Result.php
```diff
@@ -44,25 +44,28 @@ class Result implements ResultInterface
...
-     * @param int $code Error code.
+     * @param string $status Status constant equivalent.
      * @param array $messages Messages.
      * @throws InvalidArgumentException When invalid identity data is passed.
      */
-    public function __construct($data, $code, array $messages = [])
+    public function __construct($data, $status, array $messages = [])
     {
-        if (empty($data) && $code === self::SUCCESS) {
+        if ($status === self::SUCCESS && empty($data)) {
```
inoas: The boolean check is faster and if false the whole statement directly returns false afaik.

src/Authenticator/ResultInterface.php
```diff
@@ -15,34 +15,29 @@
 interface ResultInterface
 {
     /**
-     * General Failure
+     * Failure due to invalid credentials being supplied.
      */
-    const FAILURE = 0;
+    const FAILURE_CREDENTIALS_INVALID = 'FAILURE_CREDENTIALS_INVALID';
```
inoas: git/github diff doesn't detect the moving around of constants correctly. Take a look at the whole constants before/after

src/Authenticator/Result.php
dereuromark: faster is irrelevant in the scope of this functionality :) Especially on that level.

inoas: Well it is also faster in my brain ;-) parsing empty() in my brain is probably the slowest operation existing ;-)

docs/Quick-start-and-introduction.md
```diff
@@ -138,19 +138,19 @@ $result = $request->getAttribute('authentication')->getResult();
...
- * **0**: Failure (Result::FAILURE)
- * **-1**: Identify not found (Result::FAILURE_IDENTITY_NOT_FOUND)
- * **-2**: Credentials invalid (Result::FAILURE_CREDENTIAL_INVALID)
- * **-3**: Any other kind of failure (Result::FAILURE_OTHER)
- * **-4**: Credentials not found (Result::FAILURE_CREDENTIALS_NOT_FOUND)
+* `ResultInterface::SUCCESS`, when successful.
+* `ResultInterface::FAILURE`, on any general failure.
+* `ResultInterface::FAILURE_IDENTITY_NOT_FOUND`, when identity could not be found.
+* `ResultInterface::FAILURE_CREDENTIALS_INVALID`, when credentials are invalid.
+* `ResultInterface::FAILURE_CREDENTIALS_NOT_IN_REQUEST`, when credentials are missing in the request.
```
markstory: I don't think we need to separate 'not found' and 'credentials not in request' cases. They are both not found errors.

src/Authenticator/Result.php
markstory: Its the same as far as PHP is concerned.

robertpustulka: `!($data instanceof ArrayAccess)` is the same as `!$data instanceof ArrayAccess`

docs/Quick-start-and-introduction.md
robertpustulka: @markstory are you talking about `FAILURE_IDENTITY_NOT_FOUND` and `FAILURE_CREDENTIALS_NOT_IN_REQUEST`?

src/Authenticator/Result.php
dereuromark: But it is sure more clear to everyone, especially beginners. So :+1: for keeping it for clarity.

docs/Quick-start-and-introduction.md
inoas: @markstory it is not the same afaik. Identity not found means there was no match because the given identity was not found in the identity data source whereas not in request means the requesting endpoint/user forgot/omitted to supply the required credentials.

markstory: @robertpustulka I am talking about those.

@inoas The credentials not existing is ~= to empty values which would result in the identity not being found. I'm not clear on *why* the separation is necessary.

src/Authenticator/ResultInterface.php
inoas: @ADmad I changed this again... are you okay with this, now?

docs/Quick-start-and-introduction.md
inoas: Well I didn't write that code so I am not sure about its intentions:

Here are 2 snips of code maybe the primary author of those code parts can explain better why both are required (or that one can be discaded):

Keep in mind that I renamed FAILURE_CREDENTIALS_NOT_IN_REQUEST to FAILURE_CREDENTIALS_MISSING:

In HttpDigestAuthenticator.php
```php
    public function authenticate(ServerRequestInterface $request, ResponseInterface $response)
    {
        $digest = $this->_getDigest($request);
        if (empty($digest)) {
            return new Result(null, Result::FAILURE_OTHER);
        }

        $user = $this->_identifier->identify([
            IdentifierInterface::CREDENTIAL_USERNAME => $digest['username']
        ]);

        if (empty($user)) {
        }

        if (!$this->validNonce($digest['nonce'])) {
            return new Result(null, Result::FAILURE_CREDENTIALS_INVALID);
        }

//...

        return new Result(null, Result::FAILURE_CREDENTIALS_INVALID);
```

In JwtAuthenticator.php:
```php
    public function authenticate(ServerRequestInterface $request, ResponseInterface $response)
    {
        try {
            $result = $this->getPayload($request);
        } catch (Exception $e) {
            return new Result(
                null,
                Result::FAILURE_CREDENTIALS_INVALID,
                [
                    'message' => $e->getMessage(),
                    'exception' => $e
                ]
            );
        }

        if (!$result instanceof stdClass) {
            return new Result(null, Result::FAILURE_CREDENTIALS_INVALID);
        }

        $key = IdentifierInterface::CREDENTIAL_JWT_SUBJECT;
        if (empty($result[$key])) {
            return new Result(null, Result::FAILURE_CREDENTIALS_MISSING);
        }

//...

        if (empty($user)) {
            return new Result(null, Result::FAILURE_IDENTITY_NOT_FOUND, $this->_identifier->getErrors());
        }

        return new Result($user, Result::SUCCESS);
    }
```

inoas:
Anything else?

burzum:
Can we just remove the FAILURE prefix? It makes the messages just longer without adding benefit. It's obvious that CREDENTIALS_INVALID for example is an error. :)

robertpustulka:
@inoas 

> A quick glimse: I agree
> Can we do that in a stand-alone-PR?

I think we can do it here.

inoas:
@robertpustulka wrote:
> Shoudl we fix the status codes here:
https://github.com/cakephp/authentication/blob/master/src/Authenticator/TokenAuthenticator.php#L119
and https://github.com/cakephp/authentication/blob/master/src/Authenticator/HttpDigestAuthenticator.php#L82
> I think that empty token or digest mean that the credentials haven't been found.

@burzum you concur, then I will change this.

docs/Quick-start-and-introduction.md
```diff
@@ -138,19 +138,19 @@ $result = $request->getAttribute('authentication')->getResult();
...
+The result sets objects status returned from `getStatus()` will match one of these these constants in the Result object:
 
- * **1**: Success (Result::SUCCESS)
- * **0**: Failure (Result::FAILURE)
- * **-1**: Identify not found (Result::FAILURE_IDENTITY_NOT_FOUND)
- * **-2**: Credentials invalid (Result::FAILURE_CREDENTIAL_INVALID)
- * **-3**: Any other kind of failure (Result::FAILURE_OTHER)
- * **-4**: Credentials not found (Result::FAILURE_CREDENTIALS_NOT_FOUND)
+* `ResultInterface::SUCCESS`, when successful.
+* `ResultInterface::FAILURE`, on any general failure.
```
ADmad: There's no `FAILURE` anymore.

PR #163, opened 2017-12-13
Fixed License Header in source files

inoas:
Will this become a read only repo once "merged" with cakephp/cakephp or stay a stand-alone repo/plugin?
What about a `@since` tag? 1.0.0? 3.6.0?


markstory:
> Will this become a read only repo once "merged" with cakephp/cakephp or stay a stand-alone repo/plugin?

It will stay a standalone plugin like bake or debugkit.

PR #169, opened 2017-12-14
Fix request and response not being modified by persistence authenticators

robertpustulka:
Modifying a request or response from `PersistenceInterface::persistIdentity()` handler had no effect.
This results in cookies not being sent to the browser using the `CookieAuthenticator`.

This fix makes sure that request and response objects are returned by `AuthenticationService::authenticate()` method and passed through in `AuthenticationMiddleware`.

### DIFF
--- a/src/AuthenticationServiceInterface.php
+++ b/src/AuthenticationServiceInterface.php
@@ -14,10 +14,11 @@
  */
 namespace Authentication;
 
+use Authentication\Authenticator\PersistenceInterface;
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
 
-interface AuthenticationServiceInterface
+interface AuthenticationServiceInterface extends PersistenceInterface
 {
 
     /**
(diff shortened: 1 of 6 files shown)


src/AuthenticationServiceInterface.php
```diff
@@ -56,25 +56,6 @@ public function authenticate(ServerRequestInterface $request, ResponseInterface
...
     public function getIdentity();
 
-    /**
-     * Clears the identity from authenticators that store them and the request.
-     *
-     * @param \Psr\Http\Message\ServerRequestInterface $request The request.
-     * @param \Psr\Http\Message\ResponseInterface $response The response.
-     * @return array Return an array containing the request and response objects.
-     */
-    public function clearIdentity(ServerRequestInterface $request, ResponseInterface $response);
```
robertpustulka: `clearIdentity` and `setIdentity` were removed in favor of `PersistenceInterface` which now `AuthenticationServiceInterface` extends.

Issue #167, opened 2017-12-13
Login Url checking no longer supports route names


lilHermit:
The new UrlChecking doesn't support named routes anymore

For example previously you could do

```
    $authService->loadAuthenticator('Authentication.Form', [
        'fields' => $fields,
        'loginUrl' => ['_name' => 'admin:login']
    ]);
```

now you need to do 

```
    $authService->loadAuthenticator('Authentication.Form', [
        'fields' => $fields,
        'loginUrl' => '/admin/login'
    ]);
```

This seems like a backwards step

robertpustulka:
A url checker with CakePHP Router support is planned. We want to focus on releasing the stable version first, then we will start adding new features.

Thanks for the feedback!

lilHermit:
@robertpustulka I'll write a UrlChecker which translates a Routing array.

burzum:
@markstory I've told you somebody will complain about the lack of URL routing array support. 😆 I just didn't expect it to happen that fast.

@lilHermit any time frame when you're going to do it please? If it's days or weeks I might do it before.

lilHermit:
@burzum I've already written one, do you want me to contribute it?

burzum:
@lilHermit Yes, please open a PR that includes the code and tests or paste it here properly formatted.

lilHermit:
@burzum ok I'll try to do a PR but I haven't done tests and have a tight deadline here. Here this class currently

```
<?php

namespace App\Authentication;

use Authentication\UrlChecker\UrlCheckerInterface;
use Cake\Routing\Exception\MissingRouteException;
use Cake\Routing\Router;
use Psr\Http\Message\ServerRequestInterface;

class CakephpRouteUrlChecker implements UrlCheckerInterface {

    /**
     * Checks the requests if it is the configured login action
     *
     * @param ServerRequestInterface $request   Server Request
     * @param string|array           $loginUrls Login URL string or array of URLs
     * @param array                  $options   Array of options
     *
     * @return bool
     */
    public function check(ServerRequestInterface $request, $loginUrls, array $options = []) {

        $url = $request->getUri()->getPath();

        // Support string loginUrls
        if (is_string($loginUrls)) {
            return ($loginUrls === $url);
        }

        if (!is_array($loginUrls)) {
            return false;
        }
        if (empty($loginUrls)) {
            return true;
        }

        // If it's a single route array add to another
        if (!is_numeric(key($loginUrls))) {
            $loginUrls = [$loginUrls];
[...]

PR #171, opened 2017-12-14
Checks URLs against an URL array using the Cake router

burzum:
Refs https://github.com/cakephp/authentication/issues/167

### DIFF
--- a/docs/URL-Checkers.md
+++ b/docs/URL-Checkers.md
@@ -1,0 +1,25 @@
+# URL Checkers
+
+To provide an abstract and framework agnostic solution as much as possible there are URL checkers implemented that allow you to customize the comparision of the current URL if needed. For example to another frameworks routing.
+
+## Included Checkers
+
+### DefaultUrlChecker
+
+The default checker allows you to compare an URL by regex or string URLs.
+
+Options:
+
+ * **checkFullUrl**: To compare the full URL, including protocol, host and port or not. Default is `false`
+ * **useRegex**: Compares the URL by a regular expression provided in the `$loginUrls` argument of the checker.
+ 
+### CakeRouterChecker
+
+Options:
+
+Use this checker if you want to use the array notation of CakePHPs routing system. The checker also works with named routes. 
+ * **checkFullUrl**: To compare the full URL, including protocol, host and port or not. Default is `false`
+
+## Implementing your own Checker
+
+An URL checker **must** implement the ``UrlCheckterInterface``.
--- a/src/UrlChecker/CakeRouterChecker.php
+++ b/src/UrlChecker/CakeRouterChecker.php
@@ -14,7 +14,6 @@
  */
 namespace Authentication\UrlChecker;
 
-use Cake\Routing\Exception\MissingRouteException;
 use Cake\Routing\Router;
 use InvalidArgumentException;
 use Psr\Http\Message\ServerRequestInterface;
@@ -24,6 +23,16 @@ use Psr\Http\Message\ServerRequestInterface;
  */
 class CakeRouterChecker extends DefaultUrlChecker
 {
+    /**
+     * Default Options
+     *
+     * - `checkFullUrl` Whether or not to check the full request URI.
+     *
+     * @var array
+     */
+    protected $_defaultOptions = [
+        'checkFullUrl' => false,
+    ];
 
     /**
      * {@inheritdoc}
--- a/src/UrlChecker/CakeRouterChecker.php
+++ b/src/UrlChecker/CakeRouterChecker.php
@@ -41,7 +41,7 @@ class CakeRouterChecker implements UrlCheckerInterface
     {
         $options = $this->_mergeDefaultOptions($options);
         $url = $this->_getUrlFromRequest($request->getUri(), $options['checkFullUrl']);
-debug($url);
+
         // Support string loginUrls
         if (is_string($loginUrls)) {
             return ($loginUrls === $url);
--- a/src/UrlChecker/CakeRouterChecker.php
+++ b/src/UrlChecker/CakeRouterChecker.php
@@ -1,0 +1,105 @@
+<?php
+/**
+ * CakePHP(tm) : Rapid Development Framework (https://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ *
+ * Licensed under The MIT License
+ * For full copyright and license information, please see the LICENSE.txt
+ * Redistributions of files must retain the above copyright notice.
+ *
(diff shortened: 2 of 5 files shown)


tests/TestCase/UrlChecker/CakeRouterCheckerTest.php
```diff
@@ -0,0 +1,86 @@
...
+     * @return void
+     */
+    public function testCheckFullUrls()
+    {
+        $checker = new CakeRouterChecker();
+        $request = ServerRequestFactory::fromGlobals(
+            ['REQUEST_URI' => '/users/login']
+        );
+
+        $result = $checker->check($request, '/users/login', [
```
robertpustulka: Shouldn't there be a Router url array?

burzum: Fixed.

burzum:
@robertpustulka what do you think about this? https://gist.github.com/burzum/5decfd9147a243d55e117d224ce313ae

Instead of creating a complete new checker and duplicating two methods from the default checker, why not just extending it the way I'm doing it?

lilHermit:
I don't see the point of the `checkFullUrl` option as both urls are derived from `Router::url` it's not like you're  comparing a string from the developer? Unless I'm missing something!

src/UrlChecker/CakeRouterChecker.php
```diff
@@ -0,0 +1,106 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    public function check(ServerRequestInterface $request, $loginUrls, array $options = [])
+    {
+        $options = $this->_mergeDefaultOptions($options);
+        $url = $this->_getUrlFromRequest($request->getUri(), $options['checkFullUrl']);
+
+        // Support string loginUrls
+        if (is_string($loginUrls)) {
```
markstory: Why? I thought the goal of this checker was to use the router features?

tests/TestCase/UrlChecker/CakeRouterCheckerTest.php
```diff
@@ -0,0 +1,139 @@
...
+
+        $checker = new CakeRouterChecker();
+        $request = ServerRequestFactory::fromGlobals(
+            ['REQUEST_URI' => '/users/invalid']
+        );
+        $result = $checker->check($request, [
+            'controller' => 'Users',
+            'action' => 'login'
+        ]);
+        $this->assertFalse($result);
```
markstory: Should there be a test for a named route?

src/UrlChecker/CakeRouterChecker.php
burzum: @markstory thats a question for @lilHermit

tests/TestCase/UrlChecker/CakeRouterCheckerTest.php
burzum: OK

src/UrlChecker/CakeRouterChecker.php
robertpustulka: I also don't think that is necessary. If someone needs to support both they can implement a checker collection.

tests/TestCase/UrlChecker/CakeRouterCheckerTest.php
```diff
@@ -0,0 +1,187 @@
...
+            ['REQUEST_URI' => '/users/login']
+        );
+        $result = $checker->check($request, $url, [
+            'checkFullUrl' => true
+        ]);
+        $this->assertTrue($result);
+
+        $checker = new CakeRouterChecker();
+        $request = ServerRequestFactory::fromGlobals(
+            ['REQUEST_URI' => '/users/invalid']
```
robertpustulka: I think we need a test for url with different hosts to make sure this feature works.

docs/URL-Checkers.md
```diff
@@ -0,0 +1,25 @@
+# URL Checkers
```
robertpustulka: Checkers are already mentioned here: https://github.com/cakephp/authentication/blob/master/docs/Authenticators.md#url-checkers

You may want to link this from there or remove checker docs from Authenticators page.

src/UrlChecker/CakeRouterChecker.php
lilHermit: I just added it because thought if it was a string its easy to compare but happy for it to fail

burzum:
@markstory @robertpustulka can we merge it now?

src/UrlChecker/CakeRouterChecker.php
```diff
@@ -0,0 +1,57 @@
...
+
+        // If it's a single route array add to another
+        if (!is_numeric(key($loginUrls))) {
+            $loginUrls = [$loginUrls];
+        }
+
+        foreach ($loginUrls as $validUrl) {
+            try {
+                $validUrl = Router::url($validUrl, $options['checkFullUrl']);
+            } catch (MissingRouteException $e) {
```
robertpustulka: I don’t think we should suppress this exception and allow for missing routes silently.

src/UrlChecker/CakeRouterChecker.php
```diff
@@ -0,0 +1,57 @@
...
+
+    /**
+     * {@inheritdoc}
+     */
+    public function check(ServerRequestInterface $request, $loginUrls, array $options = [])
+    {
+        $options = $this->_mergeDefaultOptions($options);
+        $url = $this->_getUrlFromRequest($request->getUri(), $options['checkFullUrl']);
+
+        if (!is_array($loginUrls) || empty($loginUrls)) {
```
robertpustulka: Default checker returns true for empty url. Should we allow for the same behavior here?

burzum: Why does it return true? Shouldn't we always return false if it doesn't match or is empty?

robertpustulka: I’m not sure why. It had been that way in the FormAuthentication, before checkers were implemented.

burzum: I'm OK with either way we just need to make a decision. :) But wouldn't it make more sense to return false if it's empty?

robertpustulka: It would probably be safer that way. So I’m ok with returning false on empty urls.

burzum: Are we OK to merge this then?

robertpustulka: I think we need another test as I mentioned here: https://github.com/cakephp/authentication/pull/171/files#r157144178

PR #173, opened 2017-12-15
Implement __get() and __isset() on Identity.

markstory:
This makes reading data simpler. I've also used field mappings consistently as they were only applied in some situations before.

I've made Identity::get() protected as I don't think we need 3 ways to access properties.

### DIFF
--- a/src/Identity.php
+++ b/src/Identity.php
@@ -27,6 +27,7 @@ class Identity implements IdentityInterface
 
     /**
      * Default configuration.
+     *
      * - `fieldMap` Mapping of fields
      *
      * @var array
@@ -72,14 +73,36 @@ class Identity implements IdentityInterface
     }
 
     /**
+     * Get data from the identity using object access.
+     *
+     * @param string $field Field in the user data.
+     * @return mixed
+     */
+    public function __get($field)
+    {
+        return $this->get($field);
+    }
+
+    /**
+     * Check if the field isset() using object access.
+     *
+     * @param string $field Field in the user data.
+     * @return mixed
+     */
+    public function __isset($field)
+    {
+        return $this->get($field) !== null;
+    }
+
+    /**
      * Get data from the identity
      *
      * @param string $field Field in the user data.
      * @return mixed
      */
-    public function get($field)
+    protected function get($field)
     {
-        $map = $this->getConfig('fieldMap');
+        $map = $this->_config['fieldMap'];
         if (isset($map[$field])) {
             $field = $map[$field];
         }
@@ -100,7 +123,7 @@ class Identity implements IdentityInterface
      */
     public function offsetExists($offset)
     {
-        return isset($this->data[$offset]);
+        return $this->get($offset) !== null;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/Identity.php
```diff
@@ -141,6 +164,10 @@ public function offsetSet($offset, $value)
      */
     public function offsetUnset($offset)
     {
+        $map = $this->_config['fieldMap'];
+        if (isset($map[$field])) {
```
robertpustulka: Should be `$map[$offset]`

src/Identity.php
```diff
@@ -71,15 +72,37 @@ public function getIdentifier()
...
+    }
+
     /**
      * Get data from the identity
      *
      * @param string $field Field in the user data.
      * @return mixed
      */
-    public function get($field)
+    protected function get($field)
```
burzum: Oh please, no! Why??? I'd like to have the choice to use get over property access.

I prefer to use get and I really hate having protected methods without the _ prefix. The prefixing is really a helper to identify on sight if it's protected or not.

markstory: So you want to have:

* `get(string)`
* `offsetGet()`
* `__get()`

All 3 methods don't allow typehinting, so I don't know why we'd allow all 3. Would you be open to dropping `ArrayAccess` and keeping `get()` and `__get()`?

markstory:
@robertpustulka I left `__set()` and `__unset()` out as I didn't want to allow the underlying object to be mutated. Having a mutable object inside a PSR7 request attribute breaks the immutability of the request.

src/Identity.php
robertpustulka: I’d stick with ArrayAccess since we force the underlying object to be array|ArrayAccess. This would allow to continue using this interface.

robertpustulka:
@markstory how about dropping setOffest and unsetOffset (throw an exception or sth) then?

markstory:
> how about dropping setOffest and unsetOffset (throw an exception or sth) then?

@robertpustulka I'm ok with that.

markstory:
@robertpustulka @burzum Updated with errors on `offsetSet()` and `offsetUnset()`

robertpustulka:
Do we need `__call` as well?

BTW, some tests are failing.

dangerdan:
Is this documentation now incorrect? https://book.cakephp.org/authentication/1.1/en/identity-object.html

"The identity object provides ArrayAccess but as well a get() method to access data. It is strongly recommended to use the get() method over array access because the get method is aware of the field mapping:

$identity->get('email');
$identity->get('username');
".

src/Identity.php
dereuromark: I am also against this, as this is a code smell for unknown keys (compared to entities where the keys and their types are known): https://github.com/cakephp/authentication/issues/249

dereuromark: That is not true, get() can be made typehinted dynamically via IDE extension, the others cannot.
See [meta](https://github.com/dereuromark/cakephp-ide-helper/blob/master/tests/test_files/meta/phpstorm/.meta.php) file.

markstory:
@dangerdan I don't think the behavior of `get()` has changed. This pull request was for a php magic get method.

Issue #174, opened 2017-12-18
Reduce the amount of `empty()` usage.


inoas:
I don't want to push even more stuff into https://github.com/cakephp/authentication/pull/159
If it gets accepted I want to take a look if I can remove most of the - currently 27 - `empty()` comparisons.

dereuromark:
If you can resolve them without changing behavior, sure, why not.

inoas:
Closing as PR is open.

PR #175, opened 2017-12-18
Remove weak type comparisons

inoas:
* Remove most of the `empty()` uses aside when comparing against `null|array|ArrayAccess`
* unified uses of `$user` vs `$identity` to latter
* replaced some `isset()` comparisons with `array_key_exists()`
* replaced `$this->_config[$something]` calls with $this->getConfig($something) calls.
* fixed ldap_connect integration which actually returns false not null.
* made some class_exists() checks less stringy.

Fixes https://github.com/cakephp/authentication/issues/174

### DIFF
--- a/src/AbstractCollection.php
+++ b/src/AbstractCollection.php
@@ -49,13 +49,13 @@ abstract class AbstractCollection extends ObjectRegistry implements IteratorAggr
     }
 
     /**
-     * Returns true if a collection is empty.
+     * Returns true if the map of identifier objects is empty.
      *
      * @return bool
      */
     public function isEmpty()
     {
-        return empty($this->_loaded);
+        return $this->_loaded === [];
     }
 
     /**
--- a/src/Authenticator/CookieAuthenticator.php
+++ b/src/Authenticator/CookieAuthenticator.php
@@ -101,10 +101,7 @@ class CookieAuthenticator extends AbstractAuthenticator implements PersistenceIn
 
         list($username, $tokenHash) = $token;
 
-        $credentials = [
-            'username' => $username
-        ];
-        $identity = $this->_identifier->identify($credentials);
+        $identity = $this->_identifier->identify(['username' => $username]);
 
         if (empty($identity)) {
             return new Result(null, Result::FAILURE_IDENTITY_NOT_FOUND, $this->_identifier->getErrors());
@@ -127,7 +124,9 @@ class CookieAuthenticator extends AbstractAuthenticator implements PersistenceIn
         $field = $this->getConfig('rememberMeField');
         $bodyData = $request->getParsedBody();
 
-        if (!$this->_checkUrl($request) || !is_array($bodyData) || empty($bodyData[$field])) {
+        if (!$this->_checkUrl($request) || !is_array($bodyData) || !array_key_exists($field, $bodyData)
+            || in_array($bodyData[$field], [false, null, ''], true)
+        ) {
             return [
                 'request' => $request,
                 'response' => $response
--- a/src/Authenticator/CookieAuthenticator.php
+++ b/src/Authenticator/CookieAuthenticator.php
@@ -73,7 +73,7 @@ class CookieAuthenticator extends AbstractAuthenticator implements PersistenceIn
      */
     protected function _checkCakeVersion()
     {
-        if (!class_exists('Cake\Http\Cookie\Cookie')) {
+        if (!class_exists(Cookie::class)) {
             throw new RuntimeException('Install CakePHP version >=3.5.0 to use the `CookieAuthenticator`.');
         }
     }
--- a/src/Authenticator/CookieAuthenticator.php
+++ b/src/Authenticator/CookieAuthenticator.php
@@ -85,7 +85,7 @@ class CookieAuthenticator extends AbstractAuthenticator implements PersistenceIn
     {
         $cookies = $request->getCookieParams();
         $cookieName = $this->getConfig('cookie.name');
-        if (!isset($cookies[$cookieName])) {
+        if (!array_key_exists($cookieName, $cookies)) {
             return new Result(null, Result::FAILURE_CREDENTIALS_MISSING, [
                 'Login credentials not found'
             ]);
--- a/src/Authenticator/FormAuthenticator.php
+++ b/src/Authenticator/FormAuthenticator.php
@@ -54,7 +54,7 @@ class FormAuthenticator extends AbstractAuthenticator
      */
     protected function _getData(ServerRequestInterface $request)
     {
-        $fields = $this->_config['fields'];
+        $fields = $this->getConfig('fields');
         $body = $request->getParsedBody();
 
         $data = [];
--- a/src/Authenticator/FormAuthenticator.php
+++ b/src/Authenticator/FormAuthenticator.php
@@ -115,12 +115,12 @@ class FormAuthenticator extends AbstractAuthenticator
             ]);
         }
 
-        $user = $this->_identifier->identify($data);
+        $identity = $this->_identifier->identify($data);
 
-        if (empty($user)) {
+        if (empty($identity)) {
             return new Result(null, Result::FAILURE_IDENTITY_NOT_FOUND, $this->_identifier->getErrors());
         }
 
-        return new Result($user, Result::SUCCESS);
+        return new Result($identity, Result::SUCCESS);
     }
 }
--- a/src/Authenticator/HttpBasicAuthenticator.php
+++ b/src/Authenticator/HttpBasicAuthenticator.php
@@ -91,7 +91,7 @@ class HttpBasicAuthenticator extends AbstractAuthenticator implements StatelessI
     protected function loginHeaders(ServerRequestInterface $request)
     {
         $server = $request->getServerParams();
-        $realm = $this->getConfig('realm') ?: $server['SERVER_NAME'];
+        $realm = $this->getConfig('realm') === null ? $server['SERVER_NAME'] : $this->getConfig('realm');
 
         return ['WWW-Authenticate' => sprintf('Basic realm="%s"', $realm)];
     }
--- a/src/Authenticator/HttpBasicAuthenticator.php
+++ b/src/Authenticator/HttpBasicAuthenticator.php
@@ -53,7 +53,7 @@ class HttpBasicAuthenticator extends AbstractAuthenticator implements StatelessI
     public function getUser(ServerRequestInterface $request)
     {
         $server = $request->getServerParams();
-        if (!isset($server['PHP_AUTH_USER']) || !isset($server['PHP_AUTH_PW'])) {
+        if (!array_key_exists('PHP_AUTH_USER', $server) || !array_key_exists('PHP_AUTH_PW', $server)) {
             return null;
         }
 
--- a/src/Authenticator/HttpBasicAuthenticator.php
+++ b/src/Authenticator/HttpBasicAuthenticator.php
@@ -35,13 +35,13 @@ class HttpBasicAuthenticator extends AbstractAuthenticator implements StatelessI
      */
     public function authenticate(ServerRequestInterface $request, ResponseInterface $response)
     {
-        $user = $this->getUser($request);
+        $identity = $this->getUser($request);
 
-        if (empty($user)) {
+        if (empty($identity)) {
             return new Result(null, Result::FAILURE_CREDENTIALS_MISSING);
         }
 
-        return new Result($user, Result::SUCCESS);
+        return new Result($identity, Result::SUCCESS);
     }
 
     /**
@@ -66,7 +66,7 @@ class HttpBasicAuthenticator extends AbstractAuthenticator implements StatelessI
 
         return $this->_identifier->identify([
             IdentifierInterface::CREDENTIAL_USERNAME => $username,
-            IdentifierInterface::CREDENTIAL_PASSWORD => $password
+            IdentifierInterface::CREDENTIAL_PASSWORD => $password,
         ]);
     }
 
--- a/src/Authenticator/HttpDigestAuthenticator.php
+++ b/src/Authenticator/HttpDigestAuthenticator.php
@@ -78,15 +78,13 @@ class HttpDigestAuthenticator extends HttpBasicAuthenticator
     public function authenticate(ServerRequestInterface $request, ResponseInterface $response)
     {
         $digest = $this->_getDigest($request);
-        if (empty($digest)) {
+        if ($digest === null) {
             return new Result(null, Result::FAILURE_CREDENTIALS_MISSING);
         }
 
-        $user = $this->_identifier->identify([
-            IdentifierInterface::CREDENTIAL_USERNAME => $digest['username']
-        ]);
+        $identity = $this->_identifier->identify([IdentifierInterface::CREDENTIAL_USERNAME => $digest['username']]);
 
-        if (empty($user)) {
+        if (empty($identity)) {
             return new Result(null, Result::FAILURE_IDENTITY_NOT_FOUND);
         }
 
@@ -94,17 +92,17 @@ class HttpDigestAuthenticator extends HttpBasicAuthenticator
             return new Result(null, Result::FAILURE_CREDENTIALS_INVALID);
         }
 
-        $field = $this->_config['fields'][IdentifierInterface::CREDENTIAL_PASSWORD];
-        $password = $user[$field];
+        $passwordField = $this->getConfig('fields')[IdentifierInterface::CREDENTIAL_PASSWORD];
+        $password = $identity[$passwordField];
 
         $server = $request->getServerParams();
-        if (!isset($server['ORIGINAL_REQUEST_METHOD'])) {
+        if (!array_key_exists('ORIGINAL_REQUEST_METHOD', $server)) {
             $server['ORIGINAL_REQUEST_METHOD'] = $server['REQUEST_METHOD'];
         }
 
         $hash = $this->generateResponseHash($digest, $password, $server['ORIGINAL_REQUEST_METHOD']);
         if (hash_equals($hash, $digest['response'])) {
-            return new Result($user, Result::SUCCESS);
+            return new Result($identity, Result::SUCCESS);
         }
 
         return new Result(null, Result::FAILURE_CREDENTIALS_INVALID);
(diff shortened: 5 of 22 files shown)


inoas:
Generally I really dislike array|ArrayAccess ... it should either be identity objects or just array, not either or messy. Is there a good reason why both is possible?

src/Authenticator/HttpBasicAuthenticator.php
```diff
@@ -91,7 +91,7 @@ public function unauthorizedChallenge(ServerRequestInterface $request)
     protected function loginHeaders(ServerRequestInterface $request)
     {
         $server = $request->getServerParams();
-        $realm = $this->getConfig('realm') ?: $server['SERVER_NAME'];
+        $realm = $this->getConfig('realm') === null ? $server['SERVER_NAME'] : $this->getConfig('realm');
```
dereuromark: This is a case where weak totally suffices and is usually even more correct.

src/Authenticator/HttpDigestAuthenticator.php
```diff
@@ -78,33 +78,31 @@ public function __construct(IdentifierInterface $identifier, array $config = [])
...
         }
 
-        $field = $this->_config['fields'][IdentifierInterface::CREDENTIAL_PASSWORD];
-        $password = $user[$field];
+        $passwordField = $this->getConfig('fields')[IdentifierInterface::CREDENTIAL_PASSWORD];
+        $password = $identity[$passwordField];
 
         $server = $request->getServerParams();
-        if (!isset($server['ORIGINAL_REQUEST_METHOD'])) {
+        if (!array_key_exists('ORIGINAL_REQUEST_METHOD', $server)) {
```
dereuromark: isset() is 100x faster and more correct IMO, a key with null has no meaning here and should not do anything

src/Authenticator/HttpDigestAuthenticator.php
```diff
@@ -114,20 +112,20 @@ public function authenticate(ServerRequestInterface $request, ResponseInterface
      * Gets the digest headers from the request/environment.
      *
      * @param \Psr\Http\Message\ServerRequestInterface $request The request that contains login information.
-     * @return array Array of digest information.
+     * @return array|null Array of digest information.
```
dereuromark: Why weakinging the return types? Empty array should suffice, should it not?

src/Authenticator/HttpDigestAuthenticator.php
```diff
@@ -199,17 +197,17 @@ public static function password($username, $password, $realm)
...
-            'qop' => $this->_config['qop'],
+            'qop' => $this->getConfig('qop'),
             'nonce' => $this->generateNonce(),
-            'opaque' => $this->_config['opaque'] ?: md5($realm)
+            'opaque' => $this->getConfig('opaque') === null ? md5($realm) : $this->getConfig('opaque'),
         ];
 
         $digest = $this->_getDigest($request);
-        if ($digest && isset($digest['nonce']) && !$this->validNonce($digest['nonce'])) {
+        if ($digest !== null && array_key_exists('nonce', $digest) && !$this->validNonce($digest['nonce'])) {
```
dereuromark: Same thing, anything else than isset() is not sane, especially not array_key_exists()

src/Authenticator/JwtAuthenticator.php
```diff
@@ -90,25 +91,23 @@ public function authenticate(ServerRequestInterface $request, ResponseInterface
         $result = json_decode(json_encode($result), true);
 
         $key = IdentifierInterface::CREDENTIAL_JWT_SUBJECT;
-        if (empty($result[$key])) {
+        if (!array_key_exists($key, $result)) {
```
dereuromark: Best to revert all array_key_exists code smells :)

src/Authenticator/TokenAuthenticator.php
```diff
@@ -44,12 +44,12 @@ class TokenAuthenticator extends AbstractAuthenticator implements StatelessInter
     protected function getToken(ServerRequestInterface $request)
     {
         $token = $this->getTokenFromHeader($request, $this->getConfig('header'));
-        if (empty($token)) {
+        if ($token === null) {
```
dereuromark: Are we sure this can never be empty string?

src/Authenticator/TokenAuthenticator.php
```diff
@@ -72,14 +72,14 @@ protected function stripTokenPrefix($token, $prefix)
...
+     * @param string|null $headerLine Header name
      * @return string|null
      */
     protected function getTokenFromHeader(ServerRequestInterface $request, $headerLine)
     {
-        if (!empty($headerLine)) {
+        if ($headerLine !== null) {
             $header = $request->getHeaderLine($headerLine);
-            if (!empty($header)) {
+            if ($header !== '' && $header !== null) {
```
dereuromark: In general `if (!$header)` removes the cloaking and suffices. Usually way more clean then this which only forgets other nullish values.

src/Authenticator/HttpBasicAuthenticator.php
inoas: So '0' is not a valid realm? Who says that? I will add a test against trimmed ''-string?

src/Authenticator/HttpDigestAuthenticator.php
inoas: Because null makes sense. Returning an empty array is not much of a digest, is it?
Also parseAuthData() may return null and this method returns null so its typehint was just wrong?

src/Authenticator/HttpDigestAuthenticator.php
inoas: Well depends on how precise `_getDigest()` is?

src/Authenticator/JwtAuthenticator.php
inoas: What's the problem in this case?

src/Authenticator/TokenAuthenticator.php
inoas: Can you point me to what you mean by "cloaking" so I can read something about it? Or just name me some terms https://www.google.de/search?q=php+cloaking didn't reveal anything useful to me?

src/Authenticator/HttpDigestAuthenticator.php
inoas: Do you have recent data (not from 2014) for PHP 7.2/7.1 regarding this?

src/Authenticator/TokenAuthenticator.php
inoas: If it is empty string it will fail more below. Aka the credentials where given but did not match anything (empty username can be ok imho, basically)

src/Authenticator/HttpBasicAuthenticator.php
dereuromark: Just wondering about people and their configs. Could also come from DB and accidently be an empty string. Either way shouldnt run into it, no need to undo the current weak check and break stuff for no reason IMO.

src/Authenticator/HttpDigestAuthenticator.php
dereuromark: You can easily come up with this yourself, but its still nonsense to check for keys when the values are the ones we need. And here also nullish values are invalid, all of them.

src/Authenticator/HttpDigestAuthenticator.php
dereuromark: All right.

src/Authenticator/JwtAuthenticator.php
dereuromark: We want to know if there is a value, not if a key exists that could even have null value.

src/Authenticator/HttpBasicAuthenticator.php
inoas: I don't share that philosophy obviously. Weak types create a ton of problems for me and PHP 7.1/7.2 will be harder to upgrade to if we allow implicit weak "forgiving" types everywhere.
I can add a trim() === '' check however easily.

src/Authenticator/TokenAuthenticator.php
dereuromark: https://book.cakephp.org/3.0/en/contributing/cakephp-coding-conventions.html#careful-when-using-empty-isset

inoas: And don't you think `if (!$var) {` will break if $var is non-boolean typehinted in php71/72 with strict mode?

src/Authenticator/JwtAuthenticator.php
inoas: I'll wait for more comments to come in and answer all of them / fix all of them in one go.

src/Authenticator/TokenAuthenticator.php
dereuromark: silent var not exists cloaking. if a typo is in the variable.

src/Authenticator/HttpBasicAuthenticator.php
dereuromark: Never mind then.

src/Authenticator/TokenAuthenticator.php
inoas: Okay I understand. You mean isset($foo[$bar]) "cloaks" $foo - even if it does not exists you will see no warning or error anywhere, correct?

Good news for !$var checks as documented in the book, in php71/72: https://3v4l.org/n4iK2

src/Authenticator/CookieAuthenticator.php
```diff
@@ -73,7 +73,7 @@ public function __construct(IdentifierCollection $identifiers, array $config = [
      */
     protected function _checkCakeVersion()
     {
-        if (!class_exists('Cake\Http\Cookie\Cookie')) {
+        if (!class_exists(Cookie::class)) {
```
ADmad: This needs to be reverted otherwise it will generate an error when the class doesn't exist, which is exactly what this check is trying to avoid.

src/Authenticator/CookieAuthenticator.php
```diff
@@ -85,7 +85,7 @@ public function authenticate(ServerRequestInterface $request, ResponseInterface
     {
         $cookies = $request->getCookieParams();
         $cookieName = $this->getConfig('cookie.name');
-        if (!isset($cookies[$cookieName])) {
+        if (!array_key_exists($cookieName, $cookies)) {
```
ADmad: Revert back to `isset`.

src/Authenticator/CookieAuthenticator.php
```diff
@@ -127,7 +124,9 @@ public function persistIdentity(ServerRequestInterface $request, ResponseInterfa
         $field = $this->getConfig('rememberMeField');
         $bodyData = $request->getParsedBody();
 
-        if (!$this->_checkUrl($request) || !is_array($bodyData) || empty($bodyData[$field])) {
+        if (!$this->_checkUrl($request) || !is_array($bodyData) || !array_key_exists($field, $bodyData)
```
ADmad: Revert this. It reduces readability, makes code more different to follow.

src/Authenticator/HttpDigestAuthenticator.php
ADmad: Revert back to `isset` please.

src/Authenticator/JwtAuthenticator.php
```diff
@@ -52,8 +53,8 @@ public function __construct(IdentifierInterface $identifier, array $config = [])
     {
         parent::__construct($identifier, $config);
 
-        if (empty($this->_config['secretKey'])) {
-            if (!class_exists('Cake\Utility\Security')) {
+        if ($this->getConfig('secretKey') === null) {
+            if (!class_exists(Security::class)) {
```
ADmad: Revert this to string for same reason I explained above.

src/Authenticator/JwtAuthenticator.php
ADmad: Revert back all `array_key_exists()` to `isset()`.

src/Identifier/IdentifierCollection.php
```diff
@@ -28,6 +28,13 @@ class IdentifierCollection extends AbstractCollection implements IdentifierInter
      */
     protected $_errors = [];
 
+    /**
+     * Map of loaded objects.
+     *
+     * @var object[]
+     */
+    protected $_loaded = [];
```
ADmad: Remove this. This class is a dependent of `Cake\Core\ObjectRegistry` which already has the property defined.

src/Identifier/LdapIdentifier.php
```diff
@@ -92,17 +92,17 @@ public function __construct(array $config = [])
...
             throw new InvalidArgumentException(sprintf(
-                'The `bindDN` config is not a callable. Got `%s` instead.',
-                gettype($this->_config['bindDN'])
+                'The value of `bindDN` config key is not a callable. Got `%s` instead.',
+                gettype($this->getConfig('bindDN'))
             ));
         }
-        if (!isset($this->_config['host'])) {
-            throw new RuntimeException('Config `host` is not set.');
+        if ($this->getConfig('host') === null || trim($this->getConfig('host') === '')) {
```
ADmad: Remove the `trim()`, if someone sets invalid host name with whitespace let an error be generated.

src/PasswordHasher/LegacyPasswordHasher.php
```diff
@@ -44,7 +44,7 @@ public function __construct(array $config = [])
         if (Configure::read('debug')) {
             Debugger::checkSecurityKeys();
         }
-        if (!class_exists('Cake\Utility\Security')) {
+        if (!class_exists(Security::class)) {
```
ADmad: Revert.

src/Authenticator/HttpBasicAuthenticator.php
robertpustulka: `if(!strlen($string))` works fine for rejecting `''|null`

dereuromark: Or just keeping the `?:`.

src/AbstractCollection.php
```diff
@@ -49,13 +49,13 @@ public function __construct(array $config = [])
     }
 
     /**
-     * Returns true if a collection is empty.
+     * Returns true if the map of identifier objects is empty.
```
robertpustulka: Please revert. This is an abstract collection for identifiers **and** authenticators.

src/Authenticator/FormAuthenticator.php
```diff
@@ -115,12 +115,12 @@ public function authenticate(ServerRequestInterface $request, ResponseInterface
             ]);
         }
 
-        $user = $this->_identifier->identify($data);
+        $identity = $this->_identifier->identify($data);
```
robertpustulka: Actually this is still a *user* or *user/identity data*. It becomes an *identifier* (`IdentifierInterface` instance) later in authentication service.

src/Authenticator/HttpBasicAuthenticator.php
```diff
@@ -66,7 +66,7 @@ public function getUser(ServerRequestInterface $request)
 
         return $this->_identifier->identify([
             IdentifierInterface::CREDENTIAL_USERNAME => $username,
-            IdentifierInterface::CREDENTIAL_PASSWORD => $password
+            IdentifierInterface::CREDENTIAL_PASSWORD => $password,
```
robertpustulka: Why?

src/Authenticator/HttpDigestAuthenticator.php
```diff
@@ -78,33 +78,31 @@ public function __construct(IdentifierInterface $identifier, array $config = [])
...
             return new Result(null, Result::FAILURE_IDENTITY_NOT_FOUND);
         }
 
         if (!$this->validNonce($digest['nonce'])) {
             return new Result(null, Result::FAILURE_CREDENTIALS_INVALID);
         }
 
-        $field = $this->_config['fields'][IdentifierInterface::CREDENTIAL_PASSWORD];
-        $password = $user[$field];
+        $passwordField = $this->getConfig('fields')[IdentifierInterface::CREDENTIAL_PASSWORD];
```
robertpustulka: How is that better than `$this->_config['fields'][IdentifierInterface::CREDENTIAL_PASSWORD]` ?

src/UrlChecker/UrlCheckerInterface.php
```diff
@@ -25,9 +25,9 @@
      * Checks the requests if it is the configured login action
      *
      * @param \Psr\Http\Message\ServerRequestInterface $request Server Request
-     * @param string|array $loginUrls Login URL string or array of URLs
+     * @param string|array $urls Login URL string or array of URLs
```
robertpustulka: :+1:

Can you remove "Login" from description as well.

src/Authenticator/HttpBasicAuthenticator.php
dereuromark: Multiline trailing comma. We should have a sniffer for this, though IMO.

robertpustulka: OK, I can see why it's a good thing. More readable git diffs for instance :)

src/Authenticator/CookieAuthenticator.php
chinpei215: @ADmad `::class` syntax never generates any errors. See a note of the following:
http://php.net/manual/en/language.oop5.basic.php#language.oop5.basic.class.class

chinpei215:
@inoas I would suggest you to separate your pull request by the type of changes. Then those pull requests would be easy to review and get merged.

src/Authenticator/CookieAuthenticator.php
ADmad: @chinpei215 Thanks, good to know.

robertpustulka:
@inoas any progress on this?

burzum:
@markstory @inoas @robertpustulka can we close this or does anyone want to finish it?

ADmad:
I say close it. It has been open long enough and @inoas doesn't interested in making changes as per feedback.

inoas:
I'll see if I can resolve the conflicts and work through the change requests today. If I can't manage, feel free to close / I will close it.

src/Authenticator/HttpDigestAuthenticator.php
inoas: calling methods is always better. You can set the instance config trait to throw exceptions in case the keys do not exist, whereas the array access will simply throw warnings/notices, right?

inoas:
Closing in favour of #181

PR #179, opened 2018-02-06
loginUrl is expected to be string not array

rochamarcelo:


rochamarcelo:
Array did not worked for me.

dereuromark:
Yeah maybe we should set a better default for CakePHP default usage.

PR #180, opened 2018-03-05
Convenience helper for the view layer

burzum:


src/View/Helper/IdentityHelper.php
```diff
@@ -0,0 +1,117 @@
...
+            $this->_userData = $this->_userData->toArray();
+        }
+    }
+
+    /**
+     * Gets the id of the current logged in identity
+     *
+     * @return int|null|string
+     */
+    public function getId()
```
robertpustulka: IdentityInterface exposes `getIdentifier()` method which does exactly that.

src/View/Helper/IdentityHelper.php
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    /**
+     * Checks if a user is logged in
+     *
+     * @return boolean
+     */
+    public function isLoggedIn()
+    {
+        return (!empty($this->_userData));
```
robertpustulka: I think you should check here the authentication result instead.

src/View/Helper/IdentityHelper.php
```diff
@@ -0,0 +1,117 @@
...
+        }
+
+        if (!$identity instanceof IdentityInterface) {
+            throw new RuntimeException(sprintf('Identity found in request does not implement %s', IdentityInterface::class));
+        }
+
+        $this->_userData = $identity->getOriginalData();
+
+        if ($this->_userData instanceof Entity) {
+            $this->_userData = $this->_userData->toArray();
```
markstory: Do you need to keep a copy of the data? Going back to the source data each time shouldn't be expensive.

src/View/Helper/IdentityHelper.php
```diff
@@ -0,0 +1,117 @@
...
+     * @param string $key
+     * @return mixed
+     */
+    public function get($key = null)
+    {
+        if ($key === null) {
+            return $this->_userData();
+        }
+
+        return Hash::get($this->_userData, $key);
```
markstory: Given that all identities implement `ArrayAccess` you could pass the original user and not the copy here.

tests/TestCase/View/Helper/IdentityHelperTest.php
```diff
@@ -0,0 +1,79 @@
...
+ */
+class IdentityHelperTest extends TestCase
+{
+
+    /**
+     * @inheritDoc
+     */
+    public function setUp()
+    {
+        parent::setUp();
```
markstory: This method isn't doing much.

src/View/Helper/IdentityHelper.php
```diff
@@ -0,0 +1,117 @@
...
+
+    /**
+     * This check can be used to tell if a record that belongs to some user is the
+     * current logged in user
+     *
+     * @param int $userId
+     * @param string $field Name of the field in the user record to check against, id by default
+     * @return boolean
+     */
+    public function is($userId, $field = 'id')
```
markstory: This method doesn't look like it is covered by any tests. Could we have another test or remove the method?

src/View/Helper/IdentityHelper.php
```diff
@@ -0,0 +1,117 @@
...
+    ];
+
+    /**
+     * User data
+     *
+     * @array
+     */
+    protected $_userData = [];
+
+    /**
```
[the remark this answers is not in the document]
burzum: @markstory why is stickler still showing this?

src/View/Helper/IdentityHelper.php
burzum: Done.

src/View/Helper/IdentityHelper.php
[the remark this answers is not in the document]
markstory: Because this line hasn't been 'changed' so github won't collapse this comment.

src/View/Helper/IdentityHelper.php
```diff
@@ -0,0 +1,121 @@
...
+     */
+    public function is($id, $field = 'id')
+    {
+        return ($id === $this->get($field));
+    }
+
+    /**
+     * Gets user data
+     *
+     * @param string $key Key of something you want to get from the identity data
```
dereuromark: |null

src/View/Helper/IdentityHelper.php
```diff
@@ -0,0 +1,121 @@
...
+        }
+
+        if (!$this->_identity instanceof IdentityInterface) {
+            throw new RuntimeException(sprintf('Identity found in request does not implement %s', IdentityInterface::class));
+        }
+
+        $this->_userData = $this->_identity->getOriginalData();
+
+        if ($this->_userData instanceof Entity) {
+            $this->_userData = $this->_userData->toArray();
```
robertpustulka: Why? Identity implements ArrayAccess. $_userData property seems reduntant in this helper.

src/View/Helper/IdentityHelper.php
```diff
@@ -0,0 +1,121 @@
...
+
+    /**
+     * This check can be used to tell if a record that belongs to some user is the
+     * current logged in user
+     *
+     * @param int|string $id Identity id to check against
+     * @param string $field Name of the field in the identity data to check against, id by default
+     * @return bool
+     */
+    public function is($id, $field = 'id')
```
robertpustulka: This method doesn't look like it should belong here. Checking if one user is another user is authorization. We didn't want to mix those concepts.

burzum: So you want a full powered authorization plugin instead of this convenient check? This is pretty much all you need to conditionally show if the user is logged in or not.

src/View/Helper/IdentityHelper.php
robertpustulka: But identity does implement ArrayAccess so it should work with Hash::get()

https://github.com/cakephp/authentication/blob/master/src/IdentityInterface.php#L22

src/View/Helper/IdentityHelper.php
burzum: Good catch but why is none of our cs tools finding this? :)

src/View/Helper/IdentityHelper.php
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    /**
+     * Gets the id of the current logged in identity
+     *
+     * @return int|null|string
+     */
+    public function getId()
+    {
+        if (empty($this->_identity)) {
```
robertpustulka: `$this->_identity !== null`

src/View/Helper/IdentityHelper.php
robertpustulka: Is this method for checking whether a user is logged in? Docs say otherwise.

burzum: My bad wording: You have pretty often the case when you want to check if an user can see something or not. This might depend on another field as well. The method provides an easy way to do that, that is intuitive accessible and doesn't require a second plugin nor knowledge about it. We could drop the whole helper and just add getIdentity() to the view class directly?

markstory: Should we be guiding people to use Authorization policies in those scenarios? This method covers a very simple use case, but I feel like we should be recommending people with more complex requirements use the authorization plugin.

burzum: Lets stick a comment in the doc block and recommend it there?

markstory: That works for me.

dereuromark: sounds good.

burzum: @markstory can we merge it now? I've added the doc block.

dereuromark:
Tests are only failing because of deprecations, right @markstory ?

markstory:
@dereuromark Looks like it. But there is no documentation on this helper/feature either.

dereuromark:
@markstory I added some basic docs, feel free to re-word or complete etc.

docs/View-Helper.md
```diff
@@ -0,0 +1,28 @@
+# View Helper
+
+In your AppView, load the Helper as
+```php
+$this->loadHelper('Authentication.Identity');
+```
+
+For very simple checking whether the user is logged in you can use
+```php
+if ($this->Identify->isLoggedIn()) {
```
robertpustulka: Typo, should be `$this->Identity`

docs/View-Helper.md
```diff
@@ -0,0 +1,28 @@
...
+    ...
+}
+```
+
+Getting user data is as as easy as
```
robertpustulka: Double "as"

burzum:
@markstory can we merge it now?

PR #181, opened 2018-03-06
more strict on nulls, ::class over string-class, early returns

inoas:
I hope I got the (hopefully) better parts of https://github.com/cakephp/authentication/pull/175 isolated.
I still have a very bad feeling about `isset()` and `empty()` TBH and I feel things my gut feeling is that things might bite back in 7.1/7.2.

### DIFF
--- a/src/Authenticator/HttpDigestAuthenticator.php
+++ b/src/Authenticator/HttpDigestAuthenticator.php
@@ -78,7 +78,7 @@ class HttpDigestAuthenticator extends HttpBasicAuthenticator
     public function authenticate(ServerRequestInterface $request, ResponseInterface $response)
     {
         $digest = $this->_getDigest($request);
-        if (empty($digest)) {
+        if ($digest === null) {
             return new Result(null, Result::FAILURE_CREDENTIALS_MISSING);
         }
 
@@ -114,7 +114,7 @@ class HttpDigestAuthenticator extends HttpBasicAuthenticator
      * Gets the digest headers from the request/environment.
      *
      * @param \Psr\Http\Message\ServerRequestInterface $request The request that contains login information.
-     * @return array Array of digest information.
+     * @return array|null Array of digest information.
      */
     protected function _getDigest(ServerRequestInterface $request)
     {
@@ -127,7 +127,7 @@ class HttpDigestAuthenticator extends HttpBasicAuthenticator
             }
         }
         if (empty($digest)) {
-            return [];
+            return null;
         }
 
         return $this->parseAuthData($digest);
@@ -209,7 +209,7 @@ class HttpDigestAuthenticator extends HttpBasicAuthenticator
         ];
 
         $digest = $this->_getDigest($request);
-        if ($digest && isset($digest['nonce']) && !$this->validNonce($digest['nonce'])) {
+        if ($digest !== null && isset($digest['nonce']) && !$this->validNonce($digest['nonce'])) {
             $options['stale'] = true;
         }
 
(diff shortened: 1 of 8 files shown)


src/Authenticator/HttpDigestAuthenticator.php
```diff
@@ -114,7 +114,7 @@ public function authenticate(ServerRequestInterface $request, ResponseInterface
      * Gets the digest headers from the request/environment.
      *
      * @param \Psr\Http\Message\ServerRequestInterface $request The request that contains login information.
-     * @return array Array of digest information.
+     * @return array|null Array of digest information.
```
dereuromark: Do we need the null alternative here? I would simply do `if (!$result)` to check for emptiness - the strict ones can use `$result === []`. The less return types the better, not?

src/Authenticator/TokenAuthenticator.php
```diff
@@ -98,11 +98,11 @@ protected function getTokenFromQuery(ServerRequestInterface $request, $queryPara
     {
         $queryParams = $request->getQueryParams();
 
-        if (isset($queryParams[$queryParam])) {
-            return $queryParams[$queryParam];
+        if (!array_key_exists($queryParam, $queryParams)) {
```
dereuromark: why array_key_exists? It is either non-null or irrelevant IMO.

src/Authenticator/TokenAuthenticator.php
```diff
@@ -98,11 +98,11 @@ protected function getTokenFromQuery(ServerRequestInterface $request, $queryPara
...
+            return null;
         }
 
-        return null;
+        return $queryParams[$queryParam];
```
dereuromark: :+1: on inverting.

src/Authenticator/HttpDigestAuthenticator.php
inoas: You had the same question on the original ticket:
https://github.com/cakephp/authentication/pull/175#pullrequestreview-84302555
Is the conclusion still the same?

src/Authenticator/TokenAuthenticator.php
inoas: Well I thought it doesn't matter what's inside, null, 0, '0', '' - all ok, and middleware can do that (mine casts to real types for API things) a token with `(int)0` should be legit.

Let me ask the other way around: What's the benefit of using isset here which does all that autocasting? Does array_key_exists hurt?

src/Authenticator/HttpDigestAuthenticator.php
dereuromark: OK, sry, didnt remember that one.

src/Authenticator/TokenAuthenticator.php
dereuromark: isset doesnt do any auto-cast. array_key_exists() usually goes a) too fur, b) is 10-100x slower.

inoas: I still want to see the performance difference on 7.2/7.1 but going to remove this now.

src/Authenticator/JwtAuthenticator.php
```diff
@@ -126,11 +127,13 @@ public function getPayload(ServerRequestInterface $request = null)
         $payload = null;
         $token = $this->getToken($request);
 
-        if ($token) {
+        if ($token !== null) {
             $payload = $this->decodeToken($token);
         }
 
-        return $this->payload = $payload;
```
inoas: Is this good style?
If not, could it be caught by some checker in future?

dereuromark: It is not necessarily something of an issue.

markstory: The previous way uses 2 fewer lines of code, and is readable in my opinion. This form is fine too though.

inoas: One last thing that bugs me is that a getters modifies the instance property...

Issue #188, opened 2018-03-29
Silent allow Issue.


dereuromark:
In a fresh 3.next app wiring this up as documented leads to
```
	[protected] _result => object(Authentication\Authenticator\Result) {
		[protected] _status => 'FAILURE_OTHER'
		[protected] _data => null
		[protected] _errors => [
			(int) 0 => 'Login URL `http://app.local/public` did not match `/users/login`.'
		]
	}
```
inside the Result, but the controllers are still accessible.
Shouldn't it in that case through an exception maybe?

markstory:
>  Shouldn't it in that case through an exception maybe?

If you apply the middleware to your entire application, how would you go about rendering a login page, or a response that is publicly accessible?

The tutorial could be updated to add a basic authorization checks to ensure that actions have a logged in user when required.

dereuromark:
That would sure be helpful. 
Coming from AuthComponent it disallows by default. Middelware is too early, that is a fair point.

As per documentation I added the AuthenticationComponent:

        $this->loadComponent('Authentication.Authentication', [
            'logoutRedirect' => '/users/login'  // Default is false
        ]);

And at that point I would expect it to do some sanity check here.

markstory:
@robertpustulka and I had a few discussions about this topic. Would you think it reasonable for the authentication component to have a method like `requireIdentity()` that raises an exception if the request does not contain a calid identity?

I think something like that covers the use case you're describing and doesn't introduce any authorisation baggage.

dereuromark:
This will sure help to avoid any user land implementations that will otherwise be necessary here to workaround this. It will then be documented and clear.

But don't you think the "former use case" of whitelisting public ones could be even more helpful?
This way you can't "forget" to protect certain controllers, actions or plugins - exposing certain data by accident.
The "default" should be the safer way IMO.

markstory:
An alternate approach would be to default to requiring an identity and provide a `allowUnauthenticated($actions)` that can be called in a controller's beforeFilter() to mark specific actions as not requiring an identity.

dereuromark:
Yes exactly, people coming from existing 3.x applications sure would rather expect this whitelisting behavior and might even be able to upgrade to this easier then.

markstory:
Closing as we can continue this discussion in #191

PR #191, opened 2018-03-30
Unauthenticated actions

markstory:
This adds a *very* rudimentary form of access control to this plugin. By default the component will require an identity to be present in all requests. This behavior can be disabled via a setting or by using the `allowUnauthenticated()` method to whitelist the controller actions.

Refs #188

### DIFF
--- a/docs/Quick-start-and-introduction.md
+++ b/docs/Quick-start-and-introduction.md
@@ -106,7 +106,16 @@ $this->loadComponent('Authentication.Authentication', [
 ]);
 ```
 
-## Accessing the user / identity data
+Once loaded, the `AuthenticationComponent` will require that all actions have an
+authenticated user present, but perform no other access control checks. You can
+disable this check for specific actions using `allowUnauthorized()`:
+
+```php
+// In your controller's beforeFilter method.
+$this->Authentication->allowUnauthorized(['view']);
+```
+
+### Accessing the user / identity data
 
 You can get the authenticated identity data using the authentication component:
 
@@ -120,7 +129,7 @@ You can also get the identity directly from the request instance:
 $user = $request->getAttribute('identity');
 ```
 
-## Checking the login status
+### Checking the login status
 
 You can check if the authentication process was successful by accessing the result
 object:
@@ -150,7 +159,7 @@ The result sets objects status returned from `getStatus()` will match one of the
 
 The error array returned by `getErrors()` contains *additional* information coming from the specific system against which the authentication attempt was made. For example LDAP or OAuth would put errors specific to their implementation in here for easier logging and debugging the cause. But most of the included authenticators don't put anything in here.
 
-## Clearing the identity / logging the user out
+### Clearing the identity / logging the user out
 
 To log an identity out just do:
 
--- a/docs/Quick-start-and-introduction.md
+++ b/docs/Quick-start-and-introduction.md
@@ -108,11 +108,11 @@ $this->loadComponent('Authentication.Authentication', [
 
 Once loaded, the `AuthenticationComponent` will require that all actions have an
 authenticated user present, but perform no other access control checks. You can
-disable this check for specific actions using `allowUnauthorized()`:
+disable this check for specific actions using `allowUnauthenticated()`:
 
 ```php
 // In your controller's beforeFilter method.
-$this->Authentication->allowUnauthorized(['view']);
+$this->Authentication->allowUnauthenticated(['view']);
 ```
 
 ### Accessing the user / identity data
--- a/src/Controller/Component/AuthenticationComponent.php
+++ b/src/Controller/Component/AuthenticationComponent.php
@@ -151,6 +151,16 @@ class AuthenticationComponent extends Component implements EventDispatcherInterf
     }
 
     /**
+     * Get the current list of actions that don't require authentication.
+     *
+     * @return array
+     */
+    public function getUnauthenticatedActions()
+    {
+        return $this->unauthenticatedActions;
+    }
+
+    /**
      * Gets the result of the last authenticate() call.
      *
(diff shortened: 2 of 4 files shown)


src/Controller/Component/AuthenticationComponent.php
```diff
@@ -89,6 +109,47 @@ public function beforeFilter()
...
+        }
+        $request = $this->getController()->request;
+        $action = $request->getParam('action');
+        if (in_array($action, $this->unauthenticatedActions)) {
+            return;
+        }
+
+        $identity = $request->getAttribute('identity');
+        if (!$identity) {
+            throw new UnauthorizedException([]);
```
robertpustulka: Should it throw MissingIdentityException?

dereuromark: Or `UnauthenticatedException` - but MissingIdentityException as generic one can also be fine.

dereuromark: With a `loginRedirect` or `loginAction` present, wouldn't it make sense to redirect to login then instead from the component?
The exception is only useful for stateless (api) access, right?

robertpustulka: Redirection is performed from a middleware using unauthorized handlers.

markstory: @robertpustulka You're right, `MissingIdentity` would be better, as it is handled by the Unauthorized handlers.

docs/Quick-start-and-introduction.md
```diff
@@ -106,7 +106,16 @@ $this->loadComponent('Authentication.Authentication', [
...
 ```
 
-## Accessing the user / identity data
+Once loaded, the `AuthenticationComponent` will require that all actions have an
+authenticated user present, but perform no other access control checks. You can
+disable this check for specific actions using `allowUnauthorized()`:
+
+```php
+// In your controller's beforeFilter method.
+$this->Authentication->allowUnauthorized(['view']);
```
dereuromark: allowUnauthorized() inside the Authentication module might a sth to trip over?
Should we call it `allowUnauthenticated()` or just `allowPublic()`?

markstory: The docs were wrong, and fixed in 1f325dc. The method is `allowUnauthenticated()`. I was worried that `allowPublic()` could be confused with PHP method visibility.

src/Controller/Component/AuthenticationComponent.php
```diff
@@ -89,6 +109,47 @@ public function beforeFilter()
...
+    /**
+     * Set the list of actions that don't require an authentication identity to be present.
+     *
+     * Actions not in this list will require an identity to be present. Any
+     * valid identity will pass this constraint.
+     *
+     * @param array $actions The action list.
+     * @return $this
+     */
+    public function allowUnauthenticated(array $actions)
```
ADmad: Should this have a merge option to allow adding to existing list? Unless there's a method to retrieve existing list so that user can merge himself and reset.

ADmad: For e.g. I would like to allow 'view' always through AppController and then additional action inside a specific controller.

dereuromark: Would it then merge by default? Calling the param $overwrite (default false) like in other places in core?

dereuromark: Why a getter? Wouldnt this be

    allowUnauthenticated(array $actions, $merge = false)

?

robertpustulka: I think I'd prefer a seperate method for this. Or we could just utilise instance config for this `$this->Authentication->setConfig('allowUnauthenticated', ['index']);`

dereuromark: Handling/Merging on this level is not a good idea IMO. Better to have either dedicated methods or just the bool 2nd param.

robertpustulka: Right. `mergeUnauthenticated(['index'])` is cleaner in terms of what it does than `allowUnauthenticated(['index'], true)`

dereuromark: Well, you could always do `allowUnauthenticated(['index'], ['merge' => true])` :)

ADmad: Merging to existing list will be the most common usage. So it should be possible to do that with minimal code :slightly_smiling_face:.

dereuromark: So mergeUnauthenticated() is the method name to move forward here?

markstory: Added.

src/Controller/Component/AuthenticationComponent.php
robertpustulka: Should it be a new `UnauthenticatedException` ?
`UnauthorizedException` is used for http auth AFAIR

dereuromark: At this point definitely not authorize, but authenticate, yes.

markstory: I was attempting to reuse the [existing exception](https://github.com/cakephp/authentication/blob/master/src/Authenticator/UnauthorizedException.php) which generates a 401 error, which is what I think we want here too.

robertpustulka:
For anyone trying to integrate this change with Authorization plugin unauthoroized handers, this middleware might come in handy:

```php
$middleware->add(function($req, $res, $next) {
    try {
        return $next($req, $res);
    } catch (UnauthorizedException $e) {
        throw new MissingIdentityException(['identity'], null, $e);
    }
});
```

Issue #195, opened 2018-04-06
Add config identityAttribute


rochamarcelo:
Could we add a config 'identityAttribute' in the following class?
https://github.com/cakephp/authentication/blob/d0face9e81b90814b1fa449f11056d098e299e54/src/AuthenticationService.php#L249

Something like we do at 
https://github.com/cakephp/authorization/blob/master/src/Middleware/AuthorizationMiddleware.php#L55

markstory:
@rochamarcelo Any chance you would have time to do this change?

rochamarcelo:
@markstory yes I'll do that

PR #196, opened 2018-04-08
#195 added config 'identityAttribute'

rochamarcelo:
for issue #195

### DIFF
--- a/src/AuthenticationService.php
+++ b/src/AuthenticationService.php
@@ -85,7 +85,8 @@ class AuthenticationService implements AuthenticationServiceInterface
     protected $_defaultConfig = [
         'authenticators' => [],
         'identifiers' => [],
-        'identityClass' => Identity::class
+        'identityClass' => Identity::class,
+        'identityAttribute' => 'identity',
     ];
 
     /**
@@ -218,7 +219,7 @@ class AuthenticationService implements AuthenticationServiceInterface
         }
 
         return [
-            'request' => $request->withoutAttribute('identity'),
+            'request' => $request->withoutAttribute($this->getConfig('identityAttribute')),
             'response' => $response
         ];
     }
@@ -246,7 +247,7 @@ class AuthenticationService implements AuthenticationServiceInterface
         }
 
         return [
-            'request' => $request->withAttribute('identity', $identity),
+            'request' => $request->withAttribute($this->getConfig('identityAttribute'), $identity),
             'response' => $response
         ];
     }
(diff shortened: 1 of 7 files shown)


tests/TestCase/AuthenticationServiceTest.php
```diff
@@ -241,6 +308,103 @@ public function testPersistIdentity()
...
+            'florian',
+            $result['request']->getAttribute('session')->read('Auth.username')
+        );
+
+        $identity = $result['request']->getAttribute('customIdentity');
+        $this->assertInstanceOf(IdentityInterface::class, $identity);
+        $this->assertEquals($data, $identity->getOriginalData());
+
+        $data = ['username' => 'johndoe'];
+        $this->assertEquals($data, $result['request']->getAttribute('identity'));
```
markstory: Why is this attribute set if there is a custom attribute in use.

markstory: Ok.

robertpustulka:
Could you add a similar config key to the middleware and the component?

rochamarcelo:
For middleware I would need to update the constructor, is it a good idea? If the getConfig method was defined at AuthenticationServiceInterface I could do 'service->getConfig('identityAttribute'); . What approach do you recommend?

robertpustulka:
@rochamarcelo I don't think exposing `getConfig` is a good idea.

For the middleware I suppose we could still change the second argument to be a config array and use `InstanceConfigTrait` (similar to the authz plugin middleware).

This could be done in a BC way:
```php
public function __construct($subject, $config = [])
{
    if (is_string($config)) {
        $config = ['name' => $config];
    }
    $this->setConfig($config);
    ...
}

**Comment by rochamarcelo** [comment]
@robertpustulka @markstory how can I fix the following travis ci build?

 ------ ------------------------------------------------------------- 
  Line   src/Middleware/AuthenticationMiddleware.php                  
 ------ ------------------------------------------------------------- 
  66     Call to function is_string() will always evaluate to false.  
 ------ -------------------------------------------------------------

**Comment by rochamarcelo** [comment]
@robertpustulka thanks. I fixed the param type

File: src/AuthenticationService.php [hunk 2]
```diff
@@ -85,7 +85,8 @@ class AuthenticationService implements AuthenticationServiceInterface
     protected $_defaultConfig = [
         'authenticators' => [],
         'identifiers' => [],
-        'identityClass' => Identity::class
+        'identityClass' => Identity::class,
```
burzum: I don't think this should be here. It should be IdentityInterfacen::class and when we use the interface there is no need to make it configurable. We want to enforce the interface as a minimum common API of identities. There should be no way to bypass that.

Issue #204, opened 2018-04-21
Get the last result of each Authenticator


rochamarcelo:
Hi,

Could we add a method to get the last result of each authenticator processed? Today we only have one result (request attribute authenticationResult) but it would be a good idea to know why each authenticator failed.

Thanks.

markstory:
What is the use case you're interested in for this?

rochamarcelo:
Lets say I have four authenticators in this order: Session, Form, Custom, Cookie.

If the user is not successfully authenticated, the result object in request is from Cookie authenticator (last one) but  I want to know the failed result produced in Form and Custom authenticator to give a better feedback to the user.

markstory:
Ok.

dereuromark:
PR open.

PR #205, opened 2018-04-24
Get the last result of each Authenticator

rochamarcelo:
For issue #204 , please let me know if this is okay.

### DIFF
--- a/src/Authenticator/AbstractAuthenticator.php
+++ b/src/Authenticator/AbstractAuthenticator.php
@@ -88,7 +88,7 @@ abstract class AbstractAuthenticator implements AuthenticatorInterface
     /**
      * Get the last authentication result instance stored or null if none.
      *
-     * @return ResultInterface|null
+     * @return \Authentication\Authenticator\ResultInterface|null
      */
     public function getLastResult()
     {
@@ -102,7 +102,7 @@ abstract class AbstractAuthenticator implements AuthenticatorInterface
      * @param string $status Status constant equivalent.
      * @param array $messages Messages.
      * @throws \InvalidArgumentException When invalid identity data is passed.
-     * @return ResultInterface
+     * @return \Authentication\Authenticator\ResultInterface
      */
     protected function _buildLastResult($data, $status, array $messages = [])
     {
--- a/src/Authenticator/AbstractAuthenticator.php
+++ b/src/Authenticator/AbstractAuthenticator.php
@@ -44,6 +44,13 @@ abstract class AbstractAuthenticator implements AuthenticatorInterface
     protected $_identifier;
 
     /**
+     * Last authentication result stored.
+     *
+     * @var \Authentication\Authenticator\ResultInterface
+     */
+    protected $_lastResult;
+
+    /**
      * Constructor
      *
      * @param \Authentication\Identifier\IdentifierInterface $identifier Identifier or identifiers collection.
@@ -76,6 +83,32 @@ abstract class AbstractAuthenticator implements AuthenticatorInterface
         $this->_identifier = $identifier;
 
         return $this;
+    }
+
+    /**
+     * Get the last authentication result instance stored or null if none.
+     *
+     * @return ResultInterface|null
+     */
+    public function getLastResult()
+    {
+        return $this->_lastResult;
+    }
+
+    /**
+     * Build the result and set as last
+     *
+     * @param null|array|\ArrayAccess $data The identity data
+     * @param string $status Status constant equivalent.
+     * @param array $messages Messages.
+     * @throws \InvalidArgumentException When invalid identity data is passed.
+     * @return ResultInterface
+     */
+    protected function _buildLastResult($data, $status, array $messages = [])
+    {
+        $this->_lastResult = new Result($data, $status, $messages);
+
+        return $this->_lastResult;
     }
 
     /**
--- a/docs/Quick-start-and-introduction.md
+++ b/docs/Quick-start-and-introduction.md
@@ -149,6 +149,43 @@ if ($result->isValid()) {
 }
 ```
 
+### Checking login failures with multiple authenticators
+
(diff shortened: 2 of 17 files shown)


src/Authenticator/AbstractAuthenticator.php
```diff
@@ -78,6 +85,32 @@ public function setIdentifier(IdentifierInterface $identifier)
...
+    /**
+     * Build the result and set as last
+     *
+     * @param null|array|\ArrayAccess $data The identity data
+     * @param string $status Status constant equivalent.
+     * @param array $messages Messages.
+     * @throws \InvalidArgumentException When invalid identity data is passed.
+     * @return ResultInterface
+     */
+    protected function _buildLastResult($data, $status, array $messages = [])
```
saeideng: having this method does not guarantee that all custom class should use this method in their class

src/Authenticator/AbstractAuthenticator.php
```diff
@@ -78,6 +85,32 @@ public function setIdentifier(IdentifierInterface $identifier)
         return $this;
     }
 
+    /**
+     * Get the last authentication result instance stored or null if none.
+     *
+     * @return ResultInterface|null
+     */
+    public function getLastResult()
```
markstory: Does this need to go on the interface as well?

rochamarcelo: yeah, I think it does.

src/Authenticator/AbstractAuthenticator.php
rochamarcelo: any suggestions of how force the use?

saeideng: I'm not sure 
but maybe you able to  store it in `result` class

robertpustulka:
Not sure if this is a good approach. Maybe the authentication service should store all failed results named after the authenticator that returned them?

rochamarcelo:
@robertpustulka  @markstory based on the suggestion we could have a ResultCollection instead. It is enough for me. Is it okay to use the authenticator class as the key for the collection? Should we proceed this way?

markstory:
Going back to the use case, how do you see this new feature being used? Where do you want to access it, what would the documentation section have for code samples?

rochamarcelo:
**How I see the feature being used:**

The application allow authentication by token, custom form and remember me, when user access the login page and authentication failed, I want to say to the user one of these message based on authenticators result: 

- Token FAILURE_IDENTITY_NOT_FOUND:  'Token invalid or expired';
- CustomForm FAILURE_INVALID_RECAPTCHA: 'Invalid reCaptcha '
- CustomForm FAILURE_IDENTITY_NOT_FOUND: Username or password is incorrect'

 **Where I want to access it:**
In an controller action, it would be good to access it through authenticator service method where I can loop each result (having reference to authenticator)

**What would the documentation section have for code samples?**
Handle authentication errors

markstory:
Would you be able to update the documentation to cover the use case you have? Features without documentation are rarely used and poorly understood.

rochamarcelo:
Yes I can update documentation. Do you think people won't need to check all authenticators results?

markstory:
I don't know how frequently people need to find out which authentication strategies have failed. I personally haven't needed that but I don't work on any applications with multiple authentication strategies in the same url space.

burzum:
@markstory can we merge this or not?

src/Authenticator/AbstractAuthenticator.php
```diff
@@ -78,6 +85,32 @@ public function setIdentifier(IdentifierInterface $identifier)
...
+    }
+
+    /**
+     * Build the result and set as last
+     *
+     * @param null|array|\ArrayAccess $data The identity data
+     * @param string $status Status constant equivalent.
+     * @param array $messages Messages.
+     * @throws \InvalidArgumentException When invalid identity data is passed.
+     * @return ResultInterface
```
dereuromark: FQCN please.

src/Authenticator/AbstractAuthenticator.php
robertpustulka: Since we have `getLastResult()` method maybe this should be just a setter (setLastResult), not a combined factory method/setter?

src/Authenticator/AbstractAuthenticator.php
```diff
@@ -78,6 +85,32 @@ public function setIdentifier(IdentifierInterface $identifier)
         return $this;
     }
 
+    /**
+     * Get the last authentication result instance stored or null if none.
+     *
+     * @return ResultInterface|null
```
dereuromark: Same here

robertpustulka:
To be honest I don't like this implementation. It makes authenticators stateful, they need to remember the last result they return. Can we come up with another approach? Shouldn't the authentication service be aware of results that each authenticator return?

dereuromark:
Whats the downside of the statefulness? After all, this is not something that should be executed multiple times per request (compared to other business logic etc).

src/Authenticator/AbstractAuthenticator.php
```diff
@@ -78,6 +85,32 @@ public function setIdentifier(IdentifierInterface $identifier)
         return $this;
     }
 
+    /**
+     * Get the last authentication result instance stored or null if none.
+     *
+     * @return \Authentication\Authenticator\ResultInterface|null
+     */
+    public function getLastResult()
```
robertpustulka: Should this method be added to the interface?

rochamarcelo: Sorry for delay (was on vacation) Yes it should, and I just added. Please let me know if something else should change/improve.

dereuromark:
So what do you think? Is this mergeable etc?

ADmad:
I too don't like the authenticators storing the last result. Why can't the authentication service aggregate the results as it runs the authenticators?

PR #206, opened 2018-05-03
Updated login to include failure flash message

jeremyharris:


docs/Migration-from-the-AuthComponent.md
```diff
@@ -33,9 +33,17 @@ The AuthenticationMiddleware will handle checking and setting the identity based
 ```php
 function login()
 {
+    $result = $this->Authentication->getResult();
+
     // regardless of POST or GET, redirect if user is logged in
-    if ($this->Authentication->getResult()->isValid()) {
-        return $this->redirect('/');
+    if ($result->isValid()) {
+        $redirect = $this->request->getQuery('redirect', ['controller' => 'Pages', 'action' => 'display', 'home']);
```
ADmad: Perhaps use string URL instead of array? We are trying to keep this plugin decoupled from Cake.

ADmad: But meh, the whole example is Cake specific.

dereuromark: I would rather see more URL arrays then just strings, especially in the CakePHP scope.

markstory: Or named routes :bike:

markstory:
Merging as the example code is a useful improvement over what as there before.

Issue #208, opened 2018-05-05
Docs missing for login redirect on unauthenticated


dereuromark:
https://github.com/cakephp/authentication/blob/master/docs/Migration-from-the-AuthComponent.md

Currently you would just get a `Authentication\Authenticator\UnauthorizedException` exception page rendered, no redirect to login.
There seems to be a part of the docs missing doing that - maybe by catching the exception in the lower levels (middleware?) and doing the redirect.

Would be nice if the plugin offered an easy way to do that (e.g. activating/calling a specified method).

markstory:
Pull request open now.

PR #209, opened 2018-05-08
Add unauthenticated redirects

markstory:
Migrating redirect logic from AuthComponent was not easy, and would require custom application code. I thought it would be simpler for users if that middleware was packaged in the plugin and not in the docs.

Refs #208


src/Middleware/UnauthorizedRedirectMiddleware.php
```diff
@@ -0,0 +1,63 @@
...
+     * @return \Psr\Http\Message\ResponseInterface A response.
+     */
+    public function __invoke(ServerRequestInterface $request, ResponseInterface $response, $next)
+    {
+        try {
+            return $next($request, $response);
+        } catch (UnauthorizedException $e) {
+            return $response
+                ->withStatus(301)
+                ->withHeader('Location', $this->target);
```
jeremyharris: Should this allow cake array URLs as well, similar to how there's a UrlChecker for cake URLs? That would also let users have one login url that they can use throughout the various middleware.

I also noticed that the authorization middleware uses regular string urls, so there's a tad bit of inconsistency there. Perhaps we can add support for cake urls in the unauthorizedHandler too?

docs/Migration-from-the-AuthComponent.md
```diff
@@ -151,3 +151,35 @@ $this->Authentication->allowUnauthenticated(['view']);
...
+    // Various other middlewares for error handling, routing etc. added here.
+
+    // Add the authentication middleware
+    $authentication = new AuthenticationMiddleware($this);
+
+    // Add authentication
+    $middlewareQueue->add($authentication);
+
+    // Add unauthorized redirect handling
+    $middleware->add(new UnauthorizedRedirectMiddleware('/users/login'));
```
saeideng: maybe `$middlewareQueue`

src/Middleware/UnauthorizedRedirectMiddleware.php
markstory: The caller could use `Router::url()` before passing the URL into the middleware though.

markstory:
> If using the unauthorizedHandler in the authorization plugin to redirect when a missing identity exception is thrown, is it just redundant to use this middleware then

I think both middlewares have similar but different class names :cry:

src/Middleware/UnauthorizedRedirectMiddleware.php
jeremyharris: That is what I was thinking, but then why do we have `CakeRouterUrlChecker` to begin with? It's a tad confusing that within the same plugin some areas allow Cake array URLs and others don't.

If I had it my way, we wouldn't support them at all and just tell people to use the router. Would require less documentation too. But I think the inconsistency will lead to support issues.

markstory: > That is what I was thinking, but then why do we have CakeRouterUrlChecker to begin with? 

I don't know. I didn't consider having people resolve their URLs before now.

docs/Migration-from-the-AuthComponent.md
```diff
@@ -151,3 +151,36 @@ $this->Authentication->allowUnauthenticated(['view']);
 ```
 
 Each call to `allowUnauthenticated()` will overwrite the current action list.
+
+# Migrating Unauthorized Redirects
```
dereuromark: I am still in favor of using the correct high level word "Unauthenticated" everywhere.
It just confuses me and probably others as at this point - especially since we do that distinction even with the two separate plugins - there is not even authentication happened yet. And everyone learns Authorization comes after successful Authentication.

docs/Migration-from-the-AuthComponent.md
```diff
@@ -151,3 +151,36 @@ $this->Authentication->allowUnauthenticated(['view']);
...
+
+By default `AuthComponent` redirects users back to the login page when
+authentication is required. In contrast, the `AuthenticationComponent` in this 
+plugin will raise an exception in this scenario. You can convert this exception
+into a redirect by using the `UnauthorizedRedirectMiddleware` in your application.
+
+```php
+// in src/Application.php
+use Authentication\Middleware\AuthenticationMiddleware;
+use Authentication\Middleware\UnauthorizedRedirectMiddleware;
```
dereuromark: Same then here.

docs/Migration-from-the-AuthComponent.md
```diff
@@ -151,3 +151,36 @@ $this->Authentication->allowUnauthenticated(['view']);
...
+    // Various other middlewares for error handling, routing etc. added here.
+
+    // Add the authentication middleware
+    $authentication = new AuthenticationMiddleware($this);
+
+    // Add authentication
+    $middlewareQueue->add($authentication);
+
+    // Add unauthorized redirect handling
+    $middlewareQueue->add(new UnauthorizedRedirectMiddleware('/users/login'));
```
dereuromark: Can we reuse a central config here inside the middleware instead? Would be more DRY, I would also prefer the default array syntax for the CakePHP use case.

markstory: The authentication plugin doesn't have/use centralized configuration thus far.

docs/Migration-from-the-AuthComponent.md
markstory: Changed the name of the middleware to `UnauthenticatedRedirectMiddleware`. I'll rename the exception separately.

robertpustulka:
What about this feature now?https://github.com/cakephp/authorization/blob/master/docs/Middleware.md#handling-unauthorized-requests

This serves no purpose as Authorization plugin no longer throws an exception on missing identities. Unauthorized handlers handle only exceptions thrown by Authorization plugin (descendants of `Authorization\Exception\Exception`). We could change that behavior so all kinds of exceptions could be handled. The handler could be configured to catch `Authentication\Authenticator\UnauthorizedException` then.

Or maybe we could port the handler stuff from authorization to authentication.

src/Middleware/UnauthenticatedRedirectMiddleware.php
```diff
@@ -0,0 +1,65 @@
...
+     * @param \Psr\Http\Message\ServerRequestInterface $request The request.
+     * @param \Psr\Http\Message\ResponseInterface $response The response.
+     * @param callable $next The next middleware to call.
+     * @return \Psr\Http\Message\ResponseInterface A response.
+     */
+    public function __invoke(ServerRequestInterface $request, ResponseInterface $response, $next)
+    {
+        try {
+            return $next($request, $response);
+        } catch (UnauthorizedException $e) {
```
robertpustulka: Won't this redirect HTTP authentication as well?

dereuromark:
I am for handling the exceptions and properly redirecting to login.
Using this plugin standalone requires this to work smoothly IMO.

dereuromark:
unauthorized should be completely in the other plugin, if we do that strict separation.
This one here only knows about either logged in or not (guest), binary, nothing else. Right?

robertpustulka:
> unauthorized should be completely in the other plugin, if we do that strict separation.

I'm talking about an exception for `401 Unauthorized` in case of HTTP authentication.

dereuromark:
OK. I thought 401 is still covered under "unauthenticated" as the status code message is misleading, it is per specs the case of not being logged in. But I can be mistaken.

dereuromark:
Hm, UnauthenticatedException and StatelessUnauthenticatedException then maybe? Both with error code 401 of course.

robertpustulka:
@markstory do you want to continue with this or maybe we should port the more advanced solution from authorization plugin as it seems more useful here now?

dereuromark:
@robertpustulka We need to continue here. If the separation of plugins should make any sense, this one here needs the proper 401 handling completely on its own, without the need of any 2nd plugin.

robertpustulka:
Yes, that's why I'm talking about moving the handler code from Authorization plugin to this plugin. Redirect handlers are now useless there.

dereuromark:
Authorization plugin then needs Authentication, which makes sense.
And the other way around there is (and should be) no dependency.

src/Middleware/UnauthenticatedRedirectMiddleware.php
markstory: It could, but if you stack the middleware correctly it won't.

markstory:
> So in my understanding we need two exceptions - one for stateless auth so the browser asks for credentials and one for missing identity so the app redirects to the login page.

Separating the behavior for stateless authentication (challenge header) and other use cases (redirect) sounds like a good idea. If we fix up the exception hierarchy then this new middleware would not be necessary, as the existing `AuthenticationMiddleware` can handle both scenarios.

The code in Authorization can stay where it is as it handles a different case (authenticated, but permission denied). We should provide options to convert that case into a redirect or an error page.

markstory:
I've updated the middleware to handle unauthenticated users now. I'll simplify URL handling as discussed on slack (string URLs only) in a follow up pull request.

docs/Migration-from-the-AuthComponent.md
```diff
@@ -157,30 +157,25 @@ Each call to `allowUnauthenticated()` will overwrite the current action list.
...
-use Authentication\Middleware\UnauthenticatedRedirectMiddleware;
 
 public function middleware($middlewareQueue)
 {
     // Various other middlewares for error handling, routing etc. added here.
 
     // Add the authentication middleware
-    $authentication = new AuthenticationMiddleware($this);
+    $authentication = new AuthenticationMiddleware($this, [
+        'unauthorizedRedirect' => Router::url('users:login')
```
robertpustulka: Should be `unauthenticatedRedirect`

markstory:
> Since the redirect url is configured in middleware there's no way of including the redirect query parameter in the url.

That should be doable.

markstory:
Any more feedback on this?

tests/TestCase/Middleware/AuthenticationMiddlewareTest.php
```diff
@@ -390,6 +391,60 @@ public function testNonSuccessfulAuthenticationWithChallenge()
...
+        $middleware = new AuthenticationMiddleware($this->service, [
+            'unauthenticatedRedirect' => '/users/login',
+        ]);
+
+        $next = function ($request, $response) {
+            throw new UnauthenticatedException();
+        };
+
+        $response = $middleware($request, $response, $next);
+        $this->assertEquals(301, $response->getStatusCode());
```
dereuromark: should we use Same() to make sure the type (int) is also true?

burzum:
@markstory let's merge it.

docs/Migration-from-the-AuthComponent.md
Zuluru: For anyone that finds their way here looking for a way to re-implement the old AuthComponent's "authError" functionality, you can add the following snippet to your middleware queue after the authentication middleware:

```
$middlewareQueue->add(function($request, $response, $next) {
	try {
		return $next($request, $response);
	} catch (\Authentication\Authenticator\UnauthenticatedException $ex) {
		$this->Flash('error', __('You must login to access full site functionality.'));
		throw $ex;
	}
});
```

PR #212, opened 2018-05-25
Service provider interface

burzum:


src/AuthenticationServiceProviderInterface.php
```diff
@@ -0,0 +1,33 @@
...
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\ServerRequestInterface;
+
+/**
+ *  AuthenticationServiceProviderInterface
+ */
+interface AuthenticationServiceProviderInterface
+{
+    /**
+     * Returns a service provider instance.
```
robertpustulka: Returns an authentication service, not a provider.

tests/TestCase/Middleware/AuthenticationMiddlewareTest.php
```diff
@@ -160,13 +136,12 @@ public function testApplicationAuthenticationRequestResponse()
 
         $application = $this->getMockBuilder(BaseApplication::class)
```
robertpustulka: BaseApplication does not implement provider interface. Should it be `TesApp\Application` here?

burzum: Change this and also tried to mock the interface, I'm always getting `1) Authentication\Test\TestCase\Middleware\AuthenticationMiddlewareTest::testApplicationAuthenticationRequestResponse
InvalidArgumentException: Subject must be an instance of `Authentication\AuthenticationServiceInterface` or `Cake\Core\HttpApplicationInterface`, `Mock_Application_985fb108` given.` this.

robertpustulka:
Should we provide something similar for the Authorization plugin?

burzum:
Yes

PR #201, opened 2018-04-18
Moved authentication service object check to controller startup

jeremyharris:
Fixes #200

PR #217, opened 2018-07-30
Move Authentication service checks for early failures.

robertpustulka:
`null` and `instanceof` checks should be executed early.

### DIFF
--- a/src/Controller/Component/AuthenticationComponent.php
+++ b/src/Controller/Component/AuthenticationComponent.php
@@ -85,6 +85,14 @@ class AuthenticationComponent extends Component implements EventDispatcherInterf
      */
     public function beforeFilter()
     {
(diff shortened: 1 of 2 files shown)


jeremyharris:
This is basically a revert of https://github.com/cakephp/authentication/pull/201 which may cause bake to start failing again.

robertpustulka:
@jeremyharris sorry, I wasn't aware of that.

Maybe the check should be moved to the `beforeFilter` callback? `beforeFilter` expectes that the `$this->_authentication` is valid, but this is checked later in `startup`.

jeremyharris:
@robertpustulka no worries! Just wanted to make you aware. I think `beforeFilter` will work just fine :)

markstory:
Good catch @jeremyharris . `beforeFilter()` would be a good place for these.

PR #221, opened 2018-08-07
Use `302` instead of `301` for redirecting unauthenticated requests.

ndm2:
The location isn't permanently moved, therefore using `301` isn't appropriate.

I'd say this is generally problematic. In my specific case it prevented my custom authenticator from running on every non-login URL request, where it would need to apply some logic for things to work properly.

### DIFF
--- a/src/Middleware/AuthenticationMiddleware.php
+++ b/src/Middleware/AuthenticationMiddleware.php
@@ -125,7 +125,7 @@ class AuthenticationMiddleware
                 $url = $this->getRedirectUrl($target, $request);
 
                 return $response
-                    ->withStatus(301)
+                    ->withStatus(302)
                     ->withHeader('Location', $url);
             }
             throw $e;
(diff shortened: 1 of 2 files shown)


ADmad:
> I'd say this is generally problematic. In my specific case it prevented my custom authenticator from running on every non-login URL request, where it would need to apply some logic for things to work properly.

I thought we were going to provide option to not do any redirection and let the developer take appropriate action by checking the authentication result.

ndm2:
Just to make it clear, by "_this is generally problematic_" I mean using a `301` for the redirect is generally problematic, as it's rather unpredictable how clients will behave.

For some reason browsers seem to drop the cached `301` when the new location (the logic action) redirects back to the `301`ed location. If the redirect after successful login however doesn't lead back to the original `301`ed entry location, for example because the redirect URL changes or gets lost for some reason, then the `301` is kept cacheed, and visiting the entry location will cause an infinite redirect loop if the default redirect URL happens to match the entry location.

Issue #230, opened 2018-12-04
AuthenticationMiddleware.queryParam generate URL without app directory


hiryu85:
If i use AuthenticationMiddleware with ```queryParam = 'redirect'``` as setting, middleware redirect me to a invalid URL.
Seems ignore application directory from URL, becase if i run cake application in root domain works fine.

## Example 1 (app not have directory, in root of domain) [WORKS]

http://myapp.test/foo/bar redirect me to http://myapp.test/users/login?redirect=http://myapp.test/foo/bar

## Example 2 (app have directory in url) [NOT WORKS]

http://myapp.test/v1/foo/bar  redirect me to http://myapp.test/v1/users/login?redirect=http://myapp.test/foo/bar but in redirect query parameter /v1/ disappears.

Zuluru:
I've run into this same issue. It's kind of mission critical for some of my clients to be able to run in a subdirectory of their main site. What I've found:

`ServerRequestFactory::marshalUriFromServer` adds `base` and `webroot` properties to the `$uri`, but those are non-standard extensions. Most specifically, they are not handled by `\Zend\Diactoros\Uri::__toString`. As a result, when `AuthenticationMiddleware::getRedirectUrl` is called, it generates a URL without the base in it. If it generated relative URLs, that might not be an issue (or it might; I'm not really sure), but given that it generates absolute URLs, it's absolutely an issue.

Given that the code is relying on a third-party library for URL generation, and that said library doesn't support functionality that we need, I'm not sure the best solution. Make a copy of the URL object, set the `path` to be `base` + `path`, and generate the redirect URL from that? I can put together a PR for that, if it's not considered too kludgy.

markstory:
> Make a copy of the URL object, set the path to be base + path, and generate the redirect URL from that? I can put together a PR for that, if it's not considered too kludgy.

I think that is a reasonable approach. Using `$uri->withPath()` will create a new instance for you so you won't need to clone the uri.

PR #232, opened 2019-01-15
Handle URL base

Zuluru:
Addresses issue #230.


Zuluru:
Hmm, looks like the `base` is being incorrectly set to the `vendor` folder during unit testing. Wasn't a problem until now. Will need a similar fix to the one in the main cakephp repo, I guess.

Zuluru:
We probably want to implement a better fix for the unit tests, like PR12076 in the cakephp repo. The fix used here is the previous, insufficient one from there, just to demonstrate that the Travis issues with the previous commit are in the unit tests rather than my changes, to allow discussion to proceed.

src/Middleware/AuthenticationMiddleware.php
```diff
@@ -146,7 +146,9 @@ protected function getRedirectUrl($target, ServerRequestInterface $request)
             return $target;
         }
 
-        $query = urlencode($param) . '=' . urlencode($request->getUri());
+        $uri = $request->getUri();
+        $uri = $uri->withPath($uri->base . $uri->getPath());
```
markstory: Should we make sure that the uri has a `base` property? Something like `if (property_exists($uri, 'base')) {` should work. My concern is that we could get vanilla Uri objects and hit runtime errors.

tests/bootstrap.php
```diff
@@ -57,3 +57,5 @@
 Plugin::load('Auth', [
     'path' => dirname(dirname(__FILE__)) . DS,
 ]);
+
+$_SERVER['PHP_SELF'] = '/';
```
markstory: This is reasonable to me. In a web context `PHP_SELF` would not be a file path to vendor/

src/UrlChecker/DefaultUrlChecker.php
```diff
@@ -102,6 +102,8 @@ protected function _getChecker(array $options = [])
      */
     protected function _getUrlFromRequest(UriInterface $uri, $getFullUrl = false)
     {
+        $uri = $uri->withPath($uri->base . $uri->getPath());
```
markstory: Would an application in a base directory have it `loginUrl` setup to have the base path?

Zuluru: I'm not clear on the question. Is your concern that `$uri->base` will never be non-empty here, or that the resulting URL will have the base duplicated?

markstory: My question is more on how loginUrl is generally used for applications in subdirectories. Is it configured to be `/base/login` or `/login`?

src/Middleware/AuthenticationMiddleware.php
Zuluru: The uri we deal with here is the one parsed by the request, which I believe will always have the base set, though it may be null. Unless we expect this may be used with something other than Cake's ServerRequestFactory?

Zuluru:
There are also two places in `Cake\Http\Cookie\CookieCollection` (`addFromResponse` and `addToRequest`) that use `$uri->getPath()`. I don't really use cookies outside of authentication, so it's not clear to me whether those should also be updated in a similar manner.

src/UrlChecker/DefaultUrlChecker.php
Zuluru: Just realized that your concern may be that if we make this URL include the base, but the configured one didn't have that as part of it, then they won't match. I guess my response would be that, since this already isn't working in subfolders, there shouldn't really be any code out there relying on any particular behaviour. And since the URL checker does strict comparisons without any type conversion, it has to be given string URLs in the configuration, and the sensible way of generating those strings is with `Router::url()`, which will include the base.

In short, including the base here shouldn't break existing code, and works as expected with `Router::url()`.

src/Middleware/AuthenticationMiddleware.php
Zuluru: Not sure if you're waiting on me to update the PR based on this...

markstory: We tried to make this middleware usable in non-cakephp projects. Doing so requires sticking to the PSR7 interface methods and attributes and only conditionally using our extensions if they are available.

tests/bootstrap.php
Zuluru: My git-fu isn't strong enough to sort out how to resolve this conflict. :-(

Zuluru: Should have some time to get back to this on the weekend, now that a big deadline on another project is past. Anyone have suggestions on what git commands I need to run to merge the other commits into my own fork, resolve the conflict, and push back into my PR?

garas: Run `git merge upstream/master` (assuming this repo is named `upstream`). It will say `bootstrap.php` has conflict.

Open `bootstrap.php`, near end it will have
```
<<<<<<< master
Plugin::load('Auth', [
    'path' => dirname(dirname(__FILE__)) . DS,
]);

$_SERVER['PHP_SELF'] = '/';
=======
Plugin::getCollection()->add(new \Authentication\Plugin());
>>>>>>> master
```

Resolve conflict as 
```
Plugin::getCollection()->add(new \Authentication\Plugin());

$_SERVER['PHP_SELF'] = '/';
```

Now add updated file, commit and push to your fork as you always do (ex. `git add bootstrap.php`, `git commit`, `git push`, or use GUI tools)

Zuluru: Awesome, thanks! I'm sure I tried something close to that, but obviously didn't hit the right combo. I mostly work in a "team" of one, so my git usage tends to be pretty straight-line.

garas: You've resolved conflict incorrectly...

In #233, the `Plugin::load(...);` was replaced with `Plugin::getCollection()...`. However you left both of them.

Your PR adds `$_SERVER['PHP_SELF'] = '/';` after that line.

Remove `Plugin::load(...);`. After that, _Files changed_ tab above should only show the line you've added. 

Also there is no new line now after `$_SERVER['PHP_SELF'] = '/';`.

Zuluru: Yep, that was a foolish oversight, skipped through too quickly specifically because it was such a small change. :-(

PR #234, opened 2019-01-29
Enable php73 check

dereuromark:
and use reasonable test matrix.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,7 @@ language: php
 php:
   - 5.6
   - 7.0
-  - 7.1
-  - 7.2
+  - 7.3
 
 sudo: false
 
@@ -19,7 +18,7 @@ env:
 matrix:
   fast_finish: true
   include:
-    - php: 7.1
+    - php: 7.2
       env: PHPCS=1 DEFAULT=0
     - php: 7.1
       env: PHPSTAN=1 DEFAULT=0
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -3,8 +3,7 @@ language: php
 php:
   - 5.6
   - 7.0
-  - 7.1
-  - 7.2
```
markstory: I would prefer to keep 7.2 over 7.0 which is soon to be eol.

dereuromark: why? we need the upper and lower limits of the language per major.
rest in between should be fine based on this.

markstory: Ok. The way typehints works between 7.0, 7.1 and 7.2 is different enough that we should keep 7.1 and 7.2 as there are changes that could be made where code is valid in 7.2 but broken in 7.1.

dereuromark: But it would be the same for 7.3 if it changes after another minor. The changes are always linear here, not fluctuating.
Can you show be one example that would not be true for the case as I outline it?

markstory: PHP 7.2+ allows parameter widening, while 7.1 does not. https://3v4l.org/oISbn

```php
class Fruit {
    public function eat(Mammal $h) {
    }
}

class Apple extends Fruit {
    public function eat($h) {   
    }
}

class Mammal {
}
```

dereuromark: Yeah, but why wouldnt that be covered automatically by 7.3 then? Thats my point.

dereuromark: The min+max always works here for our use cases: https://github.com/dereuromark/cakephp-ide-helper/blob/master/.travis.yml#L6-L7
It can be sprinkled with coverage as `7.0` or alike to have also a min for the major - but thats all optional.

inoas: > PHP 7.2+ allows parameter widening, while 7.1 does not. https://3v4l.org/oISbn
> 
> ```
> class Fruit {
>     public function eat(Mammal $h) {
>     }
> }
> 
> class Apple extends Fruit {
>     public function eat($h) {   
>     }
> }
> 
> class Mammal {
> }
> ```

But so does 7.1 - doesn't it?

inoas: What will CakePHP target anyway? 7.2, 7.3?

dereuromark: all declared in composer json file.

markstory: > Yeah, but why wouldnt that be covered automatically by 7.3 then? Thats my point. 

Yes 7.2 and 7.3 behave the same here.  I guess 7.0 and 7.1 also behave the same. It seems strange to me to pick 7.0 over 7.1 though. But I guess it won't matter much until it does :smile:

PR #235, opened 2019-02-02
2.x and cake 4.x

saeideng:


src/Authenticator/AuthenticatorCollection.php
```diff
@@ -83,7 +83,7 @@ protected function _resolveClassName($class)
      * @return void
      * @throws \RuntimeException
      */
-    protected function _throwMissingClassError($class, $plugin)
+    protected function _throwMissingClassError($class, $plugin): void
```
ADmad: No typehint for arguments?

saeideng: for typehints I will open another PR
just `:void` for return type was required by core here :)

saeideng: looks I have to do it in this PR , php 7.1 needs it

saeideng:
I pushed these commits into `2.x` after merge

PR #240, opened 2019-03-15
Add documentation on how to upgrade a hashed password.

### LINKED ISSUE #239: Changing Hashing Algorithms
Previously when using the AuthComponent we were able to see if the password needed a rehash by calling $this->Auth->authenticationProvider()->needPasswordRehash(). https://book.cakephp.org/3.0/en/controllers/components/authentication.html#changing-hashing-algorithms 

Currently with the new middleware it seems this is impossible to determine if the password needs rehashing

markstory:
Update both the identifiers docs and migration docs.

Fixes #239

### DIFF
--- a/docs/Identifiers.md
+++ b/docs/Identifiers.md
@@ -77,6 +77,55 @@ Configuration options:
 
 * **callback**: Default is `null` and will cause an exception. You're required to pass a valid callback to this option to use the authenticator.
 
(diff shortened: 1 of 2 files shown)


markstory:
cc @challgren

challgren:
Ty ahh I was missing the `->get('Password')` part!

PR #241, opened 2019-03-15
Fix for correct ordering of hash and query params

f-israel:
Just as the title says.

I had an application which needed a hash parameter to show the login page. In addition with the 'redirect' query parameter the order was generated wrongly (URL#HASH?QUERY). 

According to [RFC 3986 Section 4.2](https://tools.ietf.org/html/rfc3986#section-4.2) 
> relative-ref  = relative-part [ "?" query ] [ "#" fragment ]

the order should be reversed. This PR does simply the following: if generated url contains a hash it gets split and joined reordered.

### DIFF
--- a/src/Middleware/AuthenticationMiddleware.php
+++ b/src/Middleware/AuthenticationMiddleware.php
@@ -160,7 +160,7 @@ class AuthenticationMiddleware
 
         $hashPos = strpos($target, '#');
         if ($hashPos !== false) {
-            $query = $query . substr($target, $hashPos);
+            $query .= substr($target, $hashPos);
             $target = substr($target, 0, $hashPos);
         }
 
--- a/src/Middleware/AuthenticationMiddleware.php
+++ b/src/Middleware/AuthenticationMiddleware.php
@@ -158,11 +158,11 @@ class AuthenticationMiddleware
             $query = '?' . $query;
         }
 
-		if( ($hashPos = strpos($target, '#')) !== false){
+        if (($hashPos = strpos($target, '#')) !== false) {
             $query = $query . substr($target, $hashPos);
             $target = substr($target, 0, $hashPos);
         }
-		
+        
         return $target . $query;
     }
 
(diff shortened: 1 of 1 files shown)


src/Middleware/AuthenticationMiddleware.php
```diff
@@ -158,6 +158,11 @@ protected function getRedirectUrl($target, ServerRequestInterface $request)
             $query = '?' . $query;
         }
 
+        if (($hashPos = strpos($target, '#')) !== false) {
```
dereuromark: No inline assignment please. A separate line is fine.

f-israel: Corrected

dereuromark:
Is this something we have to backport to AuthComponent (the former class to manage this), as well?

src/Middleware/AuthenticationMiddleware.php
```diff
@@ -158,6 +158,12 @@ protected function getRedirectUrl($target, ServerRequestInterface $request)
             $query = '?' . $query;
         }
 
+        $hashPos = strpos($target, '#');
+        if ($hashPos !== false) {
+            $query = $query . substr($target, $hashPos);
```
dereuromark: This could be

    $query .= substr($target, $hashPos);

f-israel: Commited. Didn't shortened for readability. Guess could be shortened even further.

f-israel:
> 
> 
> Is this something we have to backport to AuthComponent (the former class to manage this), as well?

Didn't checked right now. Just crossed the bug and looked into fixing it. Will check into it

f-israel:
> 
> 
> > Is this something we have to backport to AuthComponent (the former class to manage this), as well?
> 
> Didn't checked right now. Just crossed the bug and looked into fixing it. Will check into it

Follow up: quickly checked the AuthComponent. As this uses controllers redirect and thus following the routers collection I guess there shouldn't be a problem.

dereuromark:
We need a test case for this to avoid regressions - ideally for both scenarios.

f-israel:
Sorry but that's where I've to skip. Never wrote (unit) tests.

markstory:
Rebased and merged in eaae8b074f10dbe594756d7258dc283ed3d0b0e9. Thanks @f-israel

PR #243, opened 2019-03-26
Update AuthenticationMiddleware to PSR-15 standard.

ADmad:


ADmad:
As a result of this change the identity will now be set to session *after* controller has run. This might actually be a good thing as it now prevents users from directly using session to get identity and instead always have to use the identity attribute of request.

src/Identifier/Ldap/ExtensionAdapter.php
```diff
@@ -93,7 +93,11 @@ public function getConnection()
...
         $this->_setErrorHandler();
-        $this->_connection = ldap_connect($host, $port);
+        $resource = ldap_connect($host, $port);
+        if ($resource === false) {
+            throw new RuntimeException('Unable to connect to LDAP server.');
+        }
+        $this->_connection = $resource;
         $this->_unsetErrorHandler();
 
         if (is_array($options)) {
```
dereuromark: Do we need this check? Shouldnt we typehint or throw meaningful exception instead of ignoring non array $options ? :)

ADmad: Adding / fixing type hints in beyond the scope of this PR.

src/Middleware/AuthenticationMiddleware.php
```diff
@@ -118,21 +123,25 @@ public function __invoke(ServerRequestInterface $request, ResponseInterface $res
...
 
-        $response = $result['response'];
-
         try {
-            return $next($request, $response);
+            $response = $handler->handle($request);
+            $authenticator = $service->getAuthenticationProvider();
+
+            if ($authenticator !== null && !$authenticator instanceof StatelessInterface) {
+                $return = $service->persistIdentity($request, $response, $result['result']->getData());
```
markstory: This will trip people up. The session will still have the user data in it, but it will always be from the previous request and not the current one.

ADmad: It will only trip those who incorrectly use session directly to get identity instead of the identity attribute of request instance.

The alternative is to have two separate methods for persisting identity. One that is called with request object before calling `$handler->handler()` and one that's called after with response object.

markstory: Him adding more methods does increase complexity. Perhaps we can rely on docs showing how to get the identity correctly and worry about this if that doesn't work. :)

ADmad: I have updated the "quick start" page warning against directly using session.

PR #244, opened 2019-03-29
Make AuthenticationInterface::authenticate() only return ResultInterface instance.

ADmad:
The request instance is no longer modified by authenticate() by calling persistIdentity()
internally. So returning back the request instance is not necessary.


ADmad:
Returning arrays with specific keys is fugly and best avoided.

PR #248, opened 2019-04-14
Move docs to sphinx and new plugin docs system

markstory:
The deployment tooling isn't complete yet, so I won't merge this until
those changes are done.


Dockerfile
```diff
@@ -0,0 +1,24 @@
+# Basic docker based environment
+# Necessary to trick dokku into building the documentation
+# using dockerfile instead of herokuish
+FROM ubuntu:17.04
```
ADmad: Any particular reason for not using 18.04 LTS ?

markstory: Because I'm lazy and was copy/pasting like a boss... :blush:

ADmad: The world thrive on copy/pasting :smile:

Issue #249, opened 2019-05-12
Make get() public again


dereuromark:
https://github.com/cakephp/authentication/blob/master/src/Identity.php#L104

Partially revert https://github.com/cakephp/authentication/pull/173 as get() is and should stay the primary access method if the content of identity is that fluent and the keys unknown from a static analyzer point of view.
This will only create problems from here on for both tooling, and IDE discovery.

For entities it is OK since [IdeHelper](https://github.com/dereuromark/cakephp-ide-helper) can make sure it at least stays in sync with the keys possible.
But here we have no such thing and as such the magic access is a code smell, and get() should be used primarily IMO.

With get() we still dont have type safety as return type hinting is still hard, but with IDEs like PHPStorm you can create a IdeHelper task to have string magic keys annotated to return types, that can mitigate the remaining issues here.
With the magic access basically nothing is possible here.

dereuromark:
PR open

PR #250, opened 2019-05-12
Fix up get() method and docs.

dereuromark:
Resolves https://github.com/cakephp/authentication/issues/249

get() can be made typehinted dynamically via IDE extension, the others cannot.
See [meta](https://github.com/dereuromark/cakephp-ide-helper/blob/master/tests/test_files/meta/phpstorm/.meta.php) file.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -31,8 +31,6 @@
         "cakephp/utility": "Provides CakePHP security methods. Required for the JWT adapter and Legacy password hasher."
     },
     "license": "MIT",
-    "minimum-stability": "dev",
-    "prefer-stable": true,
     "autoload": {
         "psr-4": {
             "Authentication\\": "src/"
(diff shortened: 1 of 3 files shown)


dereuromark:
We can backport it, but technically it is a fatal error if someone extended that method and visibilitiy changes.

markstory:
True. But part of our BC guide lines allow us to promote a protected method to public in a minor release.

dereuromark:
Lets wait till 1.x is merged into 2.x then we see what is still left of this.

dereuromark:
So whats the status on this?

ADmad:
Skimming through the diff I think most of the changes are already done. It would be best to rebase this branch on top of 2.x to get a clear picture.

dereuromark:
Looks like it can indeed be closed now

PR #251, opened 2019-05-13
2.x - Update tests and formatting.

markstory:
* Upgrade to PHPUnit 8.0
* Apply latest codesniffer rules.

### DIFF
--- a/src/AuthenticationService.php
+++ b/src/AuthenticationService.php
@@ -160,7 +160,7 @@ class AuthenticationService implements AuthenticationServiceInterface
     }
 
     /**
-     * {@inheritdoc}
+     * {@inheritDoc}
      *
      * @throws \RuntimeException Throws a runtime exception when no authenticators are loaded.
      */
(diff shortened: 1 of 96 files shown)


ADmad:
I took the liberty of updating typehints too.

src/AuthenticationService.php
```diff
@@ -151,17 +155,17 @@ public function loadAuthenticator($name, array $config = [])
...
      */
-    public function loadIdentifier($name, array $config = [])
+    public function loadIdentifier(string $name, array $config = []): IdentifierInterface
     {
         return $this->identifiers()->load($name, $config);
     }
 
     /**
-     * {@inheritDoc}
+     * @inheritdoc
```
dereuromark: wrong one, thats not the partial overwrite

ADmad: Casualty of search - replace :)

dereuromark:
I fixed those also in my branch :)

PR #252, opened 2019-05-13
Make Identity::get() public.

dereuromark:
Backport https://github.com/cakephp/authentication/pull/250 as per request

Also fix up phpstan to be easier verified locally.
Backport fixes from 2.x to 1.x to make phpstan level 4 work.

### DIFF
--- a/docs/en/identity-object.rst
+++ b/docs/en/identity-object.rst
@@ -24,20 +24,20 @@ implementations/sources::
        ->getAttribute('identity')
        ->getIdentifier();
 
-The identity object provides ArrayAccess but as well a `get()` method to access data.
-It is strongly recommended to use the `get()` method over array access because the get method is aware of the field mapping.
+The identity object provides ArrayAccess but as well a ``get()`` method to
+access data.  It is strongly recommended to use the ``get()`` method over array
+access because the get method is aware of the field mapping::
 
-```php
-$identity->get('email');
-$identity->get('username');
-```
-The `get()` method can also be type-hinted via IDE meta file, e.g. through [IdeHelper](https://github.com/dereuromark/cakephp-ide-helper).
+    $identity->get('email');
+    $identity->get('username');
 
-If you want, you can use property access, however:
-```php
-$identity->email;
-$identity->username;
-```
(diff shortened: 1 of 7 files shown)


src/Identifier/Ldap/ExtensionAdapter.php
```diff
@@ -94,7 +94,11 @@ public function getConnection()
     public function connect($host, $port, $options)
     {
         $this->_setErrorHandler();
-        $this->_connection = ldap_connect($host, $port);
+        $resource = ldap_connect($host, $port);
+        if ($resource === false) {
+            throw new RuntimeException('Unable to connect to LDAP server.');
```
markstory: Is there a more verbose error we can get from the ldap extension? Right now it is to figure out why it failed.

dereuromark: This is a 1:1 backport from the fixes admad did on the 2.x branch.
I didnt want to do any specifics, only wanted to get phpstan to pass :)

ADmad: As per PHP docs `ldap_connect()` only does syntactic check of the params and returns `false` if that fails. It doesn't actually connect to the server.

So error message could be `Syntactic checking of `ldap_connect()` parameters failed`.

https://www.php.net/manual/en/function.ldap-connect.php

ADmad: Based on a example in manual the message could be `The LDAP-URI is not parseable`.

markstory:
I fixed the docs (which are now RST like the book)

docs/en/identity-object.rst
```diff
@@ -24,20 +24,20 @@ implementations/sources::
...
-$identity->get('username');
-```
-The `get()` method can also be type-hinted via IDE meta file, e.g. through [IdeHelper](https://github.com/dereuromark/cakephp-ide-helper).
-
-If you want, you can use property access, however:
-```php
-$identity->email;
-$identity->username;
-```
+The identity object provides ArrayAccess but as well a ``get()`` method to
```
dereuromark: This is markdown, there is no need for double quoting.

PR #257, opened 2019-06-03
Update identifiers.rst

julianpollmann:
If Option Item hashers is a array it needs either have 'className' key or to be a string.


saeideng:
Duplicate of #256

PR #266, opened 2019-07-12
Renew session when ID when persisting/clearing identity.

ADmad:
See https://github.com/OWASP/CheatSheetSeries/blob/master/cheatsheets/Session_Management_Cheat_Sheet.md#renew-the-session-id-after-any-privilege-level-change

A user reported to @chinpei215 that the new authen lib didn't renew session ID like the core `AuthComponent`.

### DIFF
--- a/src/Authenticator/SessionAuthenticator.php
+++ b/src/Authenticator/SessionAuthenticator.php
@@ -84,6 +84,7 @@ class SessionAuthenticator extends AbstractAuthenticator implements PersistenceI
     public function persistIdentity(ServerRequestInterface $request, ResponseInterface $response, $identity)
     {
         $sessionKey = $this->getConfig('sessionKey');
+        $request->getAttribute('session')->renew();
         $request->getAttribute('session')->write($sessionKey, $identity);
 
         return [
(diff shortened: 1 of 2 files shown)


src/Authenticator/SessionAuthenticator.php
```diff
@@ -84,6 +84,7 @@ public function authenticate(ServerRequestInterface $request, ResponseInterface
     public function persistIdentity(ServerRequestInterface $request, ResponseInterface $response, $identity)
     {
         $sessionKey = $this->getConfig('sessionKey');
+        $request->getAttribute('session')->renew();
```
markstory: This might get called on every request when this is merged up to 2.x. This block https://github.com/cakephp/authentication/blob/deacd5a03378e8284e873f17cd932cf6eae2876c/src/Middleware/AuthenticationMiddleware.php#L130-L133 runs each time the user has been authenticated which happens on every request with session authenticator.

ADmad: I guess the session authenticator will have to first check if a persisted identity already exists and renew session id only if no identity is currently exists in session or existing one is not the same as new one.

ADmad:
Forgot about this one. So good to merge?

PR #269, opened 2019-07-22
phpdoc with return type for method IdentifierCollection::get($name)

andrii-pukhalevych:
Enables autocomplete for following code:
$authService = $this->Authentication->getAuthenticationService();
$errors = $authService ->identifiers()->get('Password')->getErrors()

### DIFF
--- a/src/Identifier/IdentifierCollection.php
+++ b/src/Identifier/IdentifierCollection.php
@@ -18,6 +18,10 @@ use Authentication\AbstractCollection;
 use Cake\Core\App;
 use RuntimeException;
 
+/**
+ * Class IdentifierCollection
+ * @method \Authentication\Identifier\IdentifierInterface get($name)
+ */
 class IdentifierCollection extends AbstractCollection implements IdentifierInterface
 {
 
--- a/src/Identifier/IdentifierCollection.php
+++ b/src/Identifier/IdentifierCollection.php
@@ -20,7 +20,7 @@ use RuntimeException;
 
 /**
  * Class IdentifierCollection
- * @method \Authentication\Identifier\IdentifierInterface get($name)
+ * @method \Authentication\Identifier\IdentifierInterface|null get($name)
  */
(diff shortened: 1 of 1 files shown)


src/Identifier/IdentifierCollection.php
```diff
@@ -18,6 +18,10 @@
 use Cake\Core\App;
 use RuntimeException;
 
+/**
+ * Class IdentifierCollection
```
dereuromark: Class name should be removed IMO, just noise.

src/Identifier/IdentifierCollection.php
```diff
@@ -18,6 +18,10 @@
...
+ * @method \Authentication\Identifier\IdentifierInterface get($name)
```
dereuromark: Can we annotate the type of $name? string? I think so.

dereuromark:
Doesnt the IdentifierInterface have the get() method? Then this shouldnt be needed IMO.

src/Identifier/IdentifierCollection.php
andrii-pukhalevych: Agree. Please change. 
Added it as placeholder for future use because almost all cake classes have short description in first line.

src/Identifier/IdentifierCollection.php
andrii-pukhalevych: Agree. Please change.

> Doesnt the IdentifierInterface have the get() method? Then this shouldnt be needed IMO.
You are right, this method exists in AbstractCollection extends ObjectRegistry, but it exists with general type (object|null), so autocomplete for IdentifierCollection methods does not work

dereuromark: The |null part though is then still true?
I feel like we should have an internal method as wrapper that also throws exception in null case then.

andrii-pukhalevych: Yes, agree. Not sure if its possible to change it in submitted PR.


> The |null part though is then still true?

Correct, I forgot about that.

PR #270, opened 2019-07-22
Add php doc for method AuthenticatorCollection::get($name) and fix message for RuntimeException

andrii-pukhalevych:
RuntimeException message had invalid namespace '\Auth\Authentication', should be '\Authentication\Authenticator'.
Phpdoc enables autocomplete:
$authService = $this->Authentication->getAuthenticationService();
$authenticator = $authService->authenticators()->get('Form');
$result = $authenticator->authenticate($request, $response)

### DIFF
--- a/src/Authenticator/AuthenticatorCollection.php
+++ b/src/Authenticator/AuthenticatorCollection.php
@@ -21,7 +21,7 @@ use RuntimeException;
 
 /**
  * Class AuthenticatorCollection
- * @method \Authentication\Authenticator\AuthenticatorInterface get($name)
+ * @method \Authentication\Authenticator\AuthenticatorInterface|null get($name)
  */
 class AuthenticatorCollection extends AbstractCollection
 {
--- a/src/Authenticator/AuthenticatorCollection.php
+++ b/src/Authenticator/AuthenticatorCollection.php
@@ -20,7 +20,7 @@ use Cake\Core\App;
 use RuntimeException;
 
 /**
- * @method \Authentication\Authenticator\AuthenticatorInterface|null get($name)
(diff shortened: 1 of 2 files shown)


src/Authenticator/AuthenticatorCollection.php
```diff
@@ -19,6 +19,10 @@
 use Cake\Core\App;
 use RuntimeException;
 
+/**
+ * Class AuthenticatorCollection
```
dereuromark: Same noise

src/Authenticator/AuthenticatorCollection.php
```diff
@@ -55,8 +59,9 @@ protected function _create($className, $alias, $config)
         $authenticator = new $className($this->_identifiers, $config);
         if (!($authenticator instanceof AuthenticatorInterface)) {
             throw new RuntimeException(sprintf(
-                'Authenticator class `%s` must implement \Auth\Authentication\AuthenticatorInterface',
-                $className
+                'Authenticator class `%s` must implement `%s`.',
+                $className,
+                AuthenticatorInterface::class
```
dereuromark: :+1:

src/Authenticator/AuthenticatorCollection.php
```diff
@@ -19,6 +19,9 @@
 use Cake\Core\App;
 use RuntimeException;
 
+/**
+ * @method \Authentication\Authenticator\AuthenticatorInterface|null get(string $name)
```
ADmad: The `|null` doesn't make sense to me, `null` can't have a method `get()`.

dereuromark: This comes from the method itself.
As i said: an own wrapper that throws exception would be most clean.

PR #277, opened 2019-08-31
bind using email

waspinator:


src/Identifier/LdapIdentifier.php
```diff
@@ -197,6 +212,29 @@ protected function _bindUser($username, $password)
         return null;
     }
 
+    /**
+     * Format LDAP entry data into associative array
+     *
+     * @param array $entry The entry
+     * @return array
+     */
+    protected function _formatEntry($entry)
```
ADmad: Plural form `_formatEntries()` would be better since it works on multiple entries.

PR #278, opened 2019-09-02
changed ldap bind strategy to use filter

waspinator:


docs/en/identifiers.rst
```diff
@@ -81,8 +81,12 @@ requires the PHP LDAP extension.
    ``['username' => 'username', 'password' => 'password']``.
 -  **host**: The FQDN of your LDAP server.
 -  **port**: The port of your LDAP server. Defaults to ``389``.
--  **bindDN**: The Distinguished Name of the user to authenticate. Must
-   be a callable. Anonymous binds are not supported.
+-  **bindDN**: The Distinguished Name of the ldap lookup user.
```
markstory:
```suggestion
-  **bindDN**: The Distinguished Name of the LDAP lookup user.
```

docs/en/identifiers.rst
```diff
@@ -81,8 +81,12 @@ requires the PHP LDAP extension.
...
+   Anonymous binds are not supported.
+-  **bindPassword**: The password of the ldap lookup user.
```
markstory:
```suggestion
-  **bindPassword**: The password of the LDAP lookup user.
```

src/Identifier/Ldap/ExtensionAdapter.php
```diff
@@ -161,6 +161,23 @@ public function unbind()
...
+     *
+     * @param string $baseDN Base DN for the directory
+     * @param string $filter search filter
+     * @return array
+     */
+    public function search($baseDN, $filter)
+    {
+        $this->_setErrorHandler();
+        $result = ldap_search($this->_connection, $baseDN, $filter);
+        $entries = ldap_get_entries($this->_connection, $result);
```
markstory: If these lines result in an error being thrown you won't restore the error handler.

src/Identifier/LdapIdentifier.php
```diff
@@ -180,14 +197,25 @@ protected function _connectLdap()
...
-                ]);
+                for ($i = 0; $i < $entries['count']; $i++) {
+                    $userLdapBind = $this->_ldap->bind($entries[$i]['dn'], $password);
+
+                    if ($userLdapBind === true) {
+                        $this->_ldap->unbind();
+
+                        return new ArrayObject([
+                            $config['fields'][self::CREDENTIAL_USERNAME] => $username,
+                            'ldap_attributes' => $this->_formatAttributes($entries[$i])
```
markstory: Most array keys in are `camelBacked` in CakePHP and related libraries.

src/Identifier/LdapIdentifier.php
```diff
@@ -92,18 +97,30 @@ public function __construct(array $config = [])
...
+
+        if (!isset($this->_config['bindDN'])) {
+            throw new RuntimeException('Config `bindDN` is not set.');
+        }
+
+        if (!isset($this->_config['bindPassword'])) {
+            throw new RuntimeException('Config `bindPassword` is not set.');
+        }
+
+        if (!isset($this->_config['baseDN'])) {
```
markstory: Adding required configuration values would break existing usage of this identifier, could we do this in a way that doesn't require people to update their code unless they want to use the new features?

src/Identifier/LdapIdentifier.php
```diff
@@ -30,9 +30,14 @@
  * ```
  *  new LdapIdentifier([
  *       'host' => 'ldap.example.com',
- *       'bindDN' => function($username) {
- *           return $username; //transform into a rdn or dn
- *       },
```
markstory: Changing `bindDN` from a callable into a string looks like a breaking change. Is there a way we accommodate both the current configuration options and add in the new filtering as an optional layer?

PR #279, opened 2019-09-02
Ldap update

waspinator:


markstory:
There seems to be a large overlap between this and #278 which one would you like to proceed with?

waspinator:
I'll try to fix the issues from #278 and make a new request with the changes in #279.

PR #280, opened 2019-09-02
added filtering and local user update to ldap

waspinator:


src/Identifier/Ldap/ExtensionAdapter.php
```diff
@@ -161,6 +161,26 @@ public function unbind()
...
+     */
+    public function search($baseDN, $filter)
+    {
+        $this->_setErrorHandler();
+        $result = ldap_search($this->_connection, $baseDN, $filter);
+        $this->_unsetErrorHandler();
+
+        $this->_setErrorHandler();
+        $entries = ldap_get_entries($this->_connection, $result);
+        $this->_unsetErrorHandler();
```
markstory: Should the error be thrown in the unsetErrorHandler() method? Right now if any of the ldap_ functions trigger an error they default error handler is not restored.

An alternative would be to replace both methods with a wrapper function like:

```php
$result = $this->convertErrorToException(function() use ($baseDN, $filter) {
   return ldap_search($this->_connection, $baseDN, $filter);
});
```
This wrapper could ensure that the error handler is restored before throwing an error.

src/Identifier/LdapIdentifier.php
```diff
@@ -17,7 +17,11 @@
 use ArrayObject;
 use Authentication\Identifier\Ldap\AdapterInterface;
 use Authentication\Identifier\Ldap\ExtensionAdapter;
+use Authentication\Identifier\Resolver\ResolverAwareTrait;
+use Authentication\Identifier\Resolver\ResolverInterface;
 use Cake\Core\App;
+use Cake\ORM\TableRegistry;
```
markstory: Are we able to be sure that users will always endup in an ORM/Table?

src/Identifier/LdapIdentifier.php
```diff
@@ -92,18 +118,63 @@ public function __construct(array $config = [])
...
+                if (!is_callable($this->_config['updateLocalIdentity'])) {
+                    throw new InvalidArgumentException(sprintf(
+                        'The `updateLocalIdentity` config is not a callable. Got `%s` instead.',
+                        gettype($this->_config['updateLocalIdentity'])
+                    ));
+                }
+            }
+
+            if (isset($this->_config['createLocalIdentityIfMissing'])) {
+                if (!is_callable($this->_config['createLocalIdentityIfMissing'])) {
```
markstory: Should we check for a `Closure` instead of a callable? The callable type includes a string and array types that can be problematic to safely handle.

src/Identifier/LdapIdentifier.php
```diff
@@ -136,9 +207,47 @@ public function identify(array $data)
...
+                if (!isset($bindResult['ldapAttributes'][$matchingField]) || empty($bindResult['ldapAttributes'][$matchingField])) {
+                    return null;
+                }
+
+                if (is_string($bindResult['ldapAttributes'][$matchingField])) {
+                    $identity = $this->_findIdentity($bindResult['ldapAttributes'][$matchingField]);
+                } elseif (is_array($bindResult['ldapAttributes'][$matchingField])) {
+                    foreach ($bindResult['ldapAttributes'][$matchingField] as $attribute_value) {
+                        $identity = $this->_findIdentity($attribute_value);
+                        if (!is_null($identity)) {
```
markstory: We use `$identity !== null` instead of the `is_null` function.

src/Identifier/LdapIdentifier.php
```diff
@@ -136,9 +207,47 @@ public function identify(array $data)
...
+                            break;
+                        }
+                    }
+                }
+
+                $updateLocalIdentity = $this->getConfig('updateLocalIdentity');
+                $createLocalIdentityIfMissing = $this->getConfig('createLocalIdentityIfMissing');
+
+                if (isset($identity)) {
```
markstory: This could be `$identity !== null` as well.

src/Identifier/LdapIdentifier.php
```diff
@@ -136,9 +207,47 @@ public function identify(array $data)
...
+            $bindResult = $this->_bindUserUsingLookup($data[$fields[self::CREDENTIAL_USERNAME]], $data[$fields[self::CREDENTIAL_PASSWORD]]);
+
+            if ($bindResult) {
+                $matchingField = $this->getConfig('matchingField');
+                if (!isset($bindResult['ldapAttributes'][$matchingField]) || empty($bindResult['ldapAttributes'][$matchingField])) {
+                    return null;
+                }
+
+                if (is_string($bindResult['ldapAttributes'][$matchingField])) {
+                    $identity = $this->_findIdentity($bindResult['ldapAttributes'][$matchingField]);
```
markstory: You should probably initialize `$identity` to null above these conditions.

src/Identifier/LdapIdentifier.php
```diff
@@ -197,6 +306,68 @@ protected function _bindUser($username, $password)
...
+    {
+        $config = $this->getConfig();
+
+        try {
+            $ldapBind = $this->_ldap->bind($config['lookupBindDN'], $config['lookupBindPassword']);
+
+            if ($ldapBind === true) {
+                $entries = $this->_ldap->search($config['baseDN'], $config['filter']($username));
+
+                for ($i = 0; $i < $entries['count']; $i++) {
```
markstory: Are you not able to use `foreach` here?

src/Identifier/Ldap/AdapterInterface.php
```diff
@@ -35,6 +35,15 @@ public function bind($bind, $password);
...
     public function connect($host, $port, $options);
 
+    /**
+     * Search the LDAP directory
+     *
+     * @param string $baseDN Base DN for the directory
+     * @param string $filter search filter
+     * @return array
+     */
+    public function search($baseDN, $filter);
```
dereuromark: What does DN mean?
Usually, variables are called `$baseDn` or the long version.

src/Identifier/LdapIdentifier.php
```diff
@@ -136,9 +207,48 @@ public function identify(array $data)
...
+                $updateLocalIdentity = $this->getConfig('updateLocalIdentity');
+                $createLocalIdentityIfMissing = $this->getConfig('createLocalIdentityIfMissing');
+
+                if ($identity !== null) {
+                    if (is_callable($updateLocalIdentity)) {
+                        $identity = $updateLocalIdentity($identity, $bindResult['ldapAttributes']);
+                    }
+
+                    return $identity;
+                } else {
```
dereuromark: no else necessary after return.

dereuromark:
Wow, quite a lot of code here.
Some small unit tests for part of the system might be useful.

src/Identifier/Ldap/AdapterInterface.php
waspinator: DN is an initialism for Distinguished Name.

PR #286, opened 2019-09-13
Fix Security component & session authenticator

### LINKED ISSUE #284: 1.1.3 & SecurityComponent yield "Bad Request".
When I have time later I will provide steps to reproduce the error more specifically. For now, I will say that in Cake 3.7.9 and Authentication plugin 1.1.3, using default configuration in `SecurityComponent`, I get a "Bad Request" on all form posts.

Downgrading to the 1.1.2 version of the plugin resolves it.

markstory:
The session needs to be read before being renewed. Failing to do results in data from the previous request being lost when the id is rotated.

Fixes #284

### DIFF
--- a/src/Authenticator/SessionAuthenticator.php
+++ b/src/Authenticator/SessionAuthenticator.php
@@ -86,6 +86,7 @@ class SessionAuthenticator extends AbstractAuthenticator implements PersistenceI
         $sessionKey = $this->getConfig('sessionKey');
         $session = $request->getAttribute('session');
         if (!$session->check($sessionKey)) {
+            $session->read();
             $session->renew();
             $session->write($sessionKey, $identity);
         }
--- a/src/Authenticator/SessionAuthenticator.php
+++ b/src/Authenticator/SessionAuthenticator.php
@@ -86,7 +86,6 @@ class SessionAuthenticator extends AbstractAuthenticator implements PersistenceI
         $sessionKey = $this->getConfig('sessionKey');
         $session = $request->getAttribute('session');
         if (!$session->check($sessionKey)) {
-            $session->read();
             $session->renew();
             $session->write($sessionKey, $identity);
         }
--- a/src/AuthenticationService.php
+++ b/src/AuthenticationService.php
@@ -177,12 +177,6 @@ class AuthenticationService implements AuthenticationServiceInterface
         foreach ($this->authenticators() as $authenticator) {
             $result = $authenticator->authenticate($request, $response);
             if ($result->isValid()) {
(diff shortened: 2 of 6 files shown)


lilHermit:
I'm not 100% sure this fixes the issue. My form/session login works ok `1.1.2` however broken with
`1.1.2`, `1.1.3` and `1.1.3` with this patch applied

markstory:
@lilHermit Ok. Any suggestion on how to reproduce the problem you're having? The previously reported issue was that "my login page doesn't work". I had a sample application locally that had that problem and this patch fixed that problem. Could you provide more context, as I don't have much to work with right now.

lilHermit:
@markstory I'll try and upload a repo tomorrow. But it was a very simple Session/Form authentication with ORM identity resolver All loaded from Application (not routes).

lilHermit:
@markstory see here

https://github.com/lilHermit/auth-test.git

markstory:
Thanks. I'll try to reproduce with this repo :+1:

markstory:
@lilHermit Thanks for the sample application. It was helpful.

The crux of the problem comes from the authentication plugin calling `$session->renew()` before SecurityComponent validate's POST data resulting in the validation error. Prior to 1.1.3 we didn't renew (rotate) session ids on login which is a different problem. I have a potential fix, but I wasn't able to get all the tests passing tonight.

ADmad:
https://github.com/cakephp/cakephp/issues/13652 sounds related.

ictgtvt:
I updated my description here https://github.com/cakephp/cakephp/issues/13652. thx!!!

markstory:
I 've moved `persistIdentity` to after the wrapped middleware runs.
 
Persisting the identity before calling $next results in the session being rotated before SecurityComponent handles the request. By doing the  rotation after we avoid that problem.

Moving when the session is updated will affect applications as the session data will *not* be populated during the controller action. This could be a breaking change but I think it is required to keep the session rotation fixes in place. This change will also align the behavior between 1.x and 2.x better.

src/Authenticator/SessionAuthenticator.php
```diff
@@ -86,6 +86,7 @@ public function persistIdentity(ServerRequestInterface $request, ResponseInterfa
         $sessionKey = $this->getConfig('sessionKey');
         $session = $request->getAttribute('session');
         if (!$session->check($sessionKey)) {
+            $session->read();
```
ADmad: Since the middleware now persists identity after calling $next() (aka controller code has already run) is this still necessary?

markstory: It probably isn't. I can take it out.

PR #287, opened 2019-09-13
Move redirect configuration to the service

markstory:
Moving redirect configuration and identityAttribute configuration allows application code to be more DRY and centralize configuration. Doing so requires a few new public methods as I didn't feel like adding `getConfig()` to the interface was a good idea.

Another benefit of moving then `unauthorizedRedirect` option is that users can use `Router::url()` to generate the target. Right now URLs must be a string which makes applications in sub directories more painful.

These changes will mostly backwards compatible as long as the application isn't using a custom service implementation.

### DIFF
--- a/src/AuthenticationService.php
+++ b/src/AuthenticationService.php
@@ -344,7 +344,7 @@ class AuthenticationService implements AuthenticationServiceInterface
      * If the `queryParam` option is set a query parameter
      * will be appended with the denied URL path.
      *
-     * @param Psr\Http\Message\ServerRequestInterface $request The request
+     * @param \Psr\Http\Message\ServerRequestInterface $request The request
      * @return string|null
      */
     public function getUnauthenticatedRedirectUrl(ServerRequestInterface $request)
--- a/src/AuthenticationService.php
+++ b/src/AuthenticationService.php
@@ -79,6 +79,11 @@ class AuthenticationService implements AuthenticationServiceInterface
      *      ]
      *   ]);
      *   ```
+     * - `identityAttribute` - The request attribute to store the identity in.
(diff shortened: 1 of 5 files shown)


src/Middleware/AuthenticationMiddleware.php
```diff
@@ -189,6 +152,24 @@ protected function getAuthenticationService($request, $response)
...
         }
+        $forwardKeys = ['identityAttribute', 'unauthenticatedRedirect', 'queryParam'];
+        foreach ($forwardKeys as $key) {
+            $value = $this->getConfig($key);
+            if ($value) {
+                deprecationWarning(
+                    "The `{$key}` configuration key on AuthenticationMiddleware is deprecated. " .
+                    "Instead set the `{$key}` on your AuthenticationService instance."
+                );
+                if (method_exists($subject, 'setConfig')) {
```
robertpustulka: Maybe this should check for a concrete `AuthenticationService` implementation (`$subject instanceof AuthenticationService`) instead?

src/AuthenticationService.php
```diff
@@ -317,4 +334,48 @@ public function buildIdentity($identityData)
...
+        $target = $this->getConfig('unauthenticatedRedirect');
+        if ($target === null) {
+            return null;
+        }
+        if ($param === null) {
+            return $target;
+        }
+
+        $uri = $request->getUri();
+        if (property_exists($uri, 'base')) {
```
ADmad: This makes me wonder if we should have our own Uri class/interface with additional methods for setting/getting base.

markstory: Perhaps, we only use the base property in a few places. If it comes up more extending the `UriInterface` is a good idea.

PR #289, opened 2019-09-18
1.1.x Revert session rotate

markstory:
Revert the session rotation changes made in 1.1.3 and attempted fix made in 1.1.4. This re-opens the possibility for a session fixation but restores compatibility with SecurityComponent.

If merged this would become 1.1.5. The changes in #286 would become 1.2.0. Those changes don't break an API contracts and the minor change in behavior should be ok in a minor release where we can document what changed.

### DIFF
--- a/tests/TestCase/AuthenticationServiceTest.php
+++ b/tests/TestCase/AuthenticationServiceTest.php
@@ -12,7 +12,7 @@
  * @since         1.0.0
  * @license       https://opensource.org/licenses/mit-license.php MIT License
  */
-namespace Authentication\Test\TestCase;
+namespace Authentication\Test\TestCase\Authenticator;
 
 use ArrayObject;
 use Authentication\AuthenticationService;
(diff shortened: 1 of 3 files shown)


tests/TestCase/AuthenticationServiceTest.php
```diff
@@ -12,7 +12,7 @@
  * @since         1.0.0
  * @license       https://opensource.org/licenses/mit-license.php MIT License
  */
-namespace Authentication\Test\TestCase;
+namespace Authentication\Test\TestCase\Authenticator;
```
ADmad: This seems to be done by mistake. Existing namespace is already in accordance with PSR-4.

dereuromark: The other way around: The subfolder should be created, and this put into it.
There shouldnt be tests directly in TestCase IMO, especially not when the class testes is not in src/ but src/Authenticator/.

ADmad: This file `AuthenticationServiceTest.php` is for `AuthenticationService.php` which is directly under `src` and hence it's in correct location directly under `TestCase` folder.

Issue #283, opened 2019-09-04
Result include the matching Authenticator and Identifier


Erwane:
## Context
my users table have 2 fields `password` and `password_tmp`
I want to known if user authenticate with `password` or `password_tmp`

Application.php
```php
    $service->loadAuthenticator('Authentication.Form', [
            'fields' => ['username' => 'email', 'password' => 'password'],
            'loginUrl' => Router::url(['_name' => 'auth.signin']),
        ]);
    $service->loadIdentifier('Authentication.Password', [
            'fields' => ['username' => 'email', 'password' => 'password'],
            'passwordHasher' => 'Authentication.Default',
        ]);
   $service->loadIdentifier('PasswordTmp', [
            'className' => 'Authentication.Password',
            'fields' => ['username' => 'email', 'password' => 'password_tmp'],
            'passwordHasher' => 'Authentication.Default',
        ]);
```

This works great BUT if don't known which Identifier match.
It's possible to include `Identifier` object in `Result` ?

markstory:
>  It's possible to include Identifier object in Result ?

How would that work? Where would you get that information in from the result?

Erwane:
There is 2 things : 
in `IdentifierCollection::identify`, should set the successfulIdentifier when result exists
in `XyzAuthenticator::authenticate` create `Result` and set `authenticator` ($this) and `identifier` (from IdentifierCollection::getSuccessfulIdentifier`

Then you return a `Result` and can access the correct Authenticator and Identifier.

dereuromark:
https://github.com/cakephp/authentication/pull/298

PR #296, opened 2019-09-30
#283: Authenticator and Identifier in Result

Erwane:
attache Identifier and Authenticator to result.

This allow more control in Controller when there is multiple authenticators and should do actions when a specific authentication or identifier match.

Refs https://github.com/cakephp/authentication/issues/283

### DIFF
--- a/src/Authenticator/Result.php
+++ b/src/Authenticator/Result.php
@@ -141,6 +141,7 @@ class Result implements ResultInterface
 
     /**
      * set the identifier who match the correct identity or null
+     * @param null|Authentication\Identifier\IdentifierInterface $identifier The matching identifier
      * @return void
      */
     public function setIdentifier(IdentifierInterface $identifier = null)
@@ -159,6 +160,7 @@ class Result implements ResultInterface
 
     /**
      * set the authenticator who match the correct identity or null
+     * @param null|Authentication\Authenticator\AuthenticatorInterface $authenticator The matching authenticator
      * @return void
      */
     public function setAuthenticator(AuthenticatorInterface $authenticator = null)
--- a/src/Authenticator/Result.php
+++ b/src/Authenticator/Result.php
@@ -66,7 +66,7 @@ class Result implements ResultInterface
      * @param array $messages Messages.
      * @throws \InvalidArgumentException When invalid identity data is passed.
      */
-    public function __construct($data, $status, array $messages = [])
+    public function __construct($data, $status, array $messages = [], IdentifierInterface $identifier = null, AuthenticatorInterface $authenticator = null)
     {
         if ($status === self::SUCCESS && empty($data)) {
             throw new InvalidArgumentException('Identity data can not be empty with status success.');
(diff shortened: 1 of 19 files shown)


src/Authenticator/Result.php
```diff
@@ -141,6 +141,7 @@ public function getIdentifier()
 
     /**
      * set the identifier who match the correct identity or null
+     * @param null|Authentication\Identifier\IdentifierInterface $identifier The matching identifier
```
dereuromark: Please set |null to the end, fasely ones are not the primary types here.

src/Authenticator/ResultInterface.php
```diff
@@ -79,9 +81,10 @@ public function getIdentifier();
 
     /**
      * set the identifier who match the correct identity or null
+     * @param null|Authentication\Identifier\IdentifierInterface $identifier The matching identifier
```
dereuromark: Invalid FQCN, \ missing in front of the type

src/Authenticator/Result.php
```diff
@@ -45,15 +46,29 @@ class Result implements ResultInterface
      */
     protected $_errors = [];
 
+    /**
+     * successful identifier or null
```
dereuromark: There seems to be a newline missing

src/Authenticator/Result.php
```diff
@@ -113,4 +131,42 @@ public function getErrors()
     {
         return $this->_errors;
     }
+
+    /**
+     * return null or the identifier who match the correct identity
```
dereuromark: Same here.
Also please real sentences. Capital letter start, full stop at end.

dereuromark:
Looks good.

src/Authenticator/Result.php
Erwane: i don't understand why.

```
$ composer cs-check
> phpcs --colors -p src/ tests/
............................................................ 60 / 91 (66%)
...............................                              91 / 91 (100%)


Time: 2.68 secs; Memory: 12MB
```

dereuromark: Some of those are not properly sniffed yet, missing sniffer rules.
But you can see the other doc lines on how this is handled so far.

Erwane:
Why phpstan fail when my unit tests are good ?

dereuromark:
The interface doesnt have getSuccessfulIdentifier() does it?
You can add it there as annotation above, this way we keep BC here.

dereuromark:
Is the target here 1.next or master? 1.next looks super outdated.

Erwane:
i've took 1.next because master looks for cake4 no ?

dereuromark:
No, I think 2.x is for cake4.
```
/**
 * @method type xyz()
 */
```
can be used to please phpstan. If this is a good idea or not, I cannot answer at this point.
Maybe wait for @ADmad etc to weigh in here.

Erwane:
ok, close this PR please, i will make a new one to master.
thanks a lot.

dereuromark:
You dont have to close, you can easily just set a different target as well, if you would like to.

dereuromark:
Now https://github.com/cakephp/authentication/pull/298

PR #298, opened 2019-09-30
#283: Successful identifier set in IdentifierCollection

Erwane:
get the successful identification provider from AuthenticationService or IdentifierCollection

### DIFF
--- a/src/AuthenticationService.php
+++ b/src/AuthenticationService.php
@@ -255,13 +255,23 @@ class AuthenticationService implements AuthenticationServiceInterface
     }
 
     /**
-     * Gets the successful authenticator instance if one was successful after calling authenticate
+     * Gets the successful authenticator instance if one was successful after calling authenticate.
      *
      * @return \Authentication\Authenticator\AuthenticatorInterface|null
      */
     public function getAuthenticationProvider()
     {
         return $this->_successfulAuthenticator;
+    }
+
+    /**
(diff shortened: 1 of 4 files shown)


src/Authenticator/Result.php
```diff
@@ -113,4 +130,42 @@ public function getErrors()
     {
         return $this->_errors;
     }
+
+    /**
+     * Return null or the identifier who match the correct identity.
```
dereuromark: The newlines are still missing in all docblocks.

Erwane:
i've just saw this in AuthenticationService
```php
            $result = $authenticator->authenticate($request, $response);
            if ($result->isValid()) {
                $this->_successfulAuthenticator = $authenticator;
                $this->_result = $result;
```
I think my PR is useless ... 
i will take a break and check that.

Erwane:
revamp my code and based it on the `AuthenticationService::getAuthenticationProvider`

Now, identifier is accessible thrue `AuthenticationService::getIdentificationProvider`, a convenient method for `IdentifierCollection::getIdentificationProvider`

dereuromark:
Is anyone against this? Or can we merge?

dereuromark:
We should also add some docs for this.

markstory:
@dereuromark :+1: Done

Erwane:
Thanks :)

PR #301, opened 2019-10-07
Update for Migrating Unauthenticated Redirects

eather009:
There is a Deprecated error for the given Migration steps. This is not working for latest version of CakePHP. This is deprecated from latest version of CakePHP. Instead, I tried with Service like following, 

$service->setConfig([
            'queryParam' => 'redirect',
            'unauthenticatedRedirect' => Router::url(["_name"=>"login"])
        ]);

But, it's not working.

### DIFF
--- a/docs/en/migration-from-the-authcomponent.rst
+++ b/docs/en/migration-from-the-authcomponent.rst
@@ -196,6 +196,7 @@ in this plugin will raise an exception in this scenario. You can convert
 this exception into a redirect using the ``unauthenticatedRedirect``
 when configuring the AuthenticationMiddleware.
 
+
 You can also pass the current request target URI as a query parameter
 using the ``queryParam`` option::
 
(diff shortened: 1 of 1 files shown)


docs/en/migration-from-the-authcomponent.rst
```diff
@@ -234,6 +235,11 @@ Then in your controller's login method you can use the redirect query parameter:
             return $this->redirect($target);
         }
     }
+    
+.. deprecated:: 4.0.0
+    The `unauthenticatedRedirect` and `queryParam` configuration key on AuthenticationMiddleware is deprecated.
```
markstory:
```suggestion
    The ``unauthenticatedRedirect`` and ``queryParam`` configuration key on AuthenticationMiddleware is deprecated. 
```

RST uses double ` to do inline code spans.

eather009:
As the above mentioned code is also not working, so we cannot set the example. If I found the way to make this done, I will share code.

markstory:
Ok. I'll get the examples updated.

PR #307, opened 2019-10-30
Docblock update in AuthenticationMiddleware.php

groovenectar:
Added `BaseApplication` as possible `$subject` type based on this:

"`$subject` Authentication service or application instance."

And the docs: https://book.cakephp.org/authentication/1.1/en/#getting-started

`$authentication = new AuthenticationMiddleware($this);`


groovenectar:
This was a mistake, but the documentation may be off...

PR #308, opened 2019-10-31
2.x phpunit

ADmad:
There's still something wacky which prevent running tests for individual files.

```
$ vendor/bin/phpunit tests/TestCase/Controller/Component/AuthenticationComponentTest.php 
PHPUnit 8.4.2 by Sebastian Bergmann and contributors.

W                                                                   1 / 1 (100%)

Time: 32 ms, Memory: 6.00 MB

There was 1 warning:

1) Warning
No tests found in class "Authentication\Test\TestCase\AuthenticationTestCase".
```

One can only run the full testsuite using `vendor/bin/phpunit`.

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -1,53 +1,38 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <phpunit
-	colors="true"
-	processIsolation="false"
-	stopOnFailure="false"
-	bootstrap="./tests/bootstrap.php"
-	>
-	<php>
-		<ini name="memory_limit" value="-1"/>
-		<ini name="apc.enable_cli" value="1"/>
-	</php>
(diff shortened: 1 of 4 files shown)


phpunit.xml.dist
```diff
@@ -1,53 +1,38 @@
...
-	<!-- Add any additional test suites you want to run here -->
-	<testsuites>
-		<testsuite name="authentication">
-			<directory>tests/TestCase/</directory>
-		</testsuite>
-		<!-- Add plugin test suites here. -->
-	</testsuites>
+    <!-- Add any additional test suites you want to run here -->
+    <testsuites>
+        <testsuite name="Authentication">
```
dereuromark: I recommend always using lowercase, dashed, as we do for all repos and is easier to type in cli.

PR #313, opened 2019-11-06
OrmResolver: allow fetching associations via config key 'contain'

smarek:
Re-querying the user record with associations after successful authentication seemed ineffective to me, hence this little update to default OrmResolver, so it could fetch the associations directly when authenticating and then the resulting entity with contained associations would be available in (basic example) session Auth.

### DIFF
--- a/docs/en/identifiers.rst
+++ b/docs/en/identifiers.rst
@@ -169,6 +169,7 @@ Configuration options:
 -  **userModel**: The user model identities are located in. Default is
    ``Users``.
 -  **finder**: The finder to use with the model. Default is ``all``.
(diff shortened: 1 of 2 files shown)


docs/en/identifiers.rst
```diff
@@ -169,6 +169,7 @@ Configuration options:
 -  **userModel**: The user model identities are located in. Default is
    ``Users``.
 -  **finder**: The finder to use with the model. Default is ``all``.
+-  **contain**: array of contained associations. Default is ``[]``.
```
markstory: We don't include `contain` as you should use a `finder` instead.

PR #314, opened 2019-11-08
Fix phpstan min and some docblocks.

dereuromark:
also, unused property is now used.
or is it better to remove the attribute and not use any service cache here?

### DIFF
--- a/src/Controller/Component/AuthenticationComponent.php
+++ b/src/Controller/Component/AuthenticationComponent.php
@@ -54,14 +54,14 @@ class AuthenticationComponent extends Component implements EventDispatcherInterf
     /**
      * List of actions that don't require authentication.
      *
-     * @var array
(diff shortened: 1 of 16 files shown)


src/Controller/Component/AuthenticationComponent.php
```diff
@@ -143,7 +149,7 @@ public function startup()
 
         $identity = $request->getAttribute($this->getConfig('identityAttribute'));
         if (!$identity) {
-            throw new UnauthenticatedException();
+            throw new UnauthenticatedException('No identity found. You can skip this check using `requireIdentity` config set to  `false`.');
```
markstory:
```suggestion
            throw new UnauthenticatedException('No identity found. You can skip this check by configuring  `requireIdentity` to be `false`.');
```

PR #317, opened 2019-11-09
Fix errors reported by psalm.

ADmad:


src/AuthenticationService.php
```diff
@@ -129,7 +129,7 @@ public function identifiers(): IdentifierCollection
     /**
      * Access the authenticator collection
      *
-     * @return \Authentication\Authenticator\AuthenticatorCollection|\Authentication\Authenticator\AuthenticatorInterface[]
+     * @return \Authentication\Authenticator\AuthenticatorCollection
```
dereuromark: This must be kept for iteratiin to properly discover in IDE

src/AuthenticationService.php
```diff
@@ -179,7 +179,9 @@ public function authenticate(ServerRequestInterface $request): ResultInterface
             );
         }
 
+        /** @var \Authentication\Authenticator\ResultInterface $result */
```
dereuromark: These inline shouldnt be needed, also null missing here it seems.

src/AuthenticationService.php
ADmad: Claiming it's an array of `AuthenticatorInterface` instances is a lie. It then causes static analysers to rightly complain about use of collection methods like `$this->authenticators()->load()`.

Until your IDE comes up with a way to declare the type of individual item returned by an iterator you can use inline annotation above your `foreach` loop when iterating the result of this method as done on line `177`.

dereuromark: Interesting. We do that also all the time in core
That's actually the correct way IMO to declare the iterable part of any object, also other objects and other collections with the concrete one.
We should rather keep this one line than having the need to inline fix this everywhere in code.

src/AuthenticationService.php
```diff
@@ -173,13 +173,8 @@ public function loadIdentifier(string $name, array $config = []): IdentifierInte
...
     public function authenticate(ServerRequestInterface $request): ResultInterface
     {
-        if ($this->authenticators()->isEmpty()) {
-            throw new RuntimeException(
-                'No authenticators loaded. You need to load at least one authenticator.'
-            );
-        }
-
         $result = null;
+        /** @var \Authentication\Authenticator\AuthenticatorInterface $authenticator */
```
dereuromark: Tip: Always attach it to the plural with `type[]`

src/AuthenticationService.php
dereuromark: Note: Semantically, the array part here is gone once you have another object attached.
Same with `\ArrayObject|\Foo[]`. If this was an array it would just be the last part.
Maybe we dont do these kind of magic iterations anymore in the future to prevent these smells in the first place ;)

ADmad: > Maybe we dont do these kind of magic iterations anymore in the future to prevent these smells in the first place ;)

Nah just the static analysers / IDEs need to get smarter and come up with a common way to annotating individual item type returned by iterators.

Psalm already seem to provide a way of annotating classes with implement `IteratorAggrerate` https://psalm.dev/docs/annotating_code/templated_annotations/#template-constraints (which is what `ObjectRegistry` does).  I am going to look into it again and try to annotate `ObjectRegistry` properly.

dereuromark: Dunno, somethings code can be more readable and straightforward itself. Doesn't need all the foo magic :)
But either way, how can we now keep this in a way that it properly documents itself on the methods? Instead of forcing the user implementation code to inline annotate?
The way I fixed it seemed to do that good enough also for phpstan.

ADmad: The end user almost never needs to use this method or the collection directly, so let's just keep phpstan and pslam happy.

dereuromark: Hmm, it feels like a step backwords here also for understanding and especially for working with the code..
But I wont object to moving forward given the general improvements of the PR.

pslam-baseline.xml
```diff
@@ -0,0 +1,77 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
markstory: This filename looks wrong. It is currently *pslam* when it should probably be *psalm*.

src/AbstractCollection.php
```diff
@@ -59,14 +57,4 @@ public function isEmpty(): bool
     {
         return empty($this->_loaded);
     }
-
-    /**
-     * Returns iterator.
-     *
-     * @return \ArrayIterator
-     */
-    public function getIterator(): ArrayIterator
```
markstory: How does one iterate these collections now?

pslam-baseline.xml
ADmad: Yeah, I made a typo when running command to generate it, but the file is still used :)
I'll update it.

src/AbstractCollection.php
ADmad: This collection extends core's `ObjectRegistry` which already has it.

PR #326, opened 2019-11-25
fix a typo

saeideng:
I was reviewing the code and I found this and so I did not test on local :) 

Refs dd5ba91
according to above commit , it was `unauthenticatedRedirect` before but now is `unauthorizedRedirect` !!

@markstory can you confirm this? 

this is alternative to #325 that was mistaken opened on branch 2.x


markstory:
Good find. Thank you!

PR #327, opened 2019-11-25
Merge master into 2.x.

dereuromark:


src/Controller/Component/AuthenticationComponent.php
```diff
@@ -259,14 +261,20 @@ public function getIdentityData(string $path)
...
+     * is cleared and then set to ensure that privilege escalation
+     * and de-escalation include side effects like session rotation.
      *
      * @param \ArrayAccess $identity Identity data to persist.
      * @return $this
      */
     public function setIdentity(ArrayAccess $identity)
     {
         $controller = $this->getController();
+        $service = $this->getAuthenticationService();
```
ADmad: This variable isn't used further in the method.

PR #330, opened 2019-11-27
Show URL base in login URL mismatch error message.

ndm2:
With the base missing, error messages like `http:://localhost/users/login did not match /users/login` would be produced, which cloaks the actual issue of the missing base in the configured login URL.

Ideally the error message would only show the full URL in case the assigned URL checker actually tests against a full URL, but the URL checker interface doesn't really allow that.

### DIFF
--- a/src/Authenticator/FormAuthenticator.php
+++ b/src/Authenticator/FormAuthenticator.php
@@ -81,10 +81,15 @@ class FormAuthenticator extends AbstractAuthenticator
      */
     protected function _buildLoginUrlErrorResult($request)
     {
+        $uri = $request->getUri();
+        if (property_exists($uri, 'base')) {
+            $uri = $uri->withPath($uri->base . $uri->getPath());
+        }
+
         $errors = [
             sprintf(
                 'Login URL `%s` did not match `%s`.',
-                (string)$request->getUri(),
+                (string)$uri,
                 implode('` or `', (array)$this->getConfig('loginUrl'))
             ),
         ];
--- a/src/Authenticator/FormAuthenticator.php
+++ b/src/Authenticator/FormAuthenticator.php
@@ -82,8 +82,9 @@ class FormAuthenticator extends AbstractAuthenticator
     protected function _buildLoginUrlErrorResult($request)
     {
         $uri = $request->getUri();
-        if (property_exists($uri, 'base')) {
-            $uri = $uri->withPath($uri->base . $uri->getPath());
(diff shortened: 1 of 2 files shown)


tests/TestCase/Authenticator/FormAuthenticatorTest.php
```diff
@@ -236,6 +268,38 @@ public function testMultipleLoginUrlSuccess()
...
+        $response = new Response();
+
+        $form = new FormAuthenticator($identifiers, [
+            'loginUrl' => '/base/users/login',
+        ]);
+
+        $result = $form->authenticate($request, $response);
+
+        $this->assertInstanceOf(Result::class, $result);
+        $this->assertEquals(Result::SUCCESS, $result->getStatus());
```
dereuromark: assertSame() should ideally be used for scalar comparison to avoid false positive on casting.

ndm2: Lazy me just copied the tests :) But aren't both values supposed to be strings?

dereuromark: Could be, sure.

src/Authenticator/FormAuthenticator.php
```diff
@@ -81,10 +81,15 @@ protected function _getData(ServerRequestInterface $request)
...
+        $uri = $request->getUri();
+        if (property_exists($uri, 'base')) {
+            $uri = $uri->withPath($uri->base . $uri->getPath());
+        }
+
         $errors = [
             sprintf(
                 'Login URL `%s` did not match `%s`.',
-                (string)$request->getUri(),
+                (string)$uri,
```
ADmad: You can change this line to `(string)$request->getAttribute('webroot') . (string)$request->getUri()` and avoid all other code.

ndm2: The thing is that `(string)$request->getUri()` returns an absolute URL. So it would have to use the relative URL instead then, but that would require additional code to extract the host, the scheme, the port, etc.

ADmad: Well then just replace the use of `$uri->base` above with `$request->getAttribute('base')`. That would avoid the need for property existence check.

dereuromark:
So is this going now in master (4.x) or do we also backport to 3.x?

markstory:
> So is this going now in master (4.x) or do we also backport to 3.x?

I can backport to 3.x

PR #333, opened 2019-12-16
Use stable releases

ADmad:


composer.json
```diff
@@ -31,8 +31,6 @@
         "cakephp/utility": "Provides CakePHP security methods. Required for the JWT adapter and Legacy password hasher."
     },
     "license": "MIT",
-    "minimum-stability": "dev",
-    "prefer-stable": true,
```
dereuromark: we can probably keep prefer stable right?

ADmad: We can but it would be redundant without `"minimum-stability"` set to `dev` or `beta`

dereuromark: Jep, but we have one less to always add and remove for each major.
Meh/^^

dereuromark:
something is wrong with master. it is not the valid 2.x (cake4) code...
https://github.com/cakephp/authentication/compare/2.x?expand=1
thus the fails.

PR #336, opened 2019-12-17
Clarify hook name

ravage84:
Hook name used is ``getAuthenticationService``:
https://github.com/cakephp/authentication/blob/1.4.0/src/Middleware/AuthenticationMiddleware.php#L153

### DIFF
--- a/docs/en/index.rst
+++ b/docs/en/index.rst
@@ -27,7 +27,8 @@ Add the authentication to the middleware. See the CakePHP `documentation
 <http://book.cakephp.org/3.0/en/controllers/middleware.html>`_ on how to use
 middleware if you are not familiar with it.
 
-Example of configuring the authentication middleware using ``authentication`` application hook::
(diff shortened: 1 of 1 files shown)


docs/en/index.rst
```diff
@@ -27,7 +27,8 @@ Add the authentication to the middleware. See the CakePHP `documentation
 <http://book.cakephp.org/3.0/en/controllers/middleware.html>`_ on how to use
 middleware if you are not familiar with it.
 
-Example of configuring the authentication middleware using ``authentication`` application hook::
+Example of configuring the authentication middleware using the ``getAuthenticationService``
```
dereuromark: () missing if that is a method.

ravage84: Well, I agree, but I assumed we do not add them as ``authentication`` didn't have one, either.

othercorey: I prefer `()` in method names in docs.

PR #349, opened 2020-02-04
Prevent pass unset queryParam to getTokenFromQuery

LukeC8:
Following the Token Authenticator documentation, queryParam is optional.
Add an optional param to getTokenFromQuery to avoid the code crashes when
queryParam is unset.

From documentation:

> "...
> queryParam: Name of the query parameter. Configure it if you want to get the token from the query parameters.
> header: Name of the header. Configure it if you want to get the token from the header.
> "

### DIFF
--- a/src/Authenticator/TokenAuthenticator.php
+++ b/src/Authenticator/TokenAuthenticator.php
@@ -91,10 +91,10 @@ class TokenAuthenticator extends AbstractAuthenticator implements StatelessInter
      * Gets the token from the request headers
      *
      * @param \Psr\Http\Message\ServerRequestInterface $request The request that contains login information.
-     * @param string $queryParam Request query parameter name
+     * @param string|null $queryParam Request query parameter name
      * @return string|null
      */
-    protected function getTokenFromQuery(ServerRequestInterface $request, string $queryParam): ?string
+    protected function getTokenFromQuery(ServerRequestInterface $request, ?string $queryParam): ?string
     {
         $queryParams = $request->getQueryParams();
 


src/Authenticator/TokenAuthenticator.php
```diff
@@ -94,7 +94,7 @@ protected function getTokenFromHeader(ServerRequestInterface $request, ?string $
      * @param string $queryParam Request query parameter name
```
dereuromark: This must then also be fixed.

src/Authenticator/TokenAuthenticator.php
```diff
@@ -94,7 +94,7 @@ protected function getTokenFromHeader(ServerRequestInterface $request, ?string $
...
      * @return string|null
      */
-    protected function getTokenFromQuery(ServerRequestInterface $request, string $queryParam): ?string
+    protected function getTokenFromQuery(ServerRequestInterface $request, ?string $queryParam): ?string
```
dereuromark: `if (empty($queryParams[$queryParam])) {` then makes no sense anymore with null.

dereuromark:
Based on the code I dont think you are correct.
However, maybe the calling code should not execute this method if the queryParam is null.
This would fix it a bit more cleanly IMO.

markstory:
How could someone reproduce the crash you're seeing?

LukeC8:
Hi @dereuromark, @markstory 
The context of the fix was this issue:

- In my backend API, the queryParam and prefixToken are not defined, as in documentation these fields are optional.

```php
        $authService = new \Authentication\AuthenticationService([
            'unauthenticatedRedirect' => '/404',
            'queryParam' => 'redirect'
        ]);

        $authService->loadAuthenticator('Authentication.Token',[
            'header' => 'token',
        ]);

        $authService->loadIdentifier('Authentication.Token', [
            'tokenField' => 'token',
            'dataField' => 'token',
            'resolver' => [
                'userModel' => 'tokens',
                'finder' => 'all'
            ]
        ]);
```

- So a 3th-party front-end code submitted a request to the backend without the __token field__ in the header. The code crashes with:

> Argument 2 passed to Authentication\Authenticator\TokenAuthenticator::getTokenFromQuery() must be of the type string, null given, called in vendor/cakephp/authentication/src/Authenticator/TokenAuthenticator.php on line 48

dereuromark:
Replaced by https://github.com/cakephp/authentication/pull/351

PR #351, opened 2020-02-11
Handle missing param config

cleptric:
Replaces #349 
We can't assume that a user sets up a config value for using the queryParam.

### DIFF
--- a/src/Authenticator/TokenAuthenticator.php
+++ b/src/Authenticator/TokenAuthenticator.php
@@ -88,21 +88,22 @@ class TokenAuthenticator extends AbstractAuthenticator implements StatelessInter
     }
 
     /**
-     * Gets the token from the request headers
+     * Gets the token from the request query
      *
      * @param \Psr\Http\Message\ServerRequestInterface $request The request that contains login information.
      * @param string $queryParam Request query parameter name
      * @return string|null
      */
-    protected function getTokenFromQuery(ServerRequestInterface $request, string $queryParam): ?string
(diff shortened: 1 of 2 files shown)


src/Authenticator/TokenAuthenticator.php
```diff
@@ -88,21 +88,22 @@ protected function getTokenFromHeader(ServerRequestInterface $request, ?string $
...
      * @return string|null
      */
-    protected function getTokenFromQuery(ServerRequestInterface $request, string $queryParam): ?string
+    protected function getTokenFromQuery(ServerRequestInterface $request, ?string $queryParam): ?string
     {
-        $queryParams = $request->getQueryParams();
-
-        if (empty($queryParams[$queryParam])) {
-            return null;
+        if (!empty($queryParam)) {
```
ADmad: You can return early is `$queryParam` is empty and avoid the nested `if`s.

src/Authenticator/TokenAuthenticator.php
```diff
@@ -88,21 +88,21 @@ protected function getTokenFromHeader(ServerRequestInterface $request, ?string $
...
      * @param string $queryParam Request query parameter name
      * @return string|null
      */
-    protected function getTokenFromQuery(ServerRequestInterface $request, string $queryParam): ?string
+    protected function getTokenFromQuery(ServerRequestInterface $request, ?string $queryParam): ?string
     {
-        $queryParams = $request->getQueryParams();
-
-        if (empty($queryParams[$queryParam])) {
+        if (empty($queryParam)) {
```
dereuromark: This could also be `=== null` check I think.

PR #356, opened 2020-03-06
Feature - Legacy Password salt hash

KainGNX:
Hello, I usually work in an environment that has various other vanilla PHP code touching the database, and when recently upgrading to Cake 4.x I needed to check passwords with no salt as we have in our older version of the Cake app which rely on components.

I had written a class in our code base which extends the LegacyPasswordHasher to not use a salt string when needed. I figured I would submit it here as a small feature, unless I missed where this plugin has this already, or standard way of integrating it.

Hope it helps!

### DIFF
--- a/src/PasswordHasher/LegacyPasswordHasher.php
+++ b/src/PasswordHasher/LegacyPasswordHasher.php
@@ -36,7 +36,7 @@ class LegacyPasswordHasher extends AbstractPasswordHasher
      */
     protected $_defaultConfig = [
         'hashType' => null,
-        'salt' => null
+        'salt' => null,
     ];
 
     /**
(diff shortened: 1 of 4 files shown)


src/PasswordHasher/LegacyPasswordHasher.php
```diff
@@ -29,11 +29,14 @@ class LegacyPasswordHasher extends AbstractPasswordHasher
 {
     /**
      * Default config for this object.
+     * - `hashType` String identifier of the hash type to use on the password. (e.g 'sha256' or 'md5')
+     * - `salt` Boolean flag for salting the password in a hashes, and checks.
```
ADmad: If this is supposed to be boolean, why not set the default value to `true` instead of `null`?

KainGNX: Good point, I changed the default to true.

markstory:
Thank you 👍

KainGNX:
You're welcome!

PR #357, opened 2020-03-15
Wrong getAuthenticationService() declaration

WodewoodTchuckchuck:
Got a 500 error with `PHP Fatal error:  Declaration of App\Application::getAuthenticationService(Psr\Http\Message\ServerRequestInterface $request, Psr\Http\Message\ResponseInterface $response): Authentication\AuthenticationServiceInterface must be compatible with Authentication\AuthenticationServiceProviderInterface::getAuthenticationService(Psr\Http\Message\ServerRequestInterface $request): Authentication\AuthenticationServiceInterface` when following current documentation.


ndm2:
Also needs a `use` statement for that interface.

PR #359, opened 2020-03-17
Fix broken Identifiers link

lgraham1:


docs/en/authenticators.rst
```diff
@@ -20,7 +20,7 @@ Configuration options:
    ``Auth``
 -  **identify**: Set this key with a value of bool ``true`` to enable checking
    the session credentials against the identifiers. When ``true``, the configured
-   `Identifiers <./Identifers.md>`__ are used to identify the user using data
+   `Identifiers <./identifers.md>`__ are used to identify the user using data
```
dereuromark: shouldnt that be identifiers wth ie?

lgraham1: Yes, oops. Although the docs aren't even in markdown files, so even with that typo fixed it won't work. I'm not familiar with rst, but it looks like it should actually be ```:doc:`/identifiers` ``` (like the ```:doc:`/authentication-component` ``` here https://raw.githubusercontent.com/cakephp/authentication/master/docs/en/index.rst)

I'll open a new PR as this one has already been merged.

Issue #229, opened 2018-10-30
unauthenticatedRedirect skip app directory


kicaj:
Based on https://github.com/cakephp/authentication/blob/master/docs/Migration-from-the-AuthComponent.md#migrating-unauthenticated-redirects and setting `unauthenticatedRedirect` to:
```
$authentication = new AuthenticationMiddleware($this, [
    'unauthenticatedRedirect' => Router::url([
        'controller' => 'Users',
        'action' = 'login'
    ]),
]);
```
then redirect to `localhost/users/login` instead of `localhost/my_app/users/login`.

Using router array in other place works correctly.

hiryu85:
also redirect url in redirect query parameter skip app directory

markstory:
@hiryu85 The redirect query parameter missing the app directory seems like a separate but related issue. Could you make a new issue for that problem?

hiryu85:
@markstory sure.
issue #230

Zuluru:
Is the Authentication middleware supposed to be before or after the Routing middleware in the queue? I have it after, and something very like your code is working fine for me. Is it because I have my "new" call wrapped in a function instead of adding the object directly to the queue?

krugerpl:
Hi, Any progress with this issue?

markstory:
@krugerpl no. Would you like to help by sending a pull request?

krugerpl:
> @krugerpl no. Would you like to help by sending a pull request?

I will try.

kicaj:
@krugerpl How are the jobs going?

nifa-dev:
I think I'm having the same, or closely related issue.  I'm using the following code to generate a URL in my Application.php file for redirect of an Unauthenticated request:

```
Router::url(
            [
                'prefix' => false,
                'controller' => 'Users',
                'action' => 'login',
                '_full' => true
            ]);
```

My site is hosted locally at http://localhost/scoring.  I have a route defined as /login for /users/login.  

This Router properly sets the relative URL as /login with the above function, but drops the /scoring directory from the URL and returns http://localhost/login instead of http://localhost/scoring/login.  

Any additional thoughts?  I'd love to see a code example from @Zuluru of what worked in that case.

Zuluru:
My application is at https://github.com/Zuluru/Zuluru3

nifa-dev:
@Zuluru, Thank you.  I was able to use that logic and successfully solve my problem.

andrii-pukhalevych:
[...]

PR #368, opened 2020-04-20
Update index.rst

btiknas:
Hi,

changed the string URL to use the Router::url to be flexibel.

Change needs also to be followed up in the Blog Tutorial - Authentication as it is not working proper with string mentioned as URL directly. See logs https://github.com/cakephp/authentication/issues/229


docs/en/index.rst
```diff
@@ -46,7 +47,10 @@ Example of configuring the authentication middleware using ``authentication`` ap
         {
             $service = new AuthenticationService();
             $service->setConfig([
-                'unauthenticatedRedirect' => '/users/login',
+                'unauthenticatedRedirect' => Router::url([
```
dereuromark, line 50: IMO we shouldnt need the Router::url() call here in config, but lazy applied later on in code.

dereuromark: If the code doesnt handle arrays here yet, we should fix also the code.

othercorey:
Is there a fix we need to make here?

docs/en/index.rst
```diff
@@ -59,7 +63,10 @@ Example of configuring the authentication middleware using ``authentication`` ap
             $service->loadAuthenticator('Authentication.Session');
             $service->loadAuthenticator('Authentication.Form', [
                 'fields' => $fields,
-                'loginUrl' => '/users/login'
+                'loginUrl' => Router::url([
+                                    'controller' => 'Users',
+                                    'action' => 'login'
+                                    ])
```
ndm2, line 69:
```suggestion
                'loginUrl' => Router::url([
                    'prefix' => false,
                    'plugin' => null,
                    'controller' => 'Users',
                    'action' => 'login',
                ])
```

ndm2: Same would apply above, if using the router here at all, then it should also set the `prefix` and `plugin` keys in order to avoid missing route errors when navigating in prefix/plugin scopes.

I'm wondering if it's still necessary to "unset" `prefix` and `plugin` differently, ie `false` vs `null`.

docs/en/index.rst
markstory: This method would be called when the AuthenticationMiddleware is called which should be after the Routing middleware is called which means Router::url() will work.

docs/en/index.rst
markstory: > I'm wondering if it's still necessary to "unset" prefix and plugin differently, ie false vs null.

It is still important to set them to null or false.

docs/en/index.rst
RiteshParyali: Jusr array not working,

dereuromark: We should fix it so that array Syntax suffices and works.

Issue #353, opened 2020-02-24
JWT Auth examples for cake 4?


cnizzardini:
Is there any examples for implementing JWT auth in cakephp 4? The only full example is for Form logins here: https://book.cakephp.org/authentication/2/en/index.html

The only examples I could find related to JWT is for a cake 3 plugin by ADmad.

markstory:
Right now we only have the one tutorial. Adding more is something we'd like to do though.

cnizzardini:
Yeah, it puts userland developers in a bit of a pickle trying to implement this. I've tried going through the code and unit tests, but that's not always super easy.

cnizzardini:
Can you at least tell me if its at least possible to authenticate with an HTTP POST of a username (email in my case) and password, and then return JWT token representing the user? I am attempting to build an API in CakePHP 4 and would prefer this approach since my application client is a trusted application. Its not clear looking at the documentation (or code) if this is possible.

markstory:
> Can you at least tell me if its at least possible to authenticate with an HTTP POST of a username (email in my case) and password, and then return JWT token representing the user?

Yes, that all sounds like it would happen outside of the authentication plugin. The JwtAuthenticator validates and decodes the contents of the JWT in each request, and can either return the payload contents or use the contents to identify a user (generally by id).

cnizzardini:
So I got this working, sort of, in a very ugly way... I am not proud of this...

in my Plugin.php (my API is a plugin so I use this instead of Application.php)
```
    public function getAuthenticationService(ServerRequestInterface $request): AuthenticationServiceInterface
    {
        $service = new AuthenticationService();

        $fields = [
            'username' => 'email',
            'password' => 'password'
        ];
        $service->loadAuthenticator('Authentication.Jwt', [
            'returnPayload' => false,
            'queryParam' => 'token',
        ]);
        $service->loadAuthenticator('Authentication.Form', [
            'fields' => $fields,
        ]);
        $service->loadIdentifier('Authentication.JwtSubject',[
            'tokenField' => 'id',
            'dataField' => 'id'
        ]);
        $service->loadIdentifier('Authentication.Password', [
            'returnPayload' => false,
            'fields' => $fields,
        ]);

        return $service;
    }
```

Then I have this in a login controller:

Needing to reference JWT and build my own token seems super hacky...so I assume this is wrong, but it does work and returns a token that is parseable on jwt.io

```php
    public function index()
    {
        if (!$this->request->is('POST')) {
[...]

PR #376, opened 2020-05-13
Use tokenField from config if defined

GrandFelix:
fixed wrong used config fields in some places when using custom fields
use tokenField from config if defined else IdentifierInterface::CREDENTIAL_JWT_SUBJECT.

this fixes this https://github.com/cakephp/authentication/issues/353#issuecomment-590667892

### DIFF
--- a/src/Authenticator/JwtAuthenticator.php
+++ b/src/Authenticator/JwtAuthenticator.php
@@ -37,6 +37,7 @@ class JwtAuthenticator extends TokenAuthenticator
         'algorithms' => ['HS256'],
         'returnPayload' => true,
         'secretKey' => null,
+        'dataField' => IdentifierInterface::CREDENTIAL_JWT_SUBJECT
     ];
 
     /**
@@ -89,8 +90,8 @@ class JwtAuthenticator extends TokenAuthenticator
 
         $result = json_decode(json_encode($result), true);
 
-        $key = $this->getConfig('dataField', IdentifierInterface::CREDENTIAL_JWT_SUBJECT);
-        if (empty($result[$key])) {
+        $dataKey = $this->getConfig('dataField');
+        if (empty($result[$dataKey])) {
             return new Result(null, Result::FAILURE_CREDENTIALS_MISSING);
         }
 
@@ -101,7 +102,7 @@ class JwtAuthenticator extends TokenAuthenticator
         }
 
         $user = $this->_identifier->identify([
-            $key => $result[$key],
+            $dataKey => $result[$dataKey],
         ]);
 
         if (empty($user)) {
--- a/src/Authenticator/JwtAuthenticator.php
+++ b/src/Authenticator/JwtAuthenticator.php
@@ -37,7 +37,7 @@ class JwtAuthenticator extends TokenAuthenticator
         'algorithms' => ['HS256'],
         'returnPayload' => true,
         'secretKey' => null,
-        'dataField' => IdentifierInterface::CREDENTIAL_JWT_SUBJECT
+        'dataField' => IdentifierInterface::CREDENTIAL_JWT_SUBJECT,
     ];
 
     /**
--- a/src/Authenticator/JwtAuthenticator.php
+++ b/src/Authenticator/JwtAuthenticator.php
@@ -37,7 +37,7 @@ class JwtAuthenticator extends TokenAuthenticator
         'algorithms' => ['HS256'],
         'returnPayload' => true,
         'secretKey' => null,
-        'dataField' => IdentifierInterface::CREDENTIAL_JWT_SUBJECT,
+        'subjectKey' => IdentifierInterface::CREDENTIAL_JWT_SUBJECT,
     ];
 
     /**
@@ -90,8 +90,8 @@ class JwtAuthenticator extends TokenAuthenticator
 
         $result = json_decode(json_encode($result), true);
 
-        $dataKey = $this->getConfig('dataField');
-        if (empty($result[$dataKey])) {
+        $subjectKey = $this->getConfig('subjectKey');
+        if (empty($result[$subjectKey])) {
             return new Result(null, Result::FAILURE_CREDENTIALS_MISSING);
         }
 
@@ -102,7 +102,7 @@ class JwtAuthenticator extends TokenAuthenticator
         }
 
         $user = $this->_identifier->identify([
-            $dataKey => $result[$dataKey],
+            $subjectKey => $result[$subjectKey],
         ]);
 
         if (empty($user)) {
--- a/src/Authenticator/JwtAuthenticator.php
+++ b/src/Authenticator/JwtAuthenticator.php
@@ -89,7 +89,7 @@ class JwtAuthenticator extends TokenAuthenticator
 
         $result = json_decode(json_encode($result), true);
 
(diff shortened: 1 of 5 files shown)


src/Identifier/TokenIdentifier.php
```diff
@@ -41,13 +41,13 @@ class TokenIdentifier extends AbstractIdentifier
      */
     public function identify(array $data)
     {
-        $dataField = $this->getConfig('dataField');
+        $dataField = $this->getConfig('tokenField');
```
dereuromark, line 44: Should the variable be also updated?

GrandFelix: yes indeed. fixed.

src/Authenticator/JwtAuthenticator.php
```diff
@@ -89,7 +89,7 @@ public function authenticate(ServerRequestInterface $request): ResultInterface
 
         $result = json_decode(json_encode($result), true);
 
-        $key = IdentifierInterface::CREDENTIAL_JWT_SUBJECT;
+        $key = $this->getConfig('tokenField', IdentifierInterface::CREDENTIAL_JWT_SUBJECT);
```
ADmad, line 92: Why do we have implementation specific constants on the interface? That's ugly :disappointed:.

dereuromark: For 2 years now it seems: https://github.com/cakephp/authentication/commit/b24fc05f7f691205b8e84e24e9d9f0e112ad7746
But can be refactored then with cake5 or the next major in this lib.

ndm2, line 92: This ties into the question why the config options are being flipped in the identifier... Should this use `dataField` instead?

src/Identifier/TokenIdentifier.php
```diff
@@ -41,13 +41,13 @@ class TokenIdentifier extends AbstractIdentifier
...
-        $dataField = $this->getConfig('dataField');
-        if (!isset($data[$dataField])) {
+        $tokenField = $this->getConfig('tokenField');
+        if (!isset($data[$tokenField])) {
             return null;
         }
 
         $conditions = [
-            $this->getConfig('tokenField') => $data[$dataField],
+            $this->getConfig('dataField') => $data[$tokenField],
```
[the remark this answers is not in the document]
GrandFelix: you are right. I came out from this issue https://github.com/cakephp/authentication/issues/353#issuecomment-590667892 I think data and token name is confusing. I expected that data is db field and token for token. But its other way arround. And also when you use custom token key name, here https://github.com/cakephp/authentication/pull/376/files/90a2daea03f470ef1294d9ac423bd42870a17ece#diff-b7a10d2c99be4b6528582b9471a0fb86R92 was fixed to 'sub' key name and because of that custom naming did not work and I did some debug. But now I see that naming is vice vera as I thought. But this https://github.com/cakephp/authentication/blob/master/src/Authenticator/JwtAuthenticator.php#L92 is still the issue.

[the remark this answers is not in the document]
ADmad: `JwtAuthenticator` should use `JwtSubjectIdentifier`, you don't need to mess around with the `TokenIdentifier`.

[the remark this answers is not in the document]
ADmad: @GrandFelix Switching the use of config options would be a backwards compatibility break. Since the issue is about making use of `IdentifierInterface::CREDENTIAL_JWT_SUBJECT` only `JwtAuthenticator` needs changes.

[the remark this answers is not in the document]
GrandFelix: yes... will fix that. thanks.

ADmad:
Walking through the `JwtAuthenticator` I why lot of people seem to have problem using it and why they need they the use of `sub` key configurable.

The authenticator has `returnPayload` option set to `true` by default which means the decoded JWT claims are directly returned as identity.

The way the `JwtAuthenticator` of my plugin works by default is it uses the value of `sub` claim of JWT to find user record with that `id` and use that as user/identity record.

GrandFelix:
ok.. I have reverted almost everything except JwtAuthenticator $key definition.

src/Authenticator/JwtAuthenticator.php
```diff
@@ -89,7 +89,7 @@ public function authenticate(ServerRequestInterface $request): ResultInterface
 
         $result = json_decode(json_encode($result), true);
 
-        $key = IdentifierInterface::CREDENTIAL_JWT_SUBJECT;
+        $key = $this->getConfig('dataField', IdentifierInterface::CREDENTIAL_JWT_SUBJECT);
```
ADmad, line 92: You should add the new key in the default config array using the constant for it's default value. Also `dataKey` might be a better name.

GrandFelix: done

src/Authenticator/JwtAuthenticator.php
```diff
@@ -89,8 +90,8 @@ public function authenticate(ServerRequestInterface $request): ResultInterface
 
         $result = json_decode(json_encode($result), true);
 
-        $key = IdentifierInterface::CREDENTIAL_JWT_SUBJECT;
-        if (empty($result[$key])) {
+        $dataKey = $this->getConfig('dataField');
```
ADmad, line 93: I was suggesting `dataKey` for the option name, not just the variable name :slightly_smiling_face:.

GrandFelix: pfff, ok. but this would break backward compatibility? IF anyone have already defined this with the default value of sub, when he will change the value this will not work anymore.

ADmad: Don't really understand what backwards compatibility concern you have. We are adding a new, so whether it's named `dataField` or `dataKey` really doesn't matter.

That's said let's see if anyone else has other suggestions regarding the new key name.

GrandFelix: you are right. it looks its too much for today. I have looking at Authentication.JwtSubject config and have messed up. Yes, thats new config. The only concern that is is that all other components have dataField naming. If thats ok I can change to dataKey.

ADmad: > The only concern that is is that all other components have dataField naming. If thats ok I can change to dataKey.

While naming consistency is important it should be contextual too. Personally I haven't seem the term "field" used in context of JWT. In JWT parlance the `sub` key is a ["claim"](https://tools.ietf.org/html/draft-ietf-oauth-json-web-token-27#section-4.1.2). So perhaps something more verbose like `identityClaimKey` might be ever better.

GrandFelix: I agree. Maybe `subjectClaimKey`

markstory: `claimKey` or `subjectKey` are shorter names that are inline with other JWT jargon.

markstory: It would be good to add a test for this configuration option to help prevent regressions in the future.

markstory:
Looking good. Are you able to update the tests or do you need a hand with that?

ADmad:
Targeting the PR to next minor would help document this new option better.

PR #377, opened 2020-05-14
Add example on JWT token generation/validation

rchavik:


ADmad:
The index page doesn't seem like the right place for this. It should under the "JWT" section on `docs/en/authenticators.rst`.

rchavik:
It's under Quick start, right after standard login example. I think it's a good place.

I'll move it anyway

docs/en/index.rst
```diff
@@ -170,6 +170,51 @@ setter method::
         }
     }
 
+JWT Token Generation
+====================
+
+To use JWT authentication, we need to generate keys::
```
ADmad, line 176:
```suggestion
By default the `JwtAuthenticator` uses `HS256` symmetric key algorithm and uses
the value of `Cake\Utility\Security::salt()` as encryption key.

For enhanced security one can instead use the `RS256` asymmetric key algorithm.
You can generate the required keys for that as follows::
```

docs/en/index.rst
```diff
@@ -170,6 +170,51 @@ setter method::
...
+                'token' => JWT::encode($payload, $privateKey, 'RS256'),
+            ];
+        } else {
+            $this->response = $this->response->withStatus(401);
+        }
+        $this->set(compact('json'));
+        $this->viewBuilder()->setOption('serialize', 'json');
+    }
+    
+Note that this requires setting up your `Applicaion` class as follows::
```
ADmad, line 206:
```suggestion
Note that this requires setting up your `Application` class as follows::
```

rchavik:
Obsoleted by #378

PR #380, opened 2020-05-21
Correct callback identifier example

AnhNhan:
Callback identifier example is intended to be `CallbackIdentifier`


markstory:
Thank you 👍

PR #394, opened 2020-07-29
Add samesite attribute and change spelling of config-key

Tobse:
The samesite-attribute is missing in the configuration-array and the spelling of the config-key 'httpOnly' is deprecated in CookieInterface.

### DIFF
--- a/src/Authenticator/CookieAuthenticator.php
+++ b/src/Authenticator/CookieAuthenticator.php
@@ -54,7 +54,7 @@ class CookieAuthenticator extends AbstractAuthenticator implements PersistenceIn
             'domain' => '',
             'secure' => false,
             'httponly' => false,
-            'samesite' => null,
+            'samesite' => null
         ],
         'passwordHasher' => 'Authentication.Default',
     ];
@@ -226,7 +226,7 @@ class CookieAuthenticator extends AbstractAuthenticator implements PersistenceIn
             $data['domain'],
             $data['secure'],
             $data['httponly'],
-            $data['samesite'],
+            $data['samesite']
         );
 
         return $cookie;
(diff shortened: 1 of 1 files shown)


src/Authenticator/CookieAuthenticator.php
```diff
@@ -224,7 +225,8 @@ protected function _createCookie($value): CookieInterface
             $data['path'],
             $data['domain'],
             $data['secure'],
-            $data['httpOnly']
+            $data['httponly'],
+            $data['samesite'],
```
markstory, line 229: This is causing parse errors in PHP 7.2 builds.

```suggestion
            $data['samesite']
```

src/Authenticator/CookieAuthenticator.php
```diff
@@ -224,7 +225,8 @@ protected function _createCookie($value): CookieInterface
             $data['path'],
             $data['domain'],
             $data['secure'],
-            $data['httpOnly']
+            $data['httponly'],
+            $data['samesite']
```
ADmad, line 229: Using [Cookie::create()](https://api.cakephp.org/4.1/class-Cake.Http.Cookie.Cookie.html#create) instead of `new Cookie()` would avoid having to pass each option separately.

src/Authenticator/CookieAuthenticator.php
```diff
@@ -216,15 +211,14 @@ public function clearIdentity(ServerRequestInterface $request, ResponseInterface
     protected function _createCookie($value): CookieInterface
     {
         $data = $this->getConfig('cookie');
+        $name = $data['name'];
+        unset($data['name']);
+        $options = $data;
```
ADmad, line 216: This is an unnecessary assignment. You can directly use `$data` below.

Tobse: Wow. You are that quick. Thanks for your patience. I'm learning day-by-day.

Tobse:
Uhh. I have to fix backwards-compatibility-issues before trying another merge.
CookieAuthenticator uses 'httpOnly' and 'expire' instead of 'httponly' and 'expires' as used in the Cookie Object.

ADmad:
> CookieAuthenticator uses 'httpOnly' and 'expire' instead of 'httponly' and 'expires' as used in the Cookie Object.

I'll take care of it.

PR #396, opened 2020-07-31
Use Cookie::create() instead of constructor to build cookie.

ADmad:
The default options from the `cookie` config have been removed so that the defaults
set for Cake\Http\Cookie\Cookie are used.

### DIFF
--- a/docs/en/authenticators.rst
+++ b/docs/en/authenticators.rst
@@ -245,7 +245,7 @@ Configuration options:
    -  **samesite**: String/null
 
    The defaults for the various options besides ``cookie.name`` will be those
-   set for the ``Cake\Http\Cookie\Cookie`` class. See [Cookie::setDefaults()](https://api.cakephp.org/4.0/class-Cake.Http.Cookie.Cookie.html#setDefaults)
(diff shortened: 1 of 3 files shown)


docs/en/authenticators.rst
```diff
@@ -236,13 +236,17 @@ Configuration options:
...
+   -  **name**: Cookie name, defaults is ``CookieAuth``
+   -  **expires**: Expiration
+   -  **path**: Path
+   -  **domain**: Domain
+   -  **secure**: Bool
+   -  **httponly**: Bool
+   -  **samesite**: String/null
+
+   The defaults for the various options besides ``cookie.name`` will be those
+   set for the ``Cake\Http\Cookie\Cookie`` class. See [Cookie::setDefaults()](https://api.cakephp.org/4.0/class-Cake.Http.Cookie.Cookie.html#setDefaults)
```
markstory, line 248:
```suggestion
   set for the ``Cake\Http\Cookie\Cookie`` class. See `Cookie::setDefaults() <https://api.cakephp.org/4.0/class-Cake.Http.Cookie.Cookie.html#setDefaults>`_
```

PR #397, opened 2020-08-03
Add deprecation warning

Tobse:
Add deprecation warning when using configuration key 'httpOnly' and 'expire'.


ADmad:
Thanks but these changes already done `2.next` for the next minor and don't need to be done on `master` for next bugfix release.

Issue #401, opened 2020-09-01
Wrong function description in authentication/src/Authenticator/TokenAuthenticator.php


YveIce:
It looks like a copy 'n paste from FormAuthenticator in
authentication/src/Authenticator/TokenAuthenticator.php 
at:
```php
   /**
     * Authenticates the identity contained in a request. Will use the `config.userModel`, and `config.fields`
     * to find POST data that is used to find a matching record in the `config.userModel`. Will return false if
     * there is no post data, either username or password is missing, or if the scope conditions have not been met.
     *
     * @param \Psr\Http\Message\ServerRequestInterface $request The request that contains login information.
     * @return \Authentication\Authenticator\ResultInterface
     */
    public function authenticate(ServerRequestInterface $request): ResultInterface
    {
```
Maybe this would be a bit better
```php
   /**
     * Authenticates the identity by token contained in a request. 
     * Token could be passed as query using `config.queryParam` or as header param using `config.header`. Token
     * prefix will be stripped if `config.tokenPrefix`is set. Will return false if no token is provided or if the 
     * scope conditions have not been met.
     *
     * @param \Psr\Http\Message\ServerRequestInterface $request The request that contains login information.
     * @return \Authentication\Authenticator\ResultInterface
     */
    public function authenticate(ServerRequestInterface $request): ResultInterface
    {
```

I'm sorry that i did't do a PR on this, as my english is far away from first-grade : )

dereuromark:
Can you make a PR with your suggested changes?
That will also be easier to review.

PR #402, opened 2020-09-01
fix issue #401

YveIce:
Wrong function description in authentication/src/Authenticator/TokenAuthenticator.php

### DIFF
--- a/src/Authenticator/TokenAuthenticator.php
+++ b/src/Authenticator/TokenAuthenticator.php
@@ -106,9 +106,10 @@ class TokenAuthenticator extends AbstractAuthenticator implements StatelessInter
     }
 
     /**
-     * Authenticates the identity contained in a request. Will use the `config.userModel`, and `config.fields`
-     * to find POST data that is used to find a matching record in the `config.userModel`. Will return false if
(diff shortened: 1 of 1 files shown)


src/Authenticator/TokenAuthenticator.php
```diff
@@ -106,9 +106,10 @@ protected function getTokenFromQuery(ServerRequestInterface $request, ?string $q
     }
 
     /**
-     * Authenticates the identity contained in a request. Will use the `config.userModel`, and `config.fields`
-     * to find POST data that is used to find a matching record in the `config.userModel`. Will return false if
-     * there is no post data, either username or password is missing, or if the scope conditions have not been met.
+     * Authenticates the identity by token contained in a request.
+     * Token could be passed as query using `config.queryParam` or as header param using `config.header`. Token
+     * prefix will be stripped if `config.tokenPrefix`is set. Will return false if no token is provided or if the
```
markstory, line 111:
```suggestion
     * prefix will be stripped if `config.tokenPrefix` is set. Will return false if no token is provided or if the
```

markstory:
Thank you 👍

PR #406, opened 2020-10-03
Translated into Japanese. Right now it's just index, but I'll do the …

Mr-IjIj:
…others soon.


docs/ja/index.rst
```diff
@@ -0,0 +1,53 @@
+unauthenticated users are able to access it::
```
markstory, line 1: There seems to be some content missing here.

othercorey:
Let's wait for the full translation here. Copying english to other language files means we have unsupported english that might not be translated.

Mr-IjIj:
Thank you for checking, it looks like I was using the wrong branch to push. I'm very sorry for the inconvenience...
We sent you a different pull request.

othercorey:
Moved to https://github.com/cakephp/authentication/pull/408

PR #411, opened 2020-10-12
Remove unneeded check and update testcase.

ADmad:
PersistenceInterface::persistIdentity() doesn't accept IdentityInterface instance,
hence AuthenticationService::persistIdentity() cannot either.


othercorey:
Looks ok.

PR #419, opened 2020-11-10
Add github actions

othercorey:


.gitignore
```diff
@@ -1,8 +1,6 @@
 # User specific & automatically generated files #
 #################################################
 /composer.lock
-/phpunit.xml
```
dereuromark, line 4: Shouldnt we keep the phpunit one as dist?

othercorey: I think it's easier for people to just modify phpunit.xml instead of having a possibly conflicting file overriding.

dereuromark: I think the idea is to have the phpunit.xml as gitignored copy locally without the danger of accidentally committing your passwords or more into the PR.
Personally, I would like to see that kept here.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,208 @@
...
+        prefer-lowest: ['']
+        include:
+          - php-version: '7.2'
+            db-type: 'mysql'
+            prefer-lowest: 'prefer-lowest'
+
+    steps:
+    - name: Setup MySQL latest
+      if: matrix.db-type == 'mysql' && matrix.php-version != '7.2'
+      run: docker run --rm --name=mysqld -e MYSQL_ROOT_PASSWORD=root -e MYSQL_DATABASE=cakephp -p 3306:3306 -d mysql --default-authentication-plugin=mysql_native_password --disable-log-bin
```
ADmad, line 28: Fetching specific myself versions using docker is quite wasteful here. Just use whatever is provided by the distro.

othercorey: We would need to add the db setup back in since the service starts with whatever configuration ga provides.

.gitignore
othercorey: Sure.

phpunit.xml
```diff
@@ -0,0 +1,38 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<phpunit
+    colors="true"
+    processIsolation="false"
+    stopOnFailure="false"
+    bootstrap="./tests/bootstrap.php"
```
dereuromark, line 6:
```suggestion
    bootstrap="tests/bootstrap.php"
```

phpunit.xml
```diff
@@ -0,0 +1,38 @@
...
+            <directory>tests/TestCase/</directory>
+        </testsuite>
+        <!-- Add plugin test suites here. -->
+    </testsuites>
+
+    <!-- Setup a listener for fixtures -->
+    <listeners>
+        <listener class="\Cake\TestSuite\Fixture\FixtureInjector">
+            <arguments>
+                <object class="\Cake\TestSuite\Fixture\FixtureManager" />
```
dereuromark, line 24: These must be copied over from some old sources
The new ones don't have the prefix slash as this one is not needed for string representation:
```suggestion
                <object class="Cake\TestSuite\Fixture\FixtureManager"/>
```

othercorey: Removing + adding the file messed things up.

.github/workflows/ci.yml
othercorey: I reduced the configs down to a single db for each.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,200 @@
...
+          export CODECOVERAGE=1 && vendor/bin/phpunit --verbose --coverage-clover=coverage.xml
+        else
+          vendor/bin/phpunit
+        fi
+
+    - name: Submit code coverage
+      if: matrix.php-version == '7.4'
+      uses: codecov/codecov-action@v1
+
+  testsuite-windows:
```
markstory, line 94: Do we need windows builds? I don't think there is much in this plugin that depends on windows.

othercorey: Removed the windows build.

PR #426, opened 2020-11-23
Revert "Added, From where to import JWT"

RiteshParyali:
Reverts cakephp/authentication#425 
Please check https://book.cakephp.org/authentication/2/en/authenticators.html#jwt
Then revert. something went wrong in code or space


dereuromark:
Newline was missing
should be fixed with https://github.com/cakephp/authentication/blob/master/docs/en/authenticators.rst

PR #430, opened 2021-01-04
Exception message

ADmad:
Remove dev centric error message. Since `UnauthenticatedException` now extends `HttpException` this is the message that gets shown to end user which is not desirable. Also made the error message configurable.

### DIFF
--- a/src/Controller/Component/AuthenticationComponent.php
+++ b/src/Controller/Component/AuthenticationComponent.php
@@ -45,6 +45,7 @@ class AuthenticationComponent extends Component implements EventDispatcherInterf
      * - `requireIdentity` - By default AuthenticationComponent will require an
      *   identity to be present whenever it is active. You can set the option to
      *   false to disable that behavior. See allowUnauthenticated() as well.
+     * - `unauthenticatedMessage` - Error message to use when `UnauthenticatedException` is thrown.
      *
      * @var array
      */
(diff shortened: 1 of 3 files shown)


src/Controller/Component/AuthenticationComponent.php
```diff
@@ -172,9 +174,7 @@ protected function doIdentityCheck(): void
 
         $identity = $request->getAttribute($this->getConfig('identityAttribute'));
         if (!$identity) {
-            throw new UnauthenticatedException(
-                'No identity found. You can skip this check by configuring  `requireIdentity` to be `false`.'
-            );
+            throw new UnauthenticatedException($this->getConfig('unauthenticatedMessage', ''));
```
markstory, line 177: Should we have a default message? Something like 'Authentication is required'?

ADmad: The constructor already set's one: https://github.com/cakephp/authentication/blob/048648d2e1daf73eed789404663aefee6747359e/src/Authenticator/UnauthenticatedException.php#L40-L42

markstory: Good job past us.

othercorey:
Should we allow the exception message to be empty instead of forcing the default in the constructor?

markstory:
I don't think an empty message is useful behavior.

PR #436, opened 2021-02-10
Allow customizing identity attribute in helper.

ADmad:
Without this change one can't used the helper if the identity attribute is customized in the AuthenticationService.

### DIFF
--- a/src/View/Helper/IdentityHelper.php
+++ b/src/View/Helper/IdentityHelper.php
@@ -28,6 +28,17 @@ use RuntimeException;
 class IdentityHelper extends Helper
 {
     /**
+     * Configuration options
+     *
+     * - `identityAttribute` - The request attribute which holds the identity.
+     *
+     * @var array
+     */
+    protected $_defaultConfig = [
+        'identityAttribute' => 'identity',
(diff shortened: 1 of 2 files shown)


src/View/Helper/IdentityHelper.php
```diff
@@ -44,7 +55,7 @@ class IdentityHelper extends Helper
      */
     public function initialize(array $config): void
     {
-        $this->_identity = $this->getView()->getRequest()->getAttribute('identity');
+        $this->_identity = $this->_View->getRequest()->getAttribute($this->getConfig('identityAttribute'));
```
othercorey, line 58: Is the `$config` parameter supposed to override default?

ADmad: Yes.

othercorey: Why do we pass $config to initialize() in helpers?

ADmad: I don't really remember why. But it does allow you to differentiate between options that were explicitly passed vs the config merged with defaults.

PR #437, opened 2021-02-11
Update AuthenticationRequiredException to extend HttpException.

### LINKED ISSUE #435: AuthenticationRequiredException not an HttpException. Why?
Hello there,

I have the following problem:
the [AuthenticationRequiredException](https://github.com/cakephp/authentication/blob/master/src/Authenticator/AuthenticationRequiredException.php) is usefull because it enables to pass headers to the response [here](https://github.com/cakephp/authentication/blob/master/src/Middleware/AuthenticationMiddleware.php#L112). Great!

Thus, if I throw an AuthenticationRequiredException in my CustomAuthenticator, I can pass elements to the header of the response from my CustomAuthenticator. All that I need, perfect.

Now the AuthenticationRequiredException does not extend the HttpException, but a simple RuntimeException. As opposed to the [UnauthenticatedException](https://github.com/cakephp/authentication/blob/master/src/Authenticator/UnauthenticatedException.php). Strange. Is there a reason for that?

Not being an HttpException, it is not catched by my ErrorController, and this is really annoying.

My question thus: would it be possible to have the AuthenticationRequiredException extending the HttpException class? I can make a PR for that. Or is there a good reason to avoid that?

ADmad:
Closes #435

### DIFF
--- a/src/Authenticator/AuthenticationRequiredException.php
+++ b/src/Authenticator/AuthenticationRequiredException.php
@@ -16,7 +16,7 @@ declare(strict_types=1);
  */
 namespace Authentication\Authenticator;
 
-use RuntimeException;
+use Cake\Http\Exception\HttpException;
 
 /**
  * An exception for stateless authenticators when credentials are wrong/missing.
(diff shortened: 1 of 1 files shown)


othercorey:
I'm guessing we don't have a unit test in Authentication that checks http code returned by cakephp? Otherwise this should have failed with latest cakephp since it didn't use the exception code as http code.

Otherwise looks ok.

ADmad:
Yeah I am too lazy to add a test. I only created this PR so that the fix would make it into the next release :)

PR #440, opened 2021-03-15
Removed extra arguments when calling authenticator() in the test code

YuyaAbo:
Currently `AbstractAuthenticator::authenticate()` takes only `ServerRequestInterface` as an argument. There was a point in the test code where `ResponseInterface ` was passed as the second argument, so it has been fixed.

https://github.com/cakephp/authentication/blob/master/src/Authenticator/AbstractAuthenticator.php#L88


ADmad:
Nice cleanup, thanks.

PR #444, opened 2021-04-02
2.x - Improve login URL mismatch error message.

ndm2:
Only show the full URL in the login URL mismatch error when the URL checker configuration's `checkFullUrl` option is enabled.

Without breaking changes to the URL checker interface, this is the best I could come up with for now. This should make the message generally less ambiguous, and specifically improve a situation where a full URL is used for the `loginUrl` option, but `checkFullUrl` isn't enabled, the check would fail, but the error message would show two identical full URLs.

### DIFF
--- a/src/Authenticator/FormAuthenticator.php
+++ b/src/Authenticator/FormAuthenticator.php
@@ -89,10 +89,17 @@ class FormAuthenticator extends AbstractAuthenticator
             $uri = $uri->withPath((string)$base . $uri->getPath());
         }
 
+        $checkFullUrl = $this->getConfig('urlChecker.checkFullUrl', false);
+        if ($checkFullUrl) {
+            $uri = (string)$uri;
+        } else {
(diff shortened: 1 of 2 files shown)


src/Authenticator/FormAuthenticator.php
```diff
@@ -89,10 +90,17 @@ protected function _buildLoginUrlErrorResult(ServerRequestInterface $request): R
             $uri = $uri->withPath((string)$base . $uri->getPath());
         }
 
+        $checkFullUrl = Hash::get((array)$this->getConfig('urlChecker'), 'checkFullUrl', false);
```
othercorey, line 93: Why don't you use dot notation to access nested config? `urlChecker.checkFullUrl` ?

ndm2: Because I'm a muffin, I totally forgot this was a thing.

ADmad (approved):

Issue #448, opened 2021-04-22
AuthenticationComponent::setIdentity() wraps IdentityInterface instances


robertpustulka:
I'm afraid dd5e77e536b4327d00cc1b30965b69f127d56446 added a regression to `AuthenticationService::persistIdentity()`

`AuthenticationService::persistIdentity()` called with a `IdentityInterface` instance will always attach a wrapped Idenity instance to the request attribute.

This way the request is repopulated with a wrapped instance and a component picks up that wrapped identity.

Either revert this commit or change how the component updates the request identity attribute.

robertpustulka:
I'll post a PR with a failing test case and we'll decide how to fix this then

robertpustulka:
PR is open, we can discuss the issue further there.

PR #449, opened 2021-04-23
Test passing identity instance to persistIdentity() and setIdentity()

robertpustulka:
This adds missing tests for calling persistIdentity() and setIdentity() with identity instance.

Refs #448

PS. I've also removed deprecated method calls form tests.

### DIFF
--- a/tests/TestCase/AuthenticationServiceTest.php
+++ b/tests/TestCase/AuthenticationServiceTest.php
@@ -646,6 +646,29 @@ class AuthenticationServiceTest extends TestCase
         $this->assertSame($identity, $service->getIdentity());
     }
 
+    /**
+     * testGetIdentityNull
+     *
+     * @return void
+     */
+    public function testGetIdentityNull()
+    {
+        $request = new ServerRequest();
+
+        $result = new Result(null, Result::FAILURE_OTHER);
+
+        $authenticator = $this->createMock(AuthenticatorInterface::class);
+        $authenticator->method('authenticate')
+            ->willReturn($result);
+
+        $service = new AuthenticationService();
+        $service->authenticators()->set('Test', $authenticator);
+
+        $service->authenticate($request);
+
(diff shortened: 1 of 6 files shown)


tests/TestCase/AuthenticationServiceTest.php
```diff
@@ -424,6 +424,25 @@ public function testPersistIdentityArray()
...
+     */
+    public function testPersistIdentityInstance()
+    {
+        $request = new ServerRequest();
+        $response = new Response();
+        $identity = new Identity([]);
+
+        $service = new AuthenticationService();
+
+        $result = $service->persistIdentity($request, $response, $identity);
```
ADmad, line 441: Passing an `Identity` instance here would mean `SessionAuthenticate` would store that instance in the session. But when one uses form login the instance stored in session is an entity, not an `IdentityInterface` instance.

IMO it's better to avoid such discrepancy.

ADmad: That said a clean way to fix the regression would be make `buildIdentity()` simply return the argument if it's already an `IdentityInterface` instance. That's would also avoid the object type checks at multiple places.

robertpustulka: > The authenticators should always deal with the "raw" identity object and decorating it with IdentityInterface should be left to AuthenticationService.

The thing is that sometimes the "raw" identity object is an identity instance itself. If your User entity already implements `IdentityInterface` that is. And this test covers that case.

ADmad: > If your User entity already implements IdentityInterface that is.

Ah, I had overlooked that case.

robertpustulka: @ADmad done and done.

ADmad (approved):

Issue #462, opened 2021-06-30
Can I use JWT and Form/Session Authentication in same app?


Josecanalla:
I have the following code

```
//In Application.php
public function getAuthenticationService(ServerRequestInterface $request): AuthenticationServiceInterface
{
    $service = new AuthenticationService();

    // Define where users should be redirected to when they are not authenticated
    $service->setConfig([
        'unauthenticatedRedirect' => Router::url([
            'prefix' => false,
            'plugin' => null,
            'controller' => 'Users',
            'action' => 'login',
        ]),
        'queryParam' => 'redirect',
    ]);

    $fields = [
        IdentifierInterface::CREDENTIAL_USERNAME => 'nombre_usuario',
        IdentifierInterface::CREDENTIAL_PASSWORD => 'pass'
    ];
    
    // Load the authenticators.
    
    $service->loadAuthenticator('Authentication.Jwt', [
        'secretKey' => Security::getSalt(),
        'returnPayload' => false,
    ]);

    $service->loadAuthenticator('Authentication.Session');
    $service->loadAuthenticator('Authentication.Form', [
        'fields' => $fields
    ]);

    // Load identifiers
    $service->loadIdentifier('Authentication.JwtSubject');
    $service->loadIdentifier('Authentication.Password', [
        'returnPayload' => false,
        'fields' => $fields,
    ]);

    return $service;
}
```

```
//In routes.php
$routes->scope('/', function (RouteBuilder $builder) {
    $builder->applyMiddleware('csrf');
    $builder->connect('/', ['controller' => 'RecReclamos', 'action' => 'index']);
    $builder->connect('/pages/*', 'Pages::display');   
    $builder->fallbacks();
});

$routes->prefix('Api', function (RouteBuilder $builder) {
    $builder->setExtensions(['json']);
    $builder->fallbacks();
});
```

```
//In /Api/UsersController.php
public function login()
{
    $this->Authorization->skipAuthorization();
    $result = $this->Authentication->getResult();

    if ($result->isValid()) {
        $privateKey = Security::getSalt();
        $user = $result->getData();
        $payload = [
            'sub' => $user->id,
            'exp' => time() + 60,
        ];
        $json = [
            'token' => JWT::encode($payload, $privateKey),
        ];
    } else {
        $this->response = $this->response->withStatus(401);
        $json = [];
    }
    $this->set(compact('json'));
    $this->viewBuilder()->setOption('serialize', 'json');
}
```

```
//In UsersController.php
public function login()
[...]

PR #464, opened 2021-07-03
Expand docs for #462

markstory:
Add missing docs that may help other folks in the same situation later.


docs/en/authenticators.rst
```diff
@@ -452,3 +456,29 @@ the redirect target safely from the query string parameter::
...
+path, or any other request attribute::
+
+    public function getAuthenticationService(
+        ServerRequestInterface $request
+    ): AuthenticationServiceInterface {
+        $service = new AuthenticationService();
+
+        // Configuration common to both the API and web goes here.
+
+        if ($request->getParam('prefix') == 'api') {
```
ndm2, line 477:
```suggestion
        if ($request->getParam('prefix') == 'Api') {
```

PR #467, opened 2021-07-10
Added the default option to add the application salt to the cookie

### LINKED ISSUE #439: Spoof session through remember-me cookie after database leak
Not entirely sure if I'm correct, but does the remember-me cookie allow users to impersonate someone of they had access to the application database?

The remember-me token is a [hash of the username and password hash](https://github.com/cakephp/authentication/blob/master/src/Authenticator/CookieAuthenticator.php#L152) (`token = hash(username + hash(password))`), so having the hashed password would be enough to create a correct token, which could be achieved through a database leak.

Obviously a database leak is already bad, though with form authentication having a hashed password doesn't get you anywhere yet.

If this is correct, would it be an option to add a feature for generated key-token pairs to use in cookie authentication?

RobertoRoos:
…defending against spoofs after e.g. a database leak. Now an attacker would need both the database and the filesystem in order to spoof someones session.

Fixes #439 (or partly at least)

### DIFF
--- a/docs/en/authenticators.rst
+++ b/docs/en/authenticators.rst
@@ -263,6 +263,7 @@ Configuration options:
    ``null`` and all pages will be checked.
 -  **passwordHasher**: Password hasher to use for token hashing. Default
    is ``DefaultPasswordHasher::class``.
+-  **salt**: When ``true``, the application's Security salt will be appended to the token before hashing. When ``false``, no salt is used. When a string is passed this will be used as a salt instead. Default is ``true``.
 
 Usage
 -----
--- a/docs/en/authenticators.rst
+++ b/docs/en/authenticators.rst
@@ -263,7 +263,10 @@ Configuration options:
    ``null`` and all pages will be checked.
 -  **passwordHasher**: Password hasher to use for token hashing. Default
    is ``DefaultPasswordHasher::class``.
--  **salt**: When ``true``, the application's Security salt will be appended to the token before hashing. When ``false``, no salt is used. When a string is passed this will be used as a salt instead. Default is ``true``.
+-  **salt**: When ``false`` no salt is used. When a string is passed that value is used as a salt value. 
+   When ``true`` the default Security.salt is used. Default is ``true``. When a salt is used, the cookie value 
+   will contain `hash(username + password + hmac(username + password, salt))`. This helps harden tokens against possible 
+   database leaks and enables cookie values to be invalidated by rotating the salt value.
 
 Usage
 -----
--- a/src/Authenticator/CookieAuthenticator.php
+++ b/src/Authenticator/CookieAuthenticator.php
@@ -22,6 +22,8 @@ use Authentication\PasswordHasher\PasswordHasherTrait;
 use Authentication\UrlChecker\UrlCheckerTrait;
 use Cake\Http\Cookie\Cookie;
 use Cake\Http\Cookie\CookieInterface;
+use Cake\Utility\Security;
+use InvalidArgumentException;
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
 use RuntimeException;
@@ -51,6 +53,7 @@ class CookieAuthenticator extends AbstractAuthenticator implements PersistenceIn
             'name' => 'CookieAuth',
         ],
         'passwordHasher' => 'Authentication.Default',
+        'salt' => true,
     ];
 
     /**
@@ -144,7 +147,7 @@ class CookieAuthenticator extends AbstractAuthenticator implements PersistenceIn
     /**
      * Creates a plain part of a cookie token.
      *
-     * Returns concatenated username and password hash.
+     * Returns concatenated username, password hash, and HMAC signature.
      *
      * @param array|\ArrayAccess $identity Identity data.
      * @return string
@@ -154,7 +157,23 @@ class CookieAuthenticator extends AbstractAuthenticator implements PersistenceIn
         $usernameField = $this->getConfig('fields.username');
         $passwordField = $this->getConfig('fields.password');
 
-        return $identity[$usernameField] . $identity[$passwordField];
+        $salt = $this->getConfig('salt', '');
+
+        $value = $identity[$usernameField] . $identity[$passwordField];
+
+        if ($salt === false) {
+            return $value;
+        }
+        if ($salt === true) {
+            $salt = Security::getSalt();
+        } elseif (!is_string($salt) || $salt === '') {
+            throw new InvalidArgumentException('Salt must be a non-empty string.');
+        }
+
+        $hmac = hash_hmac('sha1', $value, $salt);
+        // Instead of appending the plain salt, we create a hash. This limits the chance of the salt being leaked.
+
+        return $value . $hmac;
     }
 
     /**
--- a/tests/TestCase/Authenticator/CookieAuthenticatorTest.php
+++ b/tests/TestCase/Authenticator/CookieAuthenticatorTest.php
@@ -19,10 +19,12 @@ use ArrayObject;
 use Authentication\Authenticator\CookieAuthenticator;
 use Authentication\Authenticator\Result;
 use Authentication\Identifier\IdentifierCollection;
+use Cake\Core\Configure;
 use Cake\Http\Cookie\Cookie;
 use Cake\Http\Response;
 use Cake\Http\ServerRequestFactory;
 use Cake\TestSuite\TestCase;
+use InvalidArgumentException;
 use Psr\Http\Message\RequestInterface;
 use Psr\Http\Message\ResponseInterface;
 
@@ -45,6 +47,8 @@ class CookieAuthenticatorTest extends TestCase
     {
         $this->skipIf(!class_exists(Cookie::class));
 
+        // Note: security salt is written in tests/bootstrap.php
+
         parent::setUp();
     }
 
@@ -64,7 +68,7 @@ class CookieAuthenticatorTest extends TestCase
             null,
             null,
             [
-                'CookieAuth' => '["$2y$10$1bE1SgasKoz9WmEvUfuZLeYa6pQgxUIJ5LAoS/KGmC1hNuWkUG7ES"]',
+                'CookieAuth' => '["$2y$10$O5VgLDfIqszzr0Q47Ygkc.LkoLIwlIjc/OzoGp6yJasQlxcHU4.ES"]',
             ]
         );
 
@@ -91,7 +95,8 @@ class CookieAuthenticatorTest extends TestCase
             null,
             null,
             [
-                'CookieAuth' => '["mariano","$2y$10$1bE1SgasKoz9WmEvUfuZLeYa6pQgxUIJ5LAoS/KGmC1hNuWkUG7ES"]',
+                // hash(username . password . hmac(username . password, salt)
+                'CookieAuth' => '["mariano","$2y$10$RlCAFt3e/9l42f8SIaIbqejOg9/b/HklPo.fjXY.tFGuluafugssa"]',
             ]
         );
 
@@ -118,7 +123,7 @@ class CookieAuthenticatorTest extends TestCase
             null,
             null,
             [
-                'CookieAuth' => ['mariano', '$2y$10$1bE1SgasKoz9WmEvUfuZLeYa6pQgxUIJ5LAoS/KGmC1hNuWkUG7ES'],
+                'CookieAuth' => ['mariano', '$2y$10$RlCAFt3e/9l42f8SIaIbqejOg9/b/HklPo.fjXY.tFGuluafugssa'],
             ]
         );
 
@@ -127,6 +132,62 @@ class CookieAuthenticatorTest extends TestCase
 
         $this->assertInstanceOf(Result::class, $result);
         $this->assertSame(Result::SUCCESS, $result->getStatus());
+    }
+
+    /**
+     * testAuthenticateSuccessNoSalt
+     *
+     * @return void
+     */
+    public function testAuthenticateNoSalt()
+    {
+        Configure::delete('Security.salt');
+
+        $identifiers = new IdentifierCollection([
+            'Authentication.Password',
+        ]);
+
+        $request = ServerRequestFactory::fromGlobals(
+            ['REQUEST_URI' => '/testpath'],
+            null,
+            null,
+            [
+                // hash(username . password)
+                'CookieAuth' => '["mariano","$2y$10$yq91zLgrlF0TUzPjFj49DOL44svGrOYxaBfB6QYWEvxVKzNkvcVom"]',
+            ]
+        );
+
+        $authenticator = new CookieAuthenticator($identifiers, ['salt' => false]);
+        $result = $authenticator->authenticate($request);
+
+        $this->assertInstanceOf(Result::class, $result);
+        $this->assertSame(Result::SUCCESS, $result->getStatus());
+    }
+
+    /**
+     * testAuthenticateSuccessNoSalt
+     *
+     * @return void
+     */
+    public function testAuthenticateInvalidSalt()
+    {
+        $identifiers = new IdentifierCollection([
+            'Authentication.Password',
+        ]);
+
+        $request = ServerRequestFactory::fromGlobals(
+            ['REQUEST_URI' => '/testpath'],
+            null,
+            null,
+            [
+                'CookieAuth' => '["mariano","some_hash"]',
+            ]
+        );
+
+        $authenticator = new CookieAuthenticator($identifiers, ['salt' => '']);
+
+        $this->expectException(InvalidArgumentException::class);
+        $authenticator->authenticate($request);
     }
 
     /**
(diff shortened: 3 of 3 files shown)


src/Authenticator/CookieAuthenticator.php
```diff
@@ -51,6 +52,7 @@ class CookieAuthenticator extends AbstractAuthenticator implements PersistenceIn
             'name' => 'CookieAuth',
         ],
         'passwordHasher' => 'Authentication.Default',
+        'use_salt' => true,
```
dereuromark, line 55: Please use camelBacked syntax for option keys.

RobertoRoos: Of course, thanks.

src/Authenticator/CookieAuthenticator.php
```diff
@@ -51,6 +52,7 @@ class CookieAuthenticator extends AbstractAuthenticator implements PersistenceIn
             'name' => 'CookieAuth',
         ],
         'passwordHasher' => 'Authentication.Default',
+        'useSalt' => true,
```
ADmad, line 55: Since this changes how the cookie token is created by default, the PR needs to target `1.next`.

RobertoRoos: Yeah, will do. Gotta figure out how to rebase my branch properly.

ADmad: Rebasing will be easier if you first squash all the new commits in your feature branch into a single commit.

ndm2: But we'd want it in `2.x` too, don't we? There's no `2.next` branch (yet).

ADmad: Err sorry, I mixed up the numbers. We are on `2.x` series currently, not `1.x` :)

So this is fine since we don't have a `2.next`. We can just tag a new minor after this is merged.

src/Authenticator/CookieAuthenticator.php
```diff
@@ -154,7 +156,14 @@ protected function _createPlainToken($identity): string
         $usernameField = $this->getConfig('fields.username');
         $passwordField = $this->getConfig('fields.password');
 
-        return $identity[$usernameField] . $identity[$passwordField];
+        if (!$this->getConfig('useSalt')) {
+            return $identity[$usernameField] . $identity[$passwordField];
+        }
+
+        // Prevent user from unknowingly using an empty salt
+        $salt = Configure::readOrFail('Security.salt');
```
ADmad, line 164: How about also allowing to explicitly set a different salt through the authenticator config?

ndm2: Other authenticators support it, so I'd say yes.

Also this needs to use the `Security` class, as that config setting is by default being consumed in applications base on the app skeleton. And since there hasn't been a move yet to tie the plugin harder to the core, it should then also not be a hard dependency.

https://github.com/cakephp/authentication/blob/master/src/Authenticator/HttpDigestAuthenticator.php#L60

RobertoRoos: I'll change it to the Security utility.

About user defined salt, shall we make the config option called 'salt' and use it as:

- true: Security salt
- false: No salt
- string: user-defined salt

Is that intuitive?

docs/en/authenticators.rst
```diff
@@ -263,6 +263,7 @@ Configuration options:
    ``null`` and all pages will be checked.
 -  **passwordHasher**: Password hasher to use for token hashing. Default
    is ``DefaultPasswordHasher::class``.
+-  **salt**: When `true`, the application's Security salt will be appended to the token before hashing. When `false`, no salt is used. When a string is passed this will be used as a salt instead. Default is `true`.
```
markstory, line 266:
```suggestion
-  **salt**: When ``true``, the application's Security salt will be appended to the token before hashing. When ``false``, no salt is used. When a string is passed this will be used as a salt instead. Default is ``true``.
```

src/Authenticator/CookieAuthenticator.php
```diff
@@ -154,7 +156,17 @@ protected function _createPlainToken($identity): string
...
 
-        return $identity[$usernameField] . $identity[$passwordField];
+        if (!$this->getConfig('salt')) {
+            return $identity[$usernameField] . $identity[$passwordField];
+        }
+
+        if (is_string($this->getConfig('salt'))) {
+            $salt = $this->getConfig('salt');
+        } else {
+            $salt = Security::getSalt();
```
markstory, line 166: You could put the before the if to avoid the else.

saeideng, line 166: this class needs to `cakephp/utility` dependency

src/Authenticator/CookieAuthenticator.php
RobertoRoos: Still targeting master, I'll resolve this converstation.

src/Authenticator/CookieAuthenticator.php
RobertoRoos: I think that's personal preference. I prefer this:

```
if (...) {
    $x = $a;
} else {
    $x = $b;
}
```

Over this:

```
$x = $b;
if (...) {
    $x = $a;
}
```

(If I understand you correctly.)  
Can we leave this as it is?

RobertoRoos: Good find. I've added it here: https://github.com/cakephp/authentication/pull/467/files#diff-d2ab9925cad7eac58e0ff4cc0d251a937ecf49e4b6bf57f8b95aab76648a9d34R14

That seem right?

ndm2: That's what I ment with "hard dependency". The package is currently a suggestion, and other authenticators make it optional.:

https://github.com/cakephp/authentication/blob/aafb6831bd3a9e9f3607ffb254db529d5f6cf998/src/Authenticator/HttpDigestAuthenticator.php#L63-L65

RobertoRoos: Ended up changing this anyway because the of conversation below. Resolving this.

RobertoRoos: I see, thanks @ndm2 . Updated the code and will squash commits next.

othercorey: The `utility` package is already a dependency through `http`. This isn't doing anything useful.

https://github.com/cakephp/http/blob/4.x/composer.json#L29

RobertoRoos: True, but maybe that could change in the future. I'd say we keep it flexible as it is now (and as it was already for `HttpDigestAuthenticator` as @ndm2 pointed out).

othercorey: It isn't going to change. If we have a dependency, then it's valid to assume its use.

ndm2: Looks like the code in the other authenticators was added before the `utility` dependency was added to `http`. If it's an indirect dependency now, then it should indeed be fine, and the other authenticators could drop the optional handling too then.

src/Authenticator/CookieAuthenticator.php
```diff
@@ -154,7 +156,18 @@ protected function _createPlainToken($identity): string
...
-        return $identity[$usernameField] . $identity[$passwordField];
+        $salt = '';
+
+        if ($this->getConfig('salt')) {
+            if (is_string($this->getConfig('salt'))) {
+                $salt = $this->getConfig('salt');
+            } elseif (class_exists(Security::class)) {
+                $salt = Security::getSalt();
+                // Quietly skip salt if the Security utility is not available
+            }
```
ADmad, line 167: Never skip / hide issues. Let PHP generate an error so that the dev has useful information to act upon.

ADmad: If salt is enabled, a custom salt is not set and Security class is not available throw an exception.

RobertoRoos: That makes sense. We should change the same here then:

https://github.com/cakephp/authentication/blob/aafb6831bd3a9e9f3607ffb254db529d5f6cf998/src/Authenticator/HttpDigestAuthenticator.php#L63-L65

markstory: Why does that need to change? DigestAuthentication also ensures the secret is set https://github.com/cakephp/authentication/blob/aafb6831bd3a9e9f3607ffb254db529d5f6cf998/src/Authenticator/HttpDigestAuthenticator.php#L77-L80

RobertoRoos: I misunderstood. I've copied a similar check.

src/Authenticator/CookieAuthenticator.php
RobertoRoos: Right, I've changed it here. Should I change the same for HttpDigest in this PR?

src/Authenticator/CookieAuthenticator.php
```diff
@@ -154,7 +157,21 @@ protected function _createPlainToken($identity): string
...
+            } else {
+                $salt = Security::getSalt();
+            }
+
+            if (!is_string($salt) || strlen($salt) === 0) {
+                throw new InvalidArgumentException('Salt must be a non-empty string.');
+            }
+        }
+
+        return $identity[$usernameField] . $identity[$passwordField] . $salt;
```
ndm2, line 174:
```suggestion
        $value = $identity[$usernameField] . $identity[$passwordField];
        $hmac = hash_hmac('sha1', $value, $salt);

        return $value . $hmac;
```

I know the method says `Plain`, suggesting that the return value must not be used as-is, but I'd still feel better if the salt was kept as "internal" as possible and only used to "sign" data.

src/Authenticator/CookieAuthenticator.php
```diff
@@ -144,7 +147,7 @@ public function persistIdentity(ServerRequestInterface $request, ResponseInterfa
     /**
      * Creates a plain part of a cookie token.
      *
-     * Returns concatenated username and password hash.
+     * Returns concatenated username, password and application salt hash.
```
ndm2, line 150:
```suggestion
     * Returns concatenated username, password hash, and HMAC signature.
```

ndm2:
Would be nice if we could get this included sometime soon.

src/Authenticator/CookieAuthenticator.php
markstory: I don't understand why you're hashing here, the value that comes out of this method is hashed with the password hasher before being used in a cookie. Wouldn't that be enough?

markstory: Good point about the risk of having the bare salt included.

src/Authenticator/CookieAuthenticator.php
```diff
@@ -154,7 +157,21 @@ protected function _createPlainToken($identity): string
...
+            if (is_string($this->getConfig('salt'))) {
+                $salt = $this->getConfig('salt');
+            } else {
+                $salt = Security::getSalt();
+            }
+
+            if (!is_string($salt) || strlen($salt) === 0) {
+                throw new InvalidArgumentException('Salt must be a non-empty string.');
+            }
+        }
```
ADmad:
```suggestion
        $salt = $this->getConfig('salt');
        if ($salt === true) {
            $salt = Security::getSalt();
        } elseif ($salt === false) {
            $salt = '';
        } elseif (!is_string($salt) || $salt === '') {
            throw new InvalidArgumentException('Salt must be a non-empty string.');
        }
```

ADmad:
@RobertoRoos Are you still interested in finishing this PR?

src/Authenticator/CookieAuthenticator.php
RobertoRoos: It seems this is resolved. I won't change anything outside the cookie authenticator.

RobertoRoos:
> @RobertoRoos Are you still interested in finishing this PR?

Yeah. At the time I was waiting for the discussions to settle. I'll go over the outstanding points now.

src/Authenticator/CookieAuthenticator.php
RobertoRoos: Thanks for the suggestion, changed it.

src/Authenticator/CookieAuthenticator.php
RobertoRoos: With this, when the application salt is empty or an empty salt was passed to the authenticator config, it will quietly continue without salt. I like it better when there is a clear error. The developer should just set `'salt' => false` when the salt must be skipped after all.

ADmad: I have updated the previous comment to handle empty strings too.

RobertoRoos: Committed, thanks.

src/Authenticator/CookieAuthenticator.php
```diff
@@ -154,7 +157,21 @@ protected function _createPlainToken($identity): string
...
+            $salt = Security::getSalt();
+        } elseif ($salt === false) {
+            $salt = '';
+        } elseif (!is_string($salt) || $salt === '') {
+            throw new InvalidArgumentException('Salt must be a non-empty string.');
+        }
+
+        $value = $identity[$usernameField] . $identity[$passwordField];
+
+        $hmac = hash_hmac('sha1', $value, $salt);
```
ADmad, line 173: Shouldn't generating and appending the hmac be skipped if `$salt` is empty string to preserve backwards compatibility?

RobertoRoos:
I think this is ready to be merged now.
Unless I need to add increase test coverage? Codecov is complaining, about the lines of for parsing config not being covered.

ADmad:
> Unless I need to add increase test coverage? Codecov is complaining, about the lines of for parsing config not being covered.

Yes, please cover all new code paths.

src/Authenticator/CookieAuthenticator.php
RobertoRoos: Yeah, makes sense. I changed it.

src/Authenticator/CookieAuthenticator.php
```diff
@@ -158,16 +158,19 @@ protected function _createPlainToken($identity): string
...
+        if ($salt === false) {
+            return $value;
+        }
         if ($salt === true) {
             $salt = Security::getSalt();
-        } elseif ($salt === false) {
-            $salt = '';
-        } elseif (!is_string($salt) || $salt === '') {
+        }
+        if (!is_string($salt) || $salt === '') {
```
ADmad: This can still be `elseif`. Anyone setting Security salt to any empty string will have other problems already.

RobertoRoos: Ah, that's why I added that, didn't know. Changed it.

ADmad (approved):

docs/en/authenticators.rst
```diff
@@ -263,6 +263,7 @@ Configuration options:
    ``null`` and all pages will be checked.
 -  **passwordHasher**: Password hasher to use for token hashing. Default
    is ``DefaultPasswordHasher::class``.
+-  **salt**: When ``true``, the application's Security salt will be appended to the token before hashing. When ``false``, no salt is used. When a string is passed this will be used as a salt instead. Default is ``true``.
```
markstory:
```suggestion
-  **salt**: When ``false`` no salt is used. When a string is passed that value is used as a salt value. 
   When ``true`` the default Security.salt is used. Default is ``true``. When a salt is used, the cookie value 
   will contain `hash(username + password + hmac(username + password, salt))`. This helps harden tokens against possible 
   database leaks and enables cookie values to be invalidated by rotating the salt value.
```

tests/TestCase/Authenticator/CookieAuthenticatorTest.php
```diff
@@ -91,7 +95,8 @@ public function testAuthenticateSuccess()
             null,
             null,
             [
-                'CookieAuth' => '["mariano","$2y$10$1bE1SgasKoz9WmEvUfuZLeYa6pQgxUIJ5LAoS/KGmC1hNuWkUG7ES"]',
+                // hash(username . password . hmac(username . password, salt)
```
markstory:
```suggestion
                // hash(username . password . hmac(username . password, salt))
```

markstory (commented):
A few small corrections for the docs. Currently this will invalidate all existing cookie authentication tokens. That is something we will need to call out in the release notes.

ADmad:
Ping @ndm2 (for some reason the "Re-request review" action doesn't work for me for you).

tests/TestCase/Authenticator/CookieAuthenticatorTest.php
```diff
@@ -129,6 +134,62 @@ public function testAuthenticateExpandedCookie()
...
+            null,
+            null,
+            [
+                'CookieAuth' => '["mariano","some_hash"]',
+            ]
+        );
+
+        $authenticator = new CookieAuthenticator($identifiers, ['salt' => '']);
+
+        $this->expectException(InvalidArgumentException::class);
```
ndm2, line 189:
```suggestion
        $this->expectException(InvalidArgumentException::class);
        $this->expectExceptionMessage('Salt must be a non-empty string.');
```

So that not just any invalid argument exception is going to satisfy the test.

PR #481, opened 2021-09-05
Fix a phpstan error

tamurayk:
Fix
```
Call to an undefined method Authentication\AuthenticationService::allowUnauthenticated(). 
```

### DIFF
--- a/src/AuthenticationService.php
+++ b/src/AuthenticationService.php
@@ -30,6 +30,7 @@ use RuntimeException;
 
 /**
  * Authentication Service
+ *
  * @mixin \Authentication\Controller\Component\AuthenticationComponent
  */
 class AuthenticationService implements AuthenticationServiceInterface
--- a/src/AuthenticationService.php
+++ b/src/AuthenticationService.php
@@ -30,6 +30,7 @@ use RuntimeException;
 
 /**
  * Authentication Service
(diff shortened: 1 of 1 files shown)


src/AuthenticationService.php
```diff
@@ -30,6 +30,8 @@
 
 /**
  * Authentication Service
+ *
+ * @mixin \Authentication\Controller\Component\AuthenticationComponent
```
markstory, line 34: I don't think this is correct. This class does not proxy methods to the AuthenticationComponent at all. What call is causing that phpstan error in your application?

tamurayk: I'm sorry.
The error was due to my mistake.

The following is correct.

```
/**
 * @property AuthenticationComponent $Authentication
 */
class AppController extends Controller
{
    // ...

    public function initialize(): void
    {
        parent::initialize();

        $this->loadComponent('Authentication.Authentication');

        // ...
    }

    // ...
}
```

However, I had made the following mistakes.

```
/**
 * @property \Authentication\AuthenticationService $Authentication
 */
class AppController extends Controller
{
    // ...

    public function initialize(): void
    {
        parent::initialize();

        $this->loadComponent('Authentication.Authentication');

        // ...
    }

    // ...
}
```

PR #482, opened 2021-09-09
Avoid deprecation for php 8.1

CriztianiX:


othercorey:
You have to use `#[ReturnTypeWillChange]` attribute instead until authentication is 8.0+ only.

CriztianiX:
> You have to use `#[ReturnTypeWillChange]` attribute instead until authentication is 8.0+ only.

Ok!

othercorey:
I'm not sure why this isn't failing on php 8.1.

https://github.com/cakephp/authentication/runs/3577489214

Issue #497, opened 2021-11-19
HTTP Basic Auth without Forced Challenge


ravage84:
CakePHP 3.x
Authentication 1.x

We currently develop a RESTful API with HTTP basic auth with several endpoints.
Some endpoints need authentication, some do not.

I'm aware of the following issues and that we could use the URL prefix to handle whether to auth or not.

https://github.com/cakephp/authentication/issues/258
https://github.com/cakephp/authentication/issues/381

But I found a different solution, which is rather simple and provides a more flexible or let's say more selective solution.

```` php
    public function getAuthenticationService(ServerRequestInterface $request, ResponseInterface $response)
    {
        $service = new AuthenticationService();
        $service->loadAuthenticator(
            'OurApi.HttpBasic',
            [
                'className' => ApiHttpBasicAuthenticator::class,
            ]
        );

        // snip...

        return $service;
    }
````

So instead of using ``Authentication\Authenticator\HttpBasicAuthenticator``one could implement a simpler authenticator class using the same code as the original class, except the following:

1. Not extending the ``StatelessInterface`` 
2. ``unauthorizedChallenge()`` method
3. ``loginHeaders()`` method

This prevents any challenge being sent if the auth credentials aren't provided.

Then, one can load the ``AuthenticationComponent`` like so:

```` php
    public function initialize()
    {
        parent::initialize();

        $this->loadComponent(
            'Authentication.Authentication',
            [
                'requireIdentity' => true,
            ]
        );
    }
````

Lastly, if some endpoints do not need any auth, they can be excluded by using:

```` php
    public function beforeFilter(Event $event)
    {
        $this->Authentication->allowUnauthenticated(['ping']);

        return parent::beforeFilter($event);
    }
````

1. Is there a particular reason why this is not documented? E.g. not recommended.
2. Or at least provided by the plugin?

If not, would we want to have such an alternative HTTP basic auth authenticator?

ravage84:
Another use case for a non-challenging HTTP Basic Authenticator class is when you want to use a custom error handler, instead of letting the ``AuthenticationMiddleware`` [handle it](https://github.com/cakephp/authentication/blob/1.4.2/src/Middleware/AuthenticationMiddleware.php#L97-L105).

markstory:
If you are using basic auth and don't force a challenge how does the client know to prompt the user for basic authentication credentials?

ravage84:
@markstory first of all, I simply don't want he middleware to do the challenge as code after it should be reached.
[...]

PR #498, opened 2021-11-22
Allow skipping the challenge exception generation for Http Basic/Digest Auth.

ADmad:
Also updated HttpBasicAuthenticator to set proper result status
depending on whether credentials are missing or identity is not found.

Refs #497.

### DIFF
--- a/src/Authenticator/HttpBasicAuthenticator.php
+++ b/src/Authenticator/HttpBasicAuthenticator.php
@@ -34,39 +34,24 @@ class HttpBasicAuthenticator extends AbstractAuthenticator implements StatelessI
      */
     public function authenticate(ServerRequestInterface $request): ResultInterface
     {
-        $user = $this->getUser($request);
+        $server = $request->getServerParams();
+        $username = $server['PHP_AUTH_USER'] ?? '';
+        $password = $server['PHP_AUTH_PW'] ?? '';
 
-        if (empty($user)) {
+        if ($username === '' || $password === '') {
             return new Result(null, Result::FAILURE_CREDENTIALS_MISSING);
         }
 
-        return new Result($user, Result::SUCCESS);
-    }
-
-    /**
-     * Get a user based on information in the request. Used by cookie-less auth for stateless clients.
-     *
-     * @param \Psr\Http\Message\ServerRequestInterface $request Request object.
-     * @return \ArrayAccess|array|null User entity or null on failure.
-     */
-    public function getUser(ServerRequestInterface $request)
-    {
-        $server = $request->getServerParams();
-        if (!isset($server['PHP_AUTH_USER']) || !isset($server['PHP_AUTH_PW'])) {
-            return null;
-        }
-
-        $username = $server['PHP_AUTH_USER'];
-        $password = $server['PHP_AUTH_PW'];
-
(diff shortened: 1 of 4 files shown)


ADmad:
This will require a new minor release.

tests/TestCase/AuthenticationServiceTest.php
```diff
@@ -104,6 +103,29 @@ public function testAuthenticateWithChallenge()
         }
     }
 
+    public function testAuthenticateHttpAuthCredentialsPresent()
+    {
+        $request = ServerRequestFactory::fromGlobals([
+            'SERVER_NAME' => 'example.com',
+            'REQUEST_URI' => '/testpath',
+            'PHP_AUTH_USER' => 'mariano',
+            'PHP_AUTH_PW' => 'WRONG',
```
markstory, line 118: It might be good to check that not sending a challenge still lets the user re-enter their credentials in a browser when they make a mistake.

ADmad: My mistake, I had overlooked this case. Pretty sure the browser won't show the dialog to re-enter the credentials without the `www-authenticate` header. I checked using HTTP Basic Auth through Apache config and it too returns a 401 with the `www-authenticate` header in case of incorrect credentials.

So I have reverted the changes I had done in `AuthenticationService`. Just updated `HttpBasicAuthenticator` to return `Result` with proper status when credentials are missing vs. incorrect credentials.

Should I add a config to the authenticator to allow skipping the challenge in case of missing/incorrect credentials?

markstory: > Should I add a config to the authenticator to allow skipping the challenge in case of missing/incorrect credentials?

That sounds good to me. I don't think not emitting the challenge justifies another adapter.

src/Authenticator/HttpBasicAuthenticator.php
```diff
@@ -33,40 +33,25 @@ class HttpBasicAuthenticator extends AbstractAuthenticator implements StatelessI
...
-        }
+        $username = $server['PHP_AUTH_USER'] ?? '';
+        $password = $server['PHP_AUTH_PW'] ?? '';
 
-        $username = $server['PHP_AUTH_USER'];
-        $password = $server['PHP_AUTH_PW'];
-
-        if (!is_string($username) || $username === '' || !is_string($password) || $password === '') {
-            return null;
+        if ($username === '' || $password === '') {
```
markstory, line 57: Could we have non-scalar values in username or password here? I think we should be returning credentials missing if either password or username are not scalar.

ADmad: I am unaware how user could provide non string values for the `PHP_AUTH_USER` and `PHP_AUTH_PW` env vars.

tests/TestCase/AuthenticationServiceTest.php
ADmad: Done

src/Authenticator/HttpBasicAuthenticator.php
markstory: True, I've been burned too many times from PHP surprises I guess. :smile:

markstory (approved):

PR #500, opened 2021-11-25
Bump firebase/php-jwt to v5.5.

markstory:
Update JwtAutenticator to use the new "algorithm" config and deprecate "algorithms" config.

Refs https://github.com/firebase/php-jwt/releases/tag/v5.5.0

This will be the start of the next minor release for authentication. Before this is merged we should do a bugfix release to address how configuration is merged.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,7 @@
     "require-dev": {
         "cakephp/cakephp": "^4.0",
         "cakephp/cakephp-codesniffer": "^4.0",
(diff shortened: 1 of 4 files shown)


pabloelcolombiano (approved):
This looks good to me.

ADmad:
> Before this is merged we should do a bugfix release to address how configuration is merged.

I have opened a PR for it #501.

PR #506, opened 2022-01-01
Way to access JWT payload

RiteshParyali:
An instruction to access JWT payload .

A me

### DIFF
--- a/docs/en/authentication-component.rst
+++ b/docs/en/authentication-component.rst
@@ -62,6 +62,24 @@ made. For example LDAP or OAuth would put errors specific to their
 implementation in here for easier logging and debugging the cause. But most of
 the included authenticators don't put anything in here.
 
+
+Access JWT payload 
+------------------------
+
+ To access jwt payload ::
+ 
+    use Firebase\JWT\JWT;
+
+    $privateKey = file_get_contents(CONFIG . '/jwt.pem');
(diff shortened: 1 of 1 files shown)


markstory:
Were you not able to use the `returnPayload` [option of the JwtAuthenticator](https://book.cakephp.org/authentication/2/en/authenticators.html#jwt)? 

Parsing a JWT token in application code is something I would prefer we didn't include in the docs as we should have abstractions so that configuration data doesn't need to be duplicated.

RiteshParyali:
Thank you. 
But still have an issue..

when returnPayload is false **$this->Authentication->getIdentity()** , is returning the data/row for the users from table.
and when returnPayload is true , then **$this->Authentication->getIdentity()** is returning only the _payload_,

Is is possible to get both ?

markstory:
> Is is possible to get both ?

No. You can write your own authentication adapter if you need that.

RiteshParyali:
> > Is is possible to get both ?
> 
> No. You can write your own authentication adapter if you need that.

Then this instruction can be useful

markstory:
> Then this instruction can be useful

Sure, but including documentation for every potential task a user might want to do with their application code would make the documentation hard to navigate and use.

ADmad:
In general instructions for how to extend and modify a class to suit ones specific needs is not something we can include in the docs.

You can use one of our support forums for help.

PR #507, opened 2022-01-24
bump to php-jwt 6.2 and remove temporary/deprecated code

swiffer:


markstory (commented):
Looks good if we can get the lint error fixed up.

ndm2:
I'm surprised that this didn't break any tests, given the breaking change in `JWT::decode()`. Seems like there never were any tests for anything other than the default algorithm(s) config.

ADmad:
> I'm surprised that this didn't break any tests, given the breaking change in JWT::decode(). Seems like there never were any tests for anything other than the default algorithm(s) config.

Me too. Good thing that we have static analysis :slightly_smiling_face:.

Also this will be a breaking change for users and even a minor version bump will not be enough IMO. The JWT authentication will simply break unless users upgrade the `firebase/php-jwt` packages in their apps.

cnizzardini:
Docs should be updated as part of this, especially since this PR: https://github.com/cakephp/authentication/pull/531

swiffer:
I've updated this PR.

I merged 2.x into 2.next as 2.next was several commits behind.
Docs have been updated, deprecated code has been removed.
I bumped php-jwt to [^6.2](https://github.com/firebase/php-jwt/releases/tag/v6.2.0) as this will add support for cached JWKS keysets (to be used in an upcoming PR)

As suggested by @ADmad should this become v 3.0 instead ?

ADmad (approved):
LGTM, this would be a new major release, potentially along with changes in #535.

ADmad:
@markstory ping

PR #510, opened 2022-01-31
Adds tls option to LDAP identifier

eardkdw:
Adds an option to start TLS on connection to an LDAP server. Defaults to false.

### DIFF
--- a/docs/en/identifiers.rst
+++ b/docs/en/identifiers.rst
@@ -92,7 +92,7 @@ Configuration options:
    ``\Authentication\Identifier\Ldap\ExtensionAdapter``. You can pass a
    custom object/classname here if it implements the
    ``AdapterInterface``.
--  **tls**: Boolean. If ``True``, tries to start TLS on the connection.
+-  **tls**: Boolean. If ``true``, tries to start TLS on the connection.
 -  **options**: Additional LDAP options, like
    ``LDAP_OPT_PROTOCOL_VERSION`` or ``LDAP_OPT_NETWORK_TIMEOUT``. See
    `php.net <http://php.net/manual/en/function.ldap-set-option.php>`__
--- a/docs/en/identifiers.rst
+++ b/docs/en/identifiers.rst
@@ -92,6 +92,7 @@ Configuration options:
    ``\Authentication\Identifier\Ldap\ExtensionAdapter``. You can pass a
    custom object/classname here if it implements the
    ``AdapterInterface``.
+-  **tls**: Boolean. If ``True``, tries to start TLS on the connection.
 -  **options**: Additional LDAP options, like
    ``LDAP_OPT_PROTOCOL_VERSION`` or ``LDAP_OPT_NETWORK_TIMEOUT``. See
    `php.net <http://php.net/manual/en/function.ldap-set-option.php>`__
--- a/docs/en/identifiers.rst
+++ b/docs/en/identifiers.rst
@@ -92,9 +92,10 @@ Configuration options:
    ``\Authentication\Identifier\Ldap\ExtensionAdapter``. You can pass a
    custom object/classname here if it implements the
    ``AdapterInterface``.
--  **tls**: Boolean. If ``true``, tries to start TLS on the connection.
--  **options**: Additional LDAP options, like
-   ``LDAP_OPT_PROTOCOL_VERSION`` or ``LDAP_OPT_NETWORK_TIMEOUT``. See
+-  **options**: Array of additional LDAP options, including
+    ``tls``: Boolean. If ``true``, tries to start TLS on the connection. 
+    Also LDAP config options such as
+    ``LDAP_OPT_PROTOCOL_VERSION`` or ``LDAP_OPT_NETWORK_TIMEOUT``. See
    `php.net <http://php.net/manual/en/function.ldap-set-option.php>`__
    for more valid options.
 
--- a/src/Identifier/Ldap/AdapterInterface.php
+++ b/src/Identifier/Ldap/AdapterInterface.php
@@ -36,7 +36,7 @@ interface AdapterInterface
      * @param bool $tls If true try starting TLS.
      * @return void
      */
-    public function connect(string $host, int $port, array $options, bool $tls): void;
+    public function connect(string $host, int $port, array $options, bool $tls = false): void;
 
     /**
      * Unbind from LDAP directory
--- a/src/Identifier/Ldap/AdapterInterface.php
+++ b/src/Identifier/Ldap/AdapterInterface.php
@@ -33,10 +33,9 @@ interface AdapterInterface
      * @param string $host Hostname
      * @param int $port Port
      * @param array $options Additional options
-     * @param bool $tls If true try starting TLS.
      * @return void
      */
-    public function connect(string $host, int $port, array $options, bool $tls = false): void;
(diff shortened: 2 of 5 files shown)


docs/en/identifiers.rst
```diff
@@ -92,6 +92,7 @@ Configuration options:
    ``\Authentication\Identifier\Ldap\ExtensionAdapter``. You can pass a
    custom object/classname here if it implements the
    ``AdapterInterface``.
+-  **tls**: Boolean. If ``True``, tries to start TLS on the connection.
```
dereuromark:
```suggestion
-  **tls**: Boolean. If ``true``, tries to start TLS on the connection.
```

src/Identifier/Ldap/AdapterInterface.php
```diff
@@ -35,7 +35,7 @@ public function bind(string $bind, string $password): bool;
      * @param array $options Additional options
      * @return void
      */
-    public function connect(string $host, int $port, array $options): void;
+    public function connect(string $host, int $port, array $options, bool $tls): void;
```
dereuromark: couldnt tls default to false to make this less hard a bc break?
But since this is mainly used internally, I guess it is fine as is.

src/Identifier/Ldap/ExtensionAdapter.php
```diff
@@ -91,13 +91,21 @@ public function getConnection()
...
     {
         $this->_setErrorHandler();
         $resource = ldap_connect($host, $port);
         if ($resource === false) {
             throw new RuntimeException('Unable to connect to LDAP server.');
         }
+        if($tls)
+        {
+            //convert the connection to TLS
+            $ldap_started = ldap_start_tls($resource);
```
dereuromark: var naming invalid

docs/en/identifiers.rst
eardkdw: done

src/Identifier/Ldap/AdapterInterface.php
eardkdw: It does default to false, see default config on the src/Identifier/LdapIdentifier.php

eardkdw: the default port number is set in the default config too

src/Identifier/Ldap/ExtensionAdapter.php
eardkdw: Eliminated variable completely as redundant

markstory (commented):
Looks good to me once the formatting issues are fixed up.

eardkdw:
I don't understand the phpstan error.

src/Identifier/Ldap/AdapterInterface.php
```diff
@@ -33,9 +33,10 @@ public function bind(string $bind, string $password): bool;
      * @param string $host Hostname
      * @param int $port Port
      * @param array $options Additional options
+     * @param bool $tls if to try starting TLS
```
ADmad:
```suggestion
     * @param bool $tls If true try starting TLS.
```

src/Identifier/Ldap/AdapterInterface.php
```diff
@@ -33,9 +33,10 @@ public function bind(string $bind, string $password): bool;
...
      * @return void
      */
-    public function connect(string $host, int $port, array $options): void;
+    public function connect(string $host, int $port, array $options, bool $tls): void;
```
ADmad:
```suggestion
    public function connect(string $host, int $port, array $options, bool $tls = false): void;
```

src/Identifier/Ldap/ExtensionAdapter.php
```diff
@@ -89,15 +89,22 @@ public function getConnection()
...
     {
         $this->_setErrorHandler();
         $resource = ldap_connect($host, $port);
         if ($resource === false) {
             throw new RuntimeException('Unable to connect to LDAP server.');
         }
+        if ($tls) {
+            //convert the connection to TLS
+            if (!ldap_start_tls($resource)) {
+                throw new RuntimeException('Starting TLS failed on connetion to LDAP server.');
```
ADmad:
```suggestion
                throw new RuntimeException('Starting TLS failed on connection to LDAP server.');
```

src/Identifier/LdapIdentifier.php
```diff
@@ -177,7 +178,8 @@ protected function _connectLdap(): void
         $this->_ldap->connect(
             $config['host'],
             $config['port'],
-            (array)$this->getConfig('options')
+            (array)$this->getConfig('options'),
+            (bool)$config['tls'],
```
ADmad:
```suggestion
            (bool)$config['tls']
```

ADmad (changes requested):

src/Identifier/Ldap/ExtensionAdapter.php
```diff
@@ -89,15 +89,22 @@ public function getConnection()
      * @param string $host Hostname
      * @param int $port Port
      * @param array $options Additonal LDAP options
+     * @param bool $tls if to try starting TLS
      * @return void
      */
-    public function connect(string $host, int $port, array $options): void
+    public function connect(string $host, int $port, array $options, bool $tls = false): void
```
ADmad: The signature change makes this a non backwards compatible change and would require a major version bump.

Couldn't we instead use `tls` key in `$options` instead and unset it before `$options` are looped over for `setOption()` calls?

eardkdw: Is it not backwards-compatible if it defaults to false? 

I did consider adding it to the options array (or just checking to see if any `*_TLS_*` options had been set) but it seemed clumsy.

ADmad: > Is it not backwards-compatible if it defaults to false?

For a user it is, but since the interface's signature changed any custom implementation of it would break.

eardkdw: Fair enough, although I still think that it is pretty clumsy

eardkdw: Done

docs/en/identifiers.rst
```diff
@@ -92,6 +92,7 @@ Configuration options:
    ``\Authentication\Identifier\Ldap\ExtensionAdapter``. You can pass a
    custom object/classname here if it implements the
    ``AdapterInterface``.
+-  **tls**: Boolean. If ``true``, tries to start TLS on the connection.
```
ADmad: The `tls` option is now under the `options` key, so the description needs to be updated accordingly.

src/Identifier/Ldap/ExtensionAdapter.php
```diff
@@ -98,6 +98,13 @@ public function connect(string $host, int $port, array $options): void
         if ($resource === false) {
             throw new RuntimeException('Unable to connect to LDAP server.');
         }
+        if ($options['tls']) {
```
ADmad:
```suggestion
        if (isset($options['tls']) && $options['tls']) {
```

othercorey: You could do `$options['tls'] ?? false`.

markstory:
@ADmad Does anything else need to change on this? It looks good to me.

Issue #473, opened 2021-08-08
Discussion: Impersonation/Masquerade feature


LordSimal:
As a member of the CakePHP discourse already askeded here https://discourse.cakephp.org/t/cakephp-4-x-impersonate-login-as-different-user/9559 an impersonation/masquerade feature would be very useful.

raul338 already presented a solution with a custom Authenticator based on the SessionAuthenticator here: https://discourse.cakephp.org/t/cakephp-4-x-impersonate-login-as-different-user/9559/3

The question now is if such a impersonation logic will always have to be implemented per user since the logic **which user can impersonate** can differ widely or if this can be implemented in this plugin (or an extra plugin)?

Was just curious what your opinions on that would be.

markstory:
I like the general direction of those changes. If we added an impersonation feature to the plugin I would like to see it integrated more robustly though.

How would other persistent backends implement this? Currently the example directly writes to the session. Instead I would prefer we had an interface like:

```php
// In a controller
$this->Authorization->impersonate($user)
```

By having this more abstract interface we could enable other persistent backends provide the same behavior. The other persistent adapters we provide is the `CookieAuthenticator`. A similar key could work there, but I've not thought through all the implications it could create.

steinkel:
While working on the implementation for the impersonation feature we were discussing about CookieAuthenticator. In all the cases we typically use this feature, we never assume the impersonation should survive the session. Even if we could implement CookieAuthenticator to also store the impersonation related data, or add a new cookie with the impersonation data, we think this case will be very rare. So we would NOT be implementing the CookieAuthenticator to store the impersonated user or data, if you loose the session, you will need to impersonate another user again.

What are your thoughts on this?

Thanks,

LordSimal:
I would also agree that session based impersonation is the main focus here.

jeremyharris:
Fwiw, my personal implementation of impersonation is session only and so far it's never been expected to survive past a session. 

It does feel buggy, however, when the session expires during impersonation:
- user A impersonates B
- visit page only B has permission to
- session expires
- the page reloads, authenticates with user A cookie
- error: user A cannot view this page 

One could argue this is a messaging problem or an application problem, I'm just bringing it up to share my experience.

markstory:
I'm fine with session only impersonation.

markstory:
Implemented in #532

PR #532, opened 2022-05-06
Implement Impersonation 

ajibarra:
Refs #473

### DIFF
--- a/src/AuthenticationService.php
+++ b/src/AuthenticationService.php
@@ -219,10 +219,10 @@ class AuthenticationService implements AuthenticationServiceInterface, Impersona
                 if ($authenticator->isImpersonating($request)) {
                     /** @psalm-var array{request: \Cake\Http\ServerRequest, response: \Cake\Http\Response} $stopImpersonationResult */
                     $stopImpersonationResult = $authenticator->stopImpersonating($request, $response);
-                    [$request, $response] = $stopImpersonationResult;
+                    ['request' => $request, 'response' => $response] = $stopImpersonationResult;
                 }
                 $result = $authenticator->clearIdentity($request, $response);
-                [$request, $response] = $result;
+                ['request' => $request, 'response' => $response] = $result;
             }
         }
         $this->_successfulAuthenticator = null;
--- a/src/AuthenticationService.php
+++ b/src/AuthenticationService.php
@@ -215,6 +215,7 @@ class AuthenticationService implements AuthenticationServiceInterface
     {
         foreach ($this->authenticators() as $authenticator) {
             if ($authenticator instanceof PersistenceInterface) {
+                $authenticator->stopImpersonate($request, $response);
                 $result = $authenticator->clearIdentity($request, $response);
                 $request = $result['request'];
                 $response = $result['response'];
@@ -427,5 +428,51 @@ class AuthenticationService implements AuthenticationServiceInterface
         }
 
         return $parsed['path'] . $parsed['query'];
+    }
+
+    /**
+     * Impersonates a user
+     *
+     * @param ServerRequestInterface $request
+     * @param ResponseInterface $response
+     * @param \ArrayAccess|array $impersonator
+     * @param \ArrayAccess|array $impersonated
+     * @return \ArrayAccess|array
+     */
+    public function impersonate(ServerRequestInterface $request, ResponseInterface $response, \ArrayAccess|array $impersonator, \ArrayAccess|array $impersonated): \ArrayAccess|array
+    {
+        /** @var PersistenceInterface $provider */
+        $provider = $this->getAuthenticationProvider();
+        if (!($provider instanceof PersistenceInterface)) {
+            throw new \InvalidArgumentException('Provider must implement PersistenceInterface to be able to impersonate user.');
+        }
+        return $provider->impersonate($request, $response, $impersonator, $impersonated);
+    }
+
+    /**
+     * Stops impersonation
+     *
+     * @param ServerRequestInterface $request
+     * @param ResponseInterface $response
+     * @return \ArrayAccess|array
+     */
+    public function stopImpersonate(ServerRequestInterface $request, ResponseInterface $response): \ArrayAccess|array
+    {
+        /** @var PersistenceInterface $provider */
+        $provider = $this->getAuthenticationProvider();
+        if (!($provider instanceof PersistenceInterface)) {
+            throw new \InvalidArgumentException('Provider must implement PersistenceInterface to be able to impersonate user.');
+        }
+        return $provider->stopImpersonate($request, $response);
+    }
+
+    public function isImpersonating(ServerRequestInterface $request, ResponseInterface $response): bool
+    {
+        /** @var PersistenceInterface $provider */
+        $provider = $this->getAuthenticationProvider();
+        if (!($provider instanceof PersistenceInterface)) {
+            throw new \InvalidArgumentException('Provider must implement PersistenceInterface to be able to impersonate user.');
+        }
+        return $provider->isImpersonating($request, $response);
     }
 }
--- a/src/AuthenticationService.php
+++ b/src/AuthenticationService.php
@@ -490,7 +490,7 @@ class AuthenticationService implements AuthenticationServiceInterface, Impersona
      */
     public function isImpersonating(ServerRequestInterface $request): bool
     {
-        /** @var \Authentication\Authenticator\PersistenceInterface $provider */
+        /** @var \Authentication\Authenticator\ImpersonationInterface $provider */
         $provider = $this->getAuthenticationProvider();
         if (!($provider instanceof ImpersonationInterface)) {
             $className = get_class($provider);
--- a/src/AuthenticationService.php
+++ b/src/AuthenticationService.php
@@ -18,6 +18,7 @@ namespace Authentication;
 
 use Authentication\Authenticator\AuthenticatorCollection;
 use Authentication\Authenticator\AuthenticatorInterface;
+use Authentication\Authenticator\ImpersonationInterface;
 use Authentication\Authenticator\PersistenceInterface;
 use Authentication\Authenticator\ResultInterface;
 use Authentication\Authenticator\StatelessInterface;
@@ -218,12 +219,10 @@ class AuthenticationService implements AuthenticationServiceInterface
                 if ($authenticator->isImpersonating($request)) {
                     /** @psalm-var array{request: \Cake\Http\ServerRequest, response: \Cake\Http\Response} $stopImpersonationResult */
                     $stopImpersonationResult = $authenticator->stopImpersonating($request, $response);
-                    $request = $stopImpersonationResult['request'];
-                    $response = $stopImpersonationResult['response'];
+                    [$request, $response] = $stopImpersonationResult;
                 }
                 $result = $authenticator->clearIdentity($request, $response);
-                $request = $result['request'];
-                $response = $result['response'];
+                [$request, $response] = $result;
             }
         }
         $this->_successfulAuthenticator = null;
@@ -452,9 +451,10 @@ class AuthenticationService implements AuthenticationServiceInterface
     ): array {
         /** @var \Authentication\Authenticator\PersistenceInterface $provider */
         $provider = $this->getAuthenticationProvider();
-        if (!($provider instanceof PersistenceInterface)) {
+        if (!($provider instanceof ImpersonationInterface)) {
+            $className = get_class($provider);
             throw new \InvalidArgumentException(
-                'Provider must implement PersistenceInterface to be able to impersonate user.'
+                "The {$className} Provider must implement ImpersonationInterface in order to use impersonation."
             );
         }
 
@@ -472,9 +472,10 @@ class AuthenticationService implements AuthenticationServiceInterface
     {
         /** @var \Authentication\Authenticator\PersistenceInterface $provider */
         $provider = $this->getAuthenticationProvider();
-        if (!($provider instanceof PersistenceInterface)) {
+        if (!($provider instanceof ImpersonationInterface)) {
+            $className = get_class($provider);
             throw new \InvalidArgumentException(
-                'Provider must implement PersistenceInterface to be able to impersonate user.'
+                "The {$className} Provider must implement ImpersonationInterface in order to use impersonation."
             );
         }
 
@@ -491,9 +492,10 @@ class AuthenticationService implements AuthenticationServiceInterface
     {
         /** @var \Authentication\Authenticator\PersistenceInterface $provider */
         $provider = $this->getAuthenticationProvider();
-        if (!($provider instanceof PersistenceInterface)) {
+        if (!($provider instanceof ImpersonationInterface)) {
+            $className = get_class($provider);
             throw new \InvalidArgumentException(
-                'Provider must implement PersistenceInterface to be able to impersonate user.'
+                "The {$className} Provider must implement ImpersonationInterface in order to use impersonation."
             );
         }
 
--- a/src/AuthenticationService.php
+++ b/src/AuthenticationService.php
@@ -32,7 +32,7 @@ use RuntimeException;
 /**
  * Authentication Service
  */
-class AuthenticationService implements AuthenticationServiceInterface
+class AuthenticationService implements AuthenticationServiceInterface, ImpersonationInterface
 {
     use InstanceConfigTrait;
 
@@ -484,7 +484,7 @@ class AuthenticationService implements AuthenticationServiceInterface
     /**
      * Get impersonation provider
      *
-     * @return ImpersonationInterface
+     * @return \Authentication\Authenticator\ImpersonationInterface
      * @throws \InvalidArgumentException
      */
     protected function getImpersonationProvider(): ImpersonationInterface
@@ -497,6 +497,7 @@ class AuthenticationService implements AuthenticationServiceInterface
                 "The {$className} Provider must implement ImpersonationInterface in order to use impersonation."
             );
         }
+
         return $provider;
     }
 }
--- a/src/AuthenticationService.php
+++ b/src/AuthenticationService.php
@@ -32,7 +32,7 @@ use RuntimeException;
 /**
  * Authentication Service
  */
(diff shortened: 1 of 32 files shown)


src/AuthenticationService.php
```diff
@@ -428,4 +429,50 @@ public function getLoginRedirect(ServerRequestInterface $request): ?string
...
+    }
+
+    /**
+     * Stops impersonation
+     *
+     * @param ServerRequestInterface $request
+     * @param ResponseInterface $response
+     * @return \ArrayAccess|array
+     */
+    public function stopImpersonate(ServerRequestInterface $request, ResponseInterface $response): \ArrayAccess|array
```
markstory: Should this method be `stopImpersonating()` or `stopImpersonation()`?

src/Authenticator/CookieAuthenticator.php
```diff
@@ -252,4 +252,42 @@ protected function _createCookie($value): CookieInterface
...
+    /**
+     * Returns true if impersonation is being done
+     *
+     * @param ServerRequestInterface $request
+     * @param ResponseInterface $response
+     * @return bool
+     */
+    public function isImpersonating(ServerRequestInterface $request, ResponseInterface $response): bool
+    {
+        // TODO: Implement isImpersonating() method.
```
markstory: Will you need the response for checking if impersonation is active? Seems like that could use only the request.

src/Authenticator/SessionAuthenticator.php
```diff
@@ -115,4 +117,68 @@ public function clearIdentity(ServerRequestInterface $request, ResponseInterface
...
+     * @param \ArrayAccess|array $impersonated
+     * @return \ArrayAccess|array
+     */
+    public function impersonate(ServerRequestInterface $request, ResponseInterface $response, \ArrayAccess|array $impersonator, \ArrayAccess|array $impersonated): \ArrayAccess|array
+    {   $sessionKey = $this->getConfig('sessionKey');
+        $impersonateSessionKey = $this->getConfig('impersonateSessionKey');
+        /** @var \Cake\Http\Session $session */
+        $session = $request->getAttribute('session');
+        if ($session->check($impersonateSessionKey)) {
+            //throw new UnauthorizedException('You are already impersonating a user.');
```
markstory: I think it reasonable to overwrite the existing impersonation if another is started.

src/Authenticator/SessionAuthenticator.php
```diff
@@ -115,4 +117,68 @@ public function clearIdentity(ServerRequestInterface $request, ResponseInterface
...
+     */
+    public function stopImpersonate(ServerRequestInterface $request, ResponseInterface $response): \ArrayAccess|array
+    {
+        $sessionKey = $this->getConfig('sessionKey');
+        $impersonateSessionKey = $this->getConfig('impersonateSessionKey');
+        /** @var \Cake\Http\Session $session */
+        $session = $request->getAttribute('session');
+        if (!$session->check($impersonateSessionKey)) {
+           // throw new UnauthorizedException('You are not impersonating a user.');
+        }
```
markstory: I think this is a reasonable error case to have.

markstory (commented):
You should run `vendor/bin/phpcbf` on your changes to fix the formatting issues.

ajibarra:
> You should run `vendor/bin/phpcbf` on your changes to fix the formatting issues.

Yeah..indeed it is still missing:

* ~Implement impersonation for CookieAuthenticator~ **NOT IMPLEMENTED (See #473)**
* ~Compatible with 7.X (Remove multiple return types, maybe something else)~
* ~Fix Code standards~
* ~Unit tests~
* ~Enable exceptions after finish implementation and tests~

But it is great to get some feedback about approach and code.

Thanks @markstory

markstory:
It would be helpful to know more about how you see developers using these features. Some use-cases that would be good to have examples of are:

1. How does a developer enable account impersonation?
2. Once active does impersonation stay active until it is stopped?
3. How does one turn off impersonation?

We'll need these use-cases to write the docs, so having them now will help when it is time to write the documentation. Other questions, I have around impersonation are:

* Should there be a time limit that impersonation is active for? If we don't want to build that into this plugin how would someone go about building that?
* Should we emit log messages when impersonation is activated, or should that be an application concern?

src/AuthenticationService.php
ajibarra: Done

src/Authenticator/SessionAuthenticator.php
ajibarra: Makes sense

src/Authenticator/SessionAuthenticator.php
ajibarra: Removed //

src/Authenticator/SessionAuthenticator.php
ajibarra: Wait, doing some tests I found this use case:

* User A impersonates User B
* User B then impersonates User A
* User B (as User A) stops impersonation
* User B is logged in and no impersonation trace

So maybe we should not allow if user is impersonating already OR another way would be checking if user is being impersonated, and dont replace impersonator data but only impersonated one.

What do you think?

ADmad: I don't think multi-level / nested impersonation should be allowed. It will make thinks overly complex.

markstory: Agree that we should only support a single level of impersonation. The scenarios where I've used impersonation in the past are in a developer/support person needing to view a user's account as that user. I've not ever needed nested impersonation in that scenario.

src/AuthenticationService.php
```diff
@@ -215,6 +215,12 @@ public function clearIdentity(ServerRequestInterface $request, ResponseInterface
     {
         foreach ($this->authenticators() as $authenticator) {
             if ($authenticator instanceof PersistenceInterface) {
+                if ($authenticator->isImpersonating($request)) {
+                    /** @psalm-var array{request: \Cake\Http\ServerRequest, response: \Cake\Http\Response} $stopImpersonationResult */
+                    $stopImpersonationResult = $authenticator->stopImpersonating($request, $response);
+                    $request = $stopImpersonationResult['request'];
+                    $response = $stopImpersonationResult['response'];
```
markstory: Should we use list destructuring instead? `[$request, $response] = $stopImpersonationResult;` ?

src/AuthenticationService.php
```diff
@@ -428,4 +434,69 @@ public function getLoginRedirect(ServerRequestInterface $request): ?string
...
+        ResponseInterface $response,
+        \ArrayAccess $impersonator,
+        \ArrayAccess $impersonated
+    ): array {
+        /** @var \Authentication\Authenticator\PersistenceInterface $provider */
+        $provider = $this->getAuthenticationProvider();
+        if (!($provider instanceof PersistenceInterface)) {
+            throw new \InvalidArgumentException(
+                'Provider must implement PersistenceInterface to be able to impersonate user.'
+            );
```
markstory:
```suggestion
            $className = get_class($provider);
            throw new \InvalidArgumentException(
                "The {$className} Provider must implement PersistenceInterface in order to use impersonation.'
            );
```

src/AuthenticationService.php
```diff
@@ -428,4 +434,69 @@ public function getLoginRedirect(ServerRequestInterface $request): ?string
...
+     * @return array
+     */
+    public function stopImpersonating(ServerRequestInterface $request, ResponseInterface $response): array
+    {
+        /** @var \Authentication\Authenticator\PersistenceInterface $provider */
+        $provider = $this->getAuthenticationProvider();
+        if (!($provider instanceof PersistenceInterface)) {
+            throw new \InvalidArgumentException(
+                'Provider must implement PersistenceInterface to be able to impersonate user.'
+            );
```
markstory:
```suggestion
            $className = get_class($provider);
            throw new \InvalidArgumentException(
                "The {$className} Provider must implement PersistenceInterface in order to use impersonation.'
            );
```

src/AuthenticationService.php
```diff
@@ -428,4 +434,69 @@ public function getLoginRedirect(ServerRequestInterface $request): ?string
...
+     * @return bool
+     */
+    public function isImpersonating(ServerRequestInterface $request): bool
+    {
+        /** @var \Authentication\Authenticator\PersistenceInterface $provider */
+        $provider = $this->getAuthenticationProvider();
+        if (!($provider instanceof PersistenceInterface)) {
+            throw new \InvalidArgumentException(
+                'Provider must implement PersistenceInterface to be able to impersonate user.'
+            );
```
markstory:
```suggestion
            $className = get_class($provider);
            throw new \InvalidArgumentException(
                "The {$className} Provider must implement PersistenceInterface in order to use impersonation.'
            );
```

src/Authenticator/PersistenceInterface.php
```diff
@@ -41,4 +41,37 @@ public function persistIdentity(ServerRequestInterface $request, ResponseInterfa
...
+     */
+    public function stopImpersonating(ServerRequestInterface $request, ResponseInterface $response): array;
+
+    /**
+     * Returns true if impersonation is being done
+     *
+     * @param \Psr\Http\Message\ServerRequestInterface $request The request
+     * @return bool
+     */
+    public function isImpersonating(ServerRequestInterface $request): bool;
```
markstory: We will need a major release to extend the interfaces. We are likely going to have one of those soon those as there are a few other pull requests that require a major release.

src/Authenticator/SessionAuthenticator.php
```diff
@@ -115,4 +117,80 @@ public function clearIdentity(ServerRequestInterface $request, ResponseInterface
...
+        \ArrayAccess $impersonator,
+        \ArrayAccess $impersonated
+    ): array {
+        $sessionKey = $this->getConfig('sessionKey');
+        $impersonateSessionKey = $this->getConfig('impersonateSessionKey');
+        /** @var \Cake\Http\Session $session */
+        $session = $request->getAttribute('session');
+        if ($session->check($impersonateSessionKey)) {
+            throw new UnauthorizedException(
+                'You are impersonating a user already. Stops impersonation before doing another one.'
```
markstory:
```suggestion
                'You are impersonating a user already. Stop the current impersonation before impersonating another user.'
```

src/Authenticator/SessionAuthenticator.php
```diff
@@ -115,4 +117,80 @@ public function clearIdentity(ServerRequestInterface $request, ResponseInterface
...
+     * @return array
+     */
+    public function stopImpersonating(ServerRequestInterface $request, ResponseInterface $response): array
+    {
+        $sessionKey = $this->getConfig('sessionKey');
+        $impersonateSessionKey = $this->getConfig('impersonateSessionKey');
+        /** @var \Cake\Http\Session $session */
+        $session = $request->getAttribute('session');
+        if (!$session->check($impersonateSessionKey)) {
+            throw new UnauthorizedException('You are not impersonating a user.');
```
markstory: Should this be a silent error / treated as success instead? The developer wants impersonation to stop and if no impersonation is active then all impersonation is 'stopped'.

src/Authenticator/PersistenceInterface.php
ajibarra: Since we are not going to implement impersonation for CookieAuthenticator, I have moved this code out to another interface ImpersonationInterface. I have added it to SessionAuthenticator only.

ajibarra: I will be working on unit tests and please let me know if the target branch is ok or if we should make the MR to another branch. Thank you!

markstory: This is fine. We can wait to do the major until this is merged. If folks urgently need changes they can use dev-master or pin the the sha they need.

src/Authenticator/SessionAuthenticator.php
markstory: Should this be silent error? Won't raising an exception here cause the request to fail?

markstory (commented):
So how does one user an impersonated user once they have one? Are we going to support users 'being' the impersonated user and then 'reverting' their identity back to the old one? Will that be an `Authenticator` or are you thinking it should be a manual process?

src/Authenticator/SessionAuthenticator.php
ajibarra: it has been fixed in latest commit. I will mark it as resolved but feel free to reopen to discuss anything else about this.

```php
/** @var \Cake\Http\Session $session */
$session = $request->getAttribute('session');
if ($session->check($impersonateSessionKey)) {
    $identity = $session->read($impersonateSessionKey);
    $session->delete($impersonateSessionKey);
    $session->write($sessionKey, $identity);
    $this->setConfig('identify', true);
}
```

ajibarra:
> So how does one user an impersonated user once they have one? Are we going to support users 'being' the impersonated user and then 'reverting' their identity back to the old one? Will that be an `Authenticator` or are you thinking it should be a manual process?

If a user is already impersonating, then it throws an UnauthorizedException. We are going to support impersonation and also stops impersonation (not sure if this questions is related to multiple-levels impersonations. ATM we are going to support only single-level impersonation as previously discussed.

ajibarra:
> It would be helpful to know more about how you see developers using these features. Some use-cases that would be good to have examples of are:
> 
> 1. How does a developer enable account impersonation?
> 2. Once active does impersonation stay active until it is stopped?
> 3. How does one turn off impersonation?
> 
> We'll need these use-cases to write the docs, so having them now will help when it is time to write the documentation. Other questions, I have around impersonation are:
> 
> * Should there be a time limit that impersonation is active for? If we don't want to build that into this plugin how would someone go about building that?
> * Should we emit log messages when impersonation is activated, or should that be an application concern?

1. Impersonation is enabled by default and people just need to add a couple of actions like these to the UsersController (or similar):


```php
    public function impersonate($id = null)
    {
        $user = $this->Users->get($id);
        if ($this->Authentication->impersonate($user)) {
            $this->Flash->success(__('You are now impersonating {0}.', $user->email));
        } else {
            $this->Flash->error(__('The user could not be impersonated. Please, try again.'));
        }
        return $this->redirect(['action' => 'index']);
    }

    public function stopImpersonating()
    {
        $user = $this->Authentication->getIdentity();
        if ($this->Authentication->stopImpersonating()) {
            $this->Flash->success(__('You are not impersonating {0} anymore.', $user->email));
        } else {
            $this->Flash->error(__('An error has occurred leaving impersonation. Please, try again.'));
        }
        return $this->redirect(['action' => 'index']);
    }
```

2. Once active, impersonation stays active until session expires or user stops it.
3. To stop impersonation app must implement an action like previous one which calls AuthenticationComponent::stopImpersonating. If you mean turn off (disable feature) I think it is not needed since it depends exclusively from devs calling feature in their code. We are not providing a user the ability to impersonate if developer does not add it to the controller layer.

ajibarra:
@markstory I have merged your branch `fix-build` to be able to pass checks..Additionally I have fixed the php version missing in ci.yml

src/AuthenticationService.php
```diff
@@ -214,10 +215,14 @@ public function authenticate(ServerRequestInterface $request): ResultInterface
     public function clearIdentity(ServerRequestInterface $request, ResponseInterface $response): array
     {
         foreach ($this->authenticators() as $authenticator) {
-            if ($authenticator instanceof PersistenceInterface) {
+            if ($authenticator instanceof ImpersonationInterface) {
```
ADmad: Changing the interface check here means the identity won't be cleared for `CookieAuthenticator` which only implements `PersistenceInterface`, not `ImpersonationInterface`. I don't think that's expected.

src/AuthenticationService.php
```diff
@@ -31,7 +32,7 @@
 /**
  * Authentication Service
  */
-class AuthenticationService implements AuthenticationServiceInterface
+class AuthenticationService implements AuthenticationServiceInterface, ImpersonationInterface
```
ADmad, line 35: Implementing `ImpersonationInterface` is not necessary since `AuthenticationServiceInterface` already extends it.

src/AuthenticationService.php
```diff
@@ -428,4 +433,72 @@ public function getLoginRedirect(ServerRequestInterface $request): ?string
...
+    /**
+     * Returns true if impersonation is being done
+     *
+     * @param \Psr\Http\Message\ServerRequestInterface $request The request
+     * @return bool
+     */
+    public function isImpersonating(ServerRequestInterface $request): bool
+    {
+        /** @var \Authentication\Authenticator\ImpersonationInterface $provider */
+        $provider = $this->getAuthenticationProvider();
```
ADmad, line 493: The code on 494 - 500 is repeated thrice. They could be moved into a separate protected method `getImpersonationProvider()`.

src/AuthenticationServiceInterface.php
```diff
@@ -17,12 +17,12 @@
...
 
 use Authentication\Authenticator\AuthenticatorInterface;
-use Authentication\Authenticator\PersistenceInterface;
+use Authentication\Authenticator\ImpersonationInterface;
 use Authentication\Authenticator\ResultInterface;
 use Authentication\Identifier\IdentifierInterface;
 use Psr\Http\Message\ServerRequestInterface;
 
-interface AuthenticationServiceInterface extends PersistenceInterface
+interface AuthenticationServiceInterface extends ImpersonationInterface
```
ADmad: If `AuthenticationServiceInterface` extends `ImpersonationInterface` then perhaps it should be under the namespace `Authentication` instead of `Authentication\Authenticator`.

Also changing the interface is not backwards compatible and would require a major version bump. To avoid that this extension would have to be keep and have `AuthenticationService` directly implement `ImpersonationInterface` (which is already done currently).

markstory: :+1: to adding a new interface instead of modifying the existing one.

ajibarra: I have reverted extends and created a method to check in AuthenticationComponent for the AuthenticationService implementation of ImpersonationInterface. Added a test for it as well.

src/AuthenticationService.php
ajibarra: Fixed

markstory (commented):
Looking pretty good to me. I'll write some docs later this week.

PR #533, opened 2022-05-13
Add type hints to DefaultPasswordHasher.php

markusramsak:
add type hints like in DefaultPasswordHasher of AuthComponent in CakePHP 4.x


ADmad:
For the `PasswordHasherInterface` which this class implements these arguments are of type `string|array`, so `string` type can't be added for them.

PR #535, opened 2022-05-14
Support Cached key set

swiffer:
This is meant to gather your thoughts on support for cachedKeySets.

Instead of providing an array with keys to jwks and URL could be passed which is then fetched and cached by php-jwt.
Configuration options still TBD.

This functionality has been added to php-jwt in 6.2: https://github.com/firebase/php-jwt/pull/397

The revelant change would be among these lines: https://github.com/cakephp/authentication/pull/535/files#diff-ace0e7f81ad6600f2a672e7393458f22cc0f8c44e69bd76ef9be077e677c14fbR158

### DIFF
--- a/src/Authenticator/JwtAuthenticator.php
+++ b/src/Authenticator/JwtAuthenticator.php
@@ -18,7 +18,6 @@ namespace Authentication\Authenticator;
 
 use ArrayObject;
 use Authentication\Identifier\IdentifierInterface;
-use Cake\Utility\Hash;
 use Cake\Utility\Security;
 use Exception;
 use Firebase\JWT\JWK;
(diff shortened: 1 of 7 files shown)


othercorey:
This needs to be rebased.

swiffer:
> This needs to be rebased.

Yes, Happy to do that once #507 is merged / decided and 2.x-next Up to date or 3.x ist existing

othercorey:
Merged 2.x into 2.next.

swiffer:
done - it's still containing the commit of #507

src/Authenticator/JwtAuthenticator.php
```diff
@@ -18,15 +18,19 @@
...
 use Exception;
+use Firebase\JWT\CachedKeySet;
 use Firebase\JWT\JWK;
 use Firebase\JWT\JWT;
 use Firebase\JWT\Key;
+use Laminas\Diactoros\RequestFactory;
 use Psr\Http\Message\ServerRequestInterface;
 use RuntimeException;
 use stdClass;
+use Symfony\Component\Cache\Adapter\Psr16Adapter;
```
ADmad, line 33: Why do we need this Symfony adapter? Cake's cache engine are already PSR16 compliant.

swiffer: yes, php-jwt opted for PSR-6 based Cache. Psr16Adapter returns a PSR6 Cache.
See https://symfony.com/doc/current/components/cache/psr6_psr16_adapters.html#using-a-psr-16-cache-object-as-a-psr-6-cache

othercorey: Should we write our own adapter instead?

markstory: The cake cache adapters already implement the PSR16 interfaces. We could implement a 6 to 16 adapter in cake to avoid adding another medium sized dependency.

PR #538, opened 2022-05-27
PASETO authenticator and identifier

cnizzardini:
This is a rough draft implementation of PASETO (Platform Agnostic Security Tokens). This supports local symmetric keys and public asymmetric keys, but does not yet have support for keyrings and PASERK which gives functionality similar to JWKS. If there is interest in this I can add support for those items as well as unit tests and docs.

Usage:

Usage is fairly similar to the JWT Authenticator with the addition of a few keys:

```php
        $service->loadIdentifier('Authentication.PasetoSubject');
        $service->loadAuthenticator('Authentication.Paseto', [
            'purpose' => PasetoAuthenticator::LOCAL, // or PasetoAuthenticator::PUBLIC
        ]);
```

There is also a `version` which takes an implementation of `ProtocolInterface` but defaults to the latest version in paseto (currently v4). 

Using the public option is a little PITA to generate keys and I'd like to get a shell command added to the main paseto lib to easily generate the public/private key pairs.

Building a local token:

```php
        /** @var ResultInterface $result **/
        $key = SymmetricKey::v4(Security::getSalt());

        $token = (new Builder())
            ->setKey($key)
            ->setSubject($result->getData()->get('id'))
            ->setVersion(new Version4)
            ->setPurpose(Purpose::local())
            // Set it to expire in one day
            ->setIssuedAt()
            ->setNotBefore()
            ->setExpiration(
                (new DateTime())->add(new DateInterval('P01D'))
            )
            // Store arbitrary data
            ->setClaims([
                'claim_data' => 'is encrypted',
            ])
            ->setFooterArray([
                'footer_data' => 'is unencrypted but tamper proof'
            ]);
```

Reference:
- https://github.com/paragonie/paseto
- https://github.com/paseto-standard/paseto-spec


src/Authenticator/PasetoAuthenticator.php
```diff
@@ -0,0 +1,200 @@
...
+ * CakePHP(tm) : Rapid Development Framework (https://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ *
+ * Licensed under The MIT License
+ * For full copyright and license information, please see the LICENSE.txt
+ * Redistributions of files must retain the above copyright notice.
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ * @link          https://cakephp.org CakePHP(tm) Project
+ * @since         1.0.0
```
markstory: You'll likely need 3.0.0 here.

src/Authenticator/PasetoAuthenticator.php
```diff
@@ -0,0 +1,200 @@
...
+
+        if (!in_array($this->getConfig('purpose'), [self::PUBLIC, self::LOCAL])) {
+            throw new RuntimeException('PASETO `purpose` config must one of: local or public');
+        }
+
+        if (empty($this->getConfig('secretKey'))) {
+            if (!class_exists(\Cake\Utility\Security::class)) {
+                throw new RuntimeException('PASETO `secretKey` config must be defined');
+            }
+            $this->setConfig('secretKey', \Cake\Utility\Security::getSalt());
```
markstory, line 93: Will the application salt always be long enough? Are there length requirements in paseto?

src/Authenticator/PasetoAuthenticator.php
```diff
@@ -0,0 +1,200 @@
...
+        }
+
+        if (empty($result->getSubject())) {
+            return new Result(null, Result::FAILURE_CREDENTIALS_MISSING);
+        }
+
+        if ($this->getConfig('returnPayload')) {
+            $array = array_merge(
+                $result->getClaims(),
+                ['footer' => $result->getFooterArray()]
```
markstory, line 131: What is the footer used for by application logic?

src/Authenticator/PasetoAuthenticator.php
cnizzardini: For Symmetric keys between 16 and 64 bytes. sodium_crypto_generichash will throw a `SodiumException: unsupported key length` exception otherwise, unfortunately this is not well documented here: [sodium_crypto_generichash](https://www.php.net/manual/en/function.sodium-crypto-generichash.php). We could add a validation for this in the authenticator.

For Asymmetric keys it is recommended to use [sodium_crypto_sign_keypair](https://www.php.net/manual/en/function.sodium-crypto-sign-keypair.php) to generate the key pairs:

```php
$privateKey = new AsymmetricSecretKey(sodium_crypto_sign_keypair());
$publicKey = $privateKey->getPublicKey();
var_dump($privateKey->encode());
var_dump($publicKey->encode());
```

This is where I wish the library provided a cli to easily generate those for developers.

src/Authenticator/PasetoAuthenticator.php
cnizzardini: A token is split into version, purpose, encrypted data, and unencrypted data with dots as the separator.

```
v4.local.KQVqiaVJbg_V0W9s-ekQMYyjA5sZCSMUFw6WxLJ_6806uE7ceNad9ABgGwgJ_HDN8zLDC_SF9hzFoESX8lt-v_o1LjsahdoOETFlp884ztE-vz5MZq6CfvaVhH4cjT8Wi6p43H95jegzM3BhbjrJpAiUOWISPxbR5Da90WSkCUVSQfMjCzSrTjEQRdTlbBNpjzEOen6f8Vaw-Wl2eqtmzgVDZ5LDcyR9fatO_p_IhH8cFwLXILoORpjFBgeCDXZqr9jaum6zEvAs3XSfWB86RLUM_QD-elM_2DkaCwqS_1btNYKpahRkLpr6ehM8dSKL.eyJmb290ZXJfZGF0YSI6ImlzIHVuZW5jcnlwdGVkIGJ1dCB0YW1wZXIgcHJvb2YifQ
```

So you can base64 decode the footer segment: `eyJmb290ZXJfZGF0YSI6ImlzIHVuZW5jcnlwdGVkIGJ1dCB0YW1wZXIgcHJvb2YifQ` to: 

```
{"footer_data":"is unencrypted but tamper proof"}
```

I assume to give API clients or any client that doesn't have the secret the ability to read non-sensitive data.

src/Authenticator/PasetoAuthenticator.php
markstory: > This is where I wish the library provided a cli to easily generate those for developers.

We could provide that CLI tool :smile:

markstory: >For Symmetric keys between 16 and 64 bytes. sodium_crypto_generichash will throw a SodiumException: unsupported key length exception otherwise, unfortunately this is not well documented here

That will work. If users get tripped up by the exceptions from libsodium we can always catch errors and rethrow with a better message.

src/Authenticator/PasetoAuthenticator.php
```diff
@@ -0,0 +1,196 @@
...
+     */
+    protected $_defaultConfig = [
+        'header' => 'Authorization',
+        'queryParam' => 'token',
+        'tokenPrefix' => 'bearer',
+        'returnPayload' => true,
+        'version' => null,
+        'purpose' => null,
+        'secretKey' => null,
+        //'subjectKey' => IdentifierInterface::CREDENTIAL_JWT_SUBJECT,
```
cnizzardini: I'm thinking for this we will just remove subjectKey and always use `sub`.

src/Identifier/PasetoSubjectIdentifier.php
```diff
@@ -0,0 +1,37 @@
...
+class PasetoSubjectIdentifier extends TokenIdentifier
+{
+    /**
+     * Default configuration
+     *
+     * @var array
+     */
+    protected $_defaultConfig = [
+        'tokenField' => 'id',
+        'dataField' => self::CREDENTIAL_JWT_SUBJECT,
```
cnizzardini: Thinking of hardcoding dataField to `"sub"`

src/Authenticator/PasetoAuthenticator.php
```diff
@@ -0,0 +1,196 @@
...
+                'PASETO `version` config must be an instance of ' . ProtocolInterface::class
+            );
+        }
+
+        if (!in_array($this->getConfig('purpose'), [self::PUBLIC, self::LOCAL])) {
+            throw new RuntimeException('PASETO `purpose` config must one of: local or public');
+        }
+
+        if (empty($this->getConfig('secretKey'))) {
+            if (!class_exists(\Cake\Utility\Security::class)) {
```
cnizzardini, line 90: Don't know how to test this, thinking this should be marked as code coverage ignore.

src/Authenticator/PasetoAuthenticator.php
```diff
@@ -0,0 +1,196 @@
...
+    /**
+     * Get payload data.
+     *
+     * @param \Psr\Http\Message\ServerRequestInterface|null $request Request to get authentication information from.
+     * @return object|null Payload object on success, null on failure
+     * @throws \Exception
+     */
+    public function getPayload(?ServerRequestInterface $request = null): ?object
+    {
+        if (!$request) {
```
cnizzardini, line 157: Not sure I understand this condition, it was borrowed from JwtAuthenticator. How is this scenario possible?

composer.json
```diff
@@ -21,12 +21,15 @@
         "cakephp/cakephp": "^4.0",
         "cakephp/cakephp-codesniffer": "^4.0",
         "firebase/php-jwt": "^6.2",
+        "paragonie/paseto": "^2.4",
+        "paragonie/constant_time_encoding": "^2.2",
```
cnizzardini, line 25: This was required because of prefer-lowest on 7.2. The method needed from the package was not included until 2.2 (current is 2.5).

othercorey:
@cnizzardini After fixing the phpstan error, can you squash the commits?

src/Authenticator/PasetoAuthenticator.php
```diff
@@ -0,0 +1,219 @@
...
+
+        $this->version = $this->whichVersion();
+
+        if ($this->version === null) {
+            throw new RuntimeException('PASETO `version` must be one of: v3 or v4');
+        }
+
+        if (!in_array($this->getConfig('purpose'), [self::PUBLIC, self::LOCAL])) {
+            throw new RuntimeException('PASETO `purpose` config must one of: local or public');
+        }
```
markstory, line 87: Are there no reasonably secure defaults we could use? Even 'most secure' is a reasonable defaults. Right now we're forcing a lot of decisions on the developer.

src/Authenticator/PasetoAuthenticator.php
markstory: The behavior looks like it wants to reuse the stored `payload` property. By passing `null` in, you get the last parsed result.

src/Authenticator/PasetoAuthenticator.php
```diff
@@ -0,0 +1,219 @@
...
+     * Decode PASETO token.
+     *
+     * @param string $token PASETO token to decode.
+     * @return null|\ParagonIE\Paseto\JsonToken The PASETO payload as a JsonToken object, null on failure.
+     * @throws \Exception
+     */
+    protected function decodeToken(string $token): ?JsonToken
+    {
+        if ($this->getConfig('purpose') === self::PUBLIC) {
+            $receivingKey = AsymmetricSecretKey::fromEncodedString(
```
markstory, line 183: This seems like a reasonable default `purpose` to me.

src/Authenticator/PasetoAuthenticator.php
```diff
@@ -0,0 +1,219 @@
...
+     * Returns instance of ProtocolInterface from configured version.
+     *
+     * @return \ParagonIE\Paseto\ProtocolInterface|null
+     */
+    private function whichVersion(): ?ProtocolInterface
+    {
+        switch ($this->getConfig('version')) {
+            case 'v3':
+                return new Version3();
+            case 'v4':
```
markstory, line 213: Defaulting to the highest version seems like a good default value as well.

src/Command/PasetoCommand.php
```diff
@@ -0,0 +1,96 @@
...
+            ->setDescription('Generate keys for PASETO')
+            ->addArgument('version', [
+                'help' => 'The PASETO version',
+                'required' => true,
+                'choices' => ['v3', 'v4'],
+            ])
+            ->addArgument('purpose', [
+                'help' => 'The PASETO purpose',
+                'required' => true,
+                'choices' => [PasetoAuthenticator::LOCAL, PasetoAuthenticator::PUBLIC],
```
markstory, line 46: Would be good to have a default value on these too. Good defaults help users succeed with less effort.

tests/TestCase/Authenticator/PasetoAuthenticatorTest.php
```diff
@@ -0,0 +1,553 @@
...
+    public function testAuthenticationViaQueryParamToken(string $purpose): void
+    {
+        switch ($purpose) {
+            case PasetoAuthenticator::LOCAL:
+                $token = $this->buildLocalToken(new Version4());
+                break;
+            case PasetoAuthenticator::PUBLIC:
+                /** @var Builder $token extracted */
+                /** @var AsymmetricSecretKey $privateKey extracted */
+                extract($this->buildPublicToken(new Version4()));
```
markstory: This feels risky to me. Having a method arbitrarily update local scope has not worked out well for us in the past. Can we be more explicit with what is being set in local scope here.

tests/TestCase/Authenticator/PasetoAuthenticatorTest.php
```diff
@@ -0,0 +1,553 @@
...
+                ],
+            ])
+            ->onlyMethods([
+                'getPayLoad',
+            ])
+            ->getMock();
+
+        $authenticator->expects($this->once())
+            ->method('getPayLoad')
+            ->willReturn((new JsonToken())->setSubject(''));
```
markstory: Does this need to use a mock? There is a lot of code being dedicated to mocking to avoid creating a value we could generate with fewer lines of code.

Could we try having a helper method that generates a key pair and creates an authenticator using that key pair. You would also need a method that generates a token for a given input.

This is a bit of plumbing work to do but in the long term it helps with maintenance as we don't have to worry about changes in a mocking library. Additionally it also gives us more confidence in the tests as we are testing more of the subsytem all together.

src/Authenticator/PasetoAuthenticator.php
cnizzardini: We can default to local and v4.

src/Command/PasetoCommand.php
cnizzardini: We can set defaults to v4 local here as well.

othercorey:
@cnizzardini Can you follow up on this?

cnizzardini:
Its on my mind, some personal items getting in the way of freetime dev
right now. Gimme a few days.

src/Authenticator/PasetoAuthenticator.php
cnizzardini: I was thinking LOCAL should be the default purpose. Local is similar to the default for JWT which uses an HMAC secret.

othercorey:
@markstory can you take another look?

src/Authenticator/PasetoAuthenticator.php
```diff
@@ -0,0 +1,219 @@
...
+ * CakePHP(tm) : Rapid Development Framework (https://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ *
+ * Licensed under The MIT License
+ * For full copyright and license information, please see the LICENSE.txt
+ * Redistributions of files must retain the above copyright notice.
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ * @link          https://cakephp.org CakePHP(tm) Project
+ * @since         3.0.0
```
ADmad, line 14:
```suggestion
 * @since         2.10.0
```

src/Authenticator/PasetoAuthenticator.php
```diff
@@ -0,0 +1,219 @@
...
+    /**
+     * @inheritDoc
+     */
+    protected $_defaultConfig = [
+        'header' => 'Authorization',
+        'queryParam' => 'token',
+        'tokenPrefix' => 'bearer',
+        'returnPayload' => true,
+        'version' => 'v4',
+        'purpose' => 'local',
```
ADmad, line 56:
```suggestion
        'purpose' => self::LOCAL,
```

src/Authenticator/PasetoAuthenticator.php
```diff
@@ -0,0 +1,219 @@
...
+     * Payload data.
+     *
+     * @var object|null
+     */
+    protected $payload;
+
+    /**
+     * @var \ParagonIE\Paseto\ProtocolInterface|null
+     */
+    private $version;
```
ADmad, line 70:
```suggestion
    protected $version;
```

src/Authenticator/PasetoAuthenticator.php
```diff
@@ -0,0 +1,219 @@
...
+    /**
+     * @inheritDoc
+     */
+    public function __construct(IdentifierInterface $identifier, array $config = [])
+    {
+        parent::__construct($identifier, $config);
+
+        $this->version = $this->whichVersion();
+
+        if ($this->version === null) {
```
ADmad, line 81: The `whichVersion()` method itself can throw an exception instead of returning null.

PR #543, opened 2022-06-07
Remove redundant logic from login action example

nishimura-d:
Reaches this line only when result is not valid.


markstory:
Thank you 🎉

PR #549, opened 2022-06-19
Add native type declarations

othercorey:


src/Authenticator/AuthenticatorCollection.php
```diff
@@ -49,19 +49,21 @@ public function __construct(IdentifierCollection $identifiers, array $config = [
...
      *
-     * @param string $class Authenticator class.
+     * @param object|string $class Authenticator class.
      * @param string $alias Authenticator alias.
      * @param array $config Config array.
      * @return \Authentication\Authenticator\AuthenticatorInterface
      * @throws \RuntimeException
      */
-    protected function _create($class, string $alias, array $config): AuthenticatorInterface
+    protected function _create(object|string $class, string $alias, array $config): AuthenticatorInterface
```
othercorey, line 58: We must support `object|string` to implement ObjectRegistry.

src/Identifier/IdentifierCollection.php
```diff
@@ -66,19 +67,21 @@ public function identify(array $credentials)
...
      *
-     * @param string $class Identifier class.
+     * @param object|string $class Identifier class.
      * @param string $alias Identifier alias.
      * @param array $config Config array.
      * @return \Authentication\Identifier\IdentifierInterface
      * @throws \RuntimeException
      */
-    protected function _create($class, string $alias, array $config): IdentifierInterface
+    protected function _create(object|string $class, string $alias, array $config): IdentifierInterface
```
othercorey, line 76: We must support `object|string` to implement ObjectRegistry.

src/PasswordHasher/DefaultPasswordHasher.php
```diff
@@ -40,10 +40,10 @@ class DefaultPasswordHasher extends AbstractPasswordHasher
     /**
      * Generates password hash.
      *
-     * @param string $password Plain text password to hash.
+     * @param array|string $password Plain text password to hash.
      * @return string Password hash or false on failure.
      */
-    public function hash($password): string
+    public function hash(array|string $password): string
```
othercorey: PasswordHasherInterface requires `array|string` for password, but none of the hashers support arrays.

ndm2: That doesn't sound right. Where exactly does the identifier pass an array for the password?

othercorey: It was just part of the @param annotation that was converted into a type declaration.

ndm2: Sorry, I still don't see it. Which `@param`?

othercorey: https://github.com/cakephp/authentication/blob/cake5/src/PasswordHasher/PasswordHasherInterface.php#L26

ADmad: I can't remember either why it's `string|array`. Maybe to support some edge case custom hashing is my guess.

ndm2: Ah OK, you meant the interface, I thought the typo was meant to be `PasswordIdentifier`.

ndm2: Maybe this is a chance to drop the array? With parameter type widening, one could implement a `string` only param as `array|string`, or even drop the type hint completely if needed.

The password identifier doesn't support passing arrays, so a custom identifier would be needed anyways if people wanted to pass arrays to the hasher, so they could then also "cast" the used hasher accordingly.

othercorey: So we want to drop `array`? That would fix the mismatch between needsRehash() and hash().

ndm2: I don't really see a mismatch there, as `needsRehash()` operates on the hashed password (maybe the argument should be renamed to reflect that better). What makes me feel uneasy is forcing all the hashers to accept an array when they don't support it, and will fail with type errors accordingly 🤷

othercorey: ah, you're right, it takes the hashed password. I must have been looking at a different password_* parameter.

ADmad (approved):

markstory (approved):
Looks good to me.

src/PasswordHasher/PasswordHasherInterface.php
```diff
@@ -23,21 +23,21 @@ interface PasswordHasherInterface
     /**
      * Generates password hash.
      *
-     * @param string|array $password Plain text password to hash or array of data
+     * @param string $password Plain text password to hash or array of data
      *   required to generate password hash.
```
ndm2:
```suggestion
     * @param string $password Plain text password to hash.
```

src/PasswordHasher/PasswordHasherInterface.php
```diff
@@ -23,21 +23,21 @@ interface PasswordHasherInterface
...
      */
-    public function hash($password): string;
+    public function hash(string $password): string;
 
     /**
      * Check hash. Generate hash from user provided password string or data array
      * and check against existing hash.
      *
-     * @param string|array $password Plain text password to hash or data array.
+     * @param string $password Plain text password to hash or data array.
```
ndm2:
```suggestion
     * @param string $password Plain text password to hash.
```

PR #550, opened 2022-06-23
Remove unnecessary exception on ObjectRegistry return type

othercorey:
We rely on the return type declaration to validate the object.


othercorey:
Unfortunately, psalm doesn't deduce the object type created by new after an `is_string()` check with a class-string.

src/Authenticator/AuthenticatorCollection.php
```diff
@@ -49,26 +49,21 @@ public function __construct(IdentifierCollection $identifiers, array $config = [
...
-                'Authenticator must implement `%s`.',
-                AuthenticatorInterface::class
-            ));
+            /** @var class-string<\Authentication\Authenticator\AuthenticatorInterface> $class */
+            return new $class($this->_identifiers, $config);
         }
 
-        return $authenticator;
+        /** @var \Authentication\Authenticator\AuthenticatorInterface $class */
+        return $class;
```
othercorey, line 66: psalm uses the `object|string` inferred type instead of the annotated type so have to specify $class as \Authentication\Authenticator\AuthenticatorInterface

ADmad (approved):

PR #559, opened 2022-08-02
Token Identifier: Add Option to Hash Token Before Resolving

amayer5125:
With this change developers can hash values stored in the database and specify a `hashAlgorithm` to apply before forwarding the value to the resolver.

I retained the default behavior of no hash, but we should probably apply some hash by default for security reasons. Then users can disable the hash, but it is secure by default.


markstory:
Thank you 🎉

PR #560, opened 2022-09-02
Add forceLowercase option to PasswordIdentifier

othercorey:
This allows case-insensitive usernames when the username is always forced to lowercase when storing in the db.

This can reduce the dependency on the UI to always send case-insensitive strings to the server.

### DIFF
--- a/src/Identifier/PasswordIdentifier.php
+++ b/src/Identifier/PasswordIdentifier.php
@@ -51,6 +51,8 @@ class PasswordIdentifier extends AbstractIdentifier
      * - `fields` The fields to use to identify a user by:
      *   - `username`: one or many username fields.
      *   - `password`: password field.
+     * - `forceLowercase` Forces identifier to lowercase when comparing to
+     *   `username` fields. Allows for case-insensitive comparison.
      * - `resolver` The resolver implementation to use.
      * - `passwordHasher` Password hasher class. Can be a string specifying class name
      *    or an array containing `className` key, any other keys will be passed as
@@ -63,6 +65,7 @@ class PasswordIdentifier extends AbstractIdentifier
             self::CREDENTIAL_USERNAME => 'username',
             self::CREDENTIAL_PASSWORD => 'password',
         ],
+        'forceLowercase' => false,
         'resolver' => 'Authentication.Orm',
         'passwordHasher' => null,
     ];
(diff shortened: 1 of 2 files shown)


ADmad:
Lowercasing in PHp would be not needed if case insensitive db field is used. I would rather have users do that instead.

othercorey:
> Lowercasing in PHp would be not needed if case insensitive db field is used. I would rather have users do that instead.

This isn't a supported data type in postgres unfortunately. There are ways to do case-insensitive string compares, but I don't think it's standard.

othercorey:
The alternative is a custom resolver just for usernames which is what I use.

ADmad:
Wouldn't you have to lowercase case the db value too to ensure a match? Lowercasing the user provided value won't help if the value stored in db is not in lowercase (for Postgres).

src/Identifier/PasswordIdentifier.php
```diff
@@ -51,6 +51,8 @@ class PasswordIdentifier extends AbstractIdentifier
      * - `fields` The fields to use to identify a user by:
      *   - `username`: one or many username fields.
      *   - `password`: password field.
+     * - `forceLowercase` Forces identifier to lowercase when comparing to
```
markstory, line 54: This name feels scary and I wouldn't use it. Could we do `forceLowercaseUsername` or `lowercaseUsername` instead?

markstory:
> Wouldn't you have to lowercase case the db value too to ensure a match? Lowercasing the user provided value won't help if the value stored in db is not in lowercase (for Postgres).

Yes, you would need to have entity setters or save logic to lowercase the fields.

src/Identifier/PasswordIdentifier.php
ADmad: My thoughts exactly, `lowercaseUsername` sounds much better.

ADmad:
> Yes, you would need to have entity setters or save logic to lowercase the fields.

I would hate it if an app/site did that :). For e.g. my GH username is mixed cased and that's what's shown/works everywhere (I haven't set any display name). But I can still login using lower cased version of my username.

ADmad:
So I think the db should store the exact string user provided during account creation and the lowercasing of db value too should be done in query. So perhaps this option belongs in the Orm resolver?

othercorey:
Ok, will stick with a specialized solution.

othercorey:
> So I think the db should store the exact string user provided during account creation and the lowercasing of db value too should be done in query. So perhaps this option belongs in the Orm resolver?

Yes, that's the optimal solution. I use a lowercase index to ensure unique case-insensitive email addresses.

I don't run into the display name issue you mentioned because `username` is email, but it's a good point.

dereuromark:
Even Email some people prefer different casing for readability :) so should usually also not get lowercased maybe.

othercorey:
Well, it's a non-visible config for me.

PR #563, opened 2022-10-08
cake5 updates

LordSimal:


psalm-baseline.xml
```diff
@@ -1,11 +1,32 @@
...
+      <code>\Authentication\Authenticator\AuthenticatorInterface|null</code>
+    </ImplementedReturnTypeMismatch>
     <MoreSpecificImplementedParamType occurrences="1">
       <code>$class</code>
     </MoreSpecificImplementedParamType>
   </file>
   <file src="src/Identifier/IdentifierCollection.php">
+    <ImplementedReturnTypeMismatch occurrences="1">
+      <code>\Authentication\Identifier\IdentifierInterface|null</code>
+    </ImplementedReturnTypeMismatch>
```
othercorey, line 29: What is causing the mismatch?

LordSimal: The error is:
```
ERROR: ImplementedReturnTypeMismatch - src/Identifier/IdentifierCollection.php:25:12 - The inherited return type 'object' for Cake\Core\ObjectRegistry::get is different to the implemented return type for Cake\Core\ObjectRegistry::get 'Authentication\Identifier\IdentifierInterface|null' 
```
because `ObjectRegistry::get()` returns a generic object

ADmad:
Please update GA CI config too to use the common workflows as I have done for other core plugins.

LordSimal:
@ADmad the `cake5` branch of cakephp/authentication already uses the reusable workflows

ADmad:
Oh my bad.

markstory (approved):
Looks good to me.

psalm-baseline.xml
ADmad: Adding proper annotation like [this](https://github.com/cakephp/cakephp/blob/5.x/src/Controller/ComponentRegistry.php#L30) will make static analyzers understand the specific type returned.

LordSimal: Tried that but wasn't sucessfull because here we are using an `AbstractCollection` in between (and I am not able to find a combination with whom the static analysis gods are happy with)

ADmad: I'll see if I can please the stan gods.

PR #564, opened 2022-10-10
Fix errors repoted by static analyzers.

ADmad:


psalm.xml
```diff
@@ -3,7 +3,6 @@
     xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
     xmlns="https://getpsalm.org/schema/config"
     xsi:schemaLocation="https://getpsalm.org/schema/config vendor/vimeo/psalm/config.xsd"
-    errorBaseline="psalm-baseline.xml"
```
othercorey, line 6: Should we not keep this but leave it empty?

ADmad: Don't see the point in keeping an empty file, we can regenerate it when needed.

dereuromark: It is a bit annoying that psalm doesnt look for a default file/path, like other tools do.
Verbose should only be needed for custom names or paths IMO

PR #565, opened 2022-10-11
EnvironmentAuthenticator

MolbioUnige:
Hi,
My institute uses [Shibboleth](https://shibboleth.atlassian.net/wiki/spaces/CONCEPT/overview) as SSO. After identification by the identity provider, the credentials are injected as environment variables by the http server. This authenticator is a TokenAuthenticator but:

- for server variables,  
- not limited to one token, 
- and checks url.

I don't know if there are other SSO technologies outside Shibboleth that uses environment variable to pass credentials, and have no idea how spread Shibboleth is.

Usage

```
  1         $service->loadAuthenticator('Authentication.Environment', [
  2             'loginUrl' => $service->getConfig('unauthenticatedRedirect'),
  3             'attributes' => [
  4                 'ENV_VAR_1' => 'app_var_1',
  5                 'ENV_VAR_2' => 'app_var_2',
  6             ],
  7         ]);
```

### DIFF
--- a/src/Authenticator/EnvironmentAuthenticator.php
+++ b/src/Authenticator/EnvironmentAuthenticator.php
@@ -33,7 +33,7 @@ class EnvironmentAuthenticator extends AbstractAuthenticator
      * - `loginUrl` Login URL or an array of URLs.
      * - `urlChecker` Url checker config.
      * - `fields` array of required fields to get from the environment
-     * - `optional_fields` array of optional fields to get from the environment
+     * - `optionalFields` array of optional fields to get from the environment
      *
      * @var array
      */
@@ -41,7 +41,7 @@ class EnvironmentAuthenticator extends AbstractAuthenticator
         'loginUrl' => null,
         'urlChecker' => 'Authentication.Default',
         'fields' => [],
-        'optional_fields' => [],
+        'optionalFields' => [],
     ];
 
     /**
@@ -73,14 +73,14 @@ class EnvironmentAuthenticator extends AbstractAuthenticator
     }
 
     /**
-     * Get values from the environment variables configured by `optional_fields`.
+     * Get values from the environment variables configured by `optionalFields`.
      *
      * @param \Psr\Http\Message\ServerRequestInterface $request The request that contains login information.
-     * @return array server params defined by optional_fields.
+     * @return array server params defined by optionalFields.
      */
     protected function _getOptionalData(ServerRequestInterface $request): array
     {
-        $fields = $this->_config['optional_fields'];
+        $fields = $this->_config['optionalFields'];
         $params = $request->getServerParams();
 
         $data = [];
--- a/src/Authenticator/EnvironmentAuthenticator.php
+++ b/src/Authenticator/EnvironmentAuthenticator.php
@@ -1,0 +1,156 @@
+<?php
+declare(strict_types=1);
+
+/**
+ * CakePHP(tm) : Rapid Development Framework (https://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ *
+ * Licensed under The MIT License
+ * For full copyright and license information, please see the LICENSE.txt
+ * Redistributions of files must retain the above copyright notice.
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ * @link          https://cakephp.org CakePHP(tm) Project
+ * @since         2.10.0
+ * @license       https://opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Authentication\Authenticator;
+
+use Authentication\UrlChecker\UrlCheckerTrait;
+use Psr\Http\Message\ServerRequestInterface;
+
+/**
+ * Environment Authenticator
+ *
+ * Authenticates an identity based on the POST data of the request.
+ */
+class EnvironmentAuthenticator extends AbstractAuthenticator
+{
+    use UrlCheckerTrait;
+
+    /**
+     * Default config for this object.
+     * - `loginUrl` Login URL or an array of URLs.
+     * - `urlChecker` Url checker config.
+     * - `fields` array of required fields to get from the environment
+     * - `optional_fields` array of optional fields to get from the environment
+     *
+     * @var array
+     */
+    protected $_defaultConfig = [
+        'loginUrl' => null,
+        'urlChecker' => 'Authentication.Default',
+        'fields' => [],
+        'optional_fields' => [],
+    ];
+
+    /**
+     * Get values from the environment variables configured by `fields`.
+     *
+     * @param \Psr\Http\Message\ServerRequestInterface $request The request that contains login information.
+     * @return array|null server params defined by `fields` or null if a field is missing.
+     */
+    protected function _getData(ServerRequestInterface $request): ?array
+    {
+        $fields = $this->_config['fields'];
+        $params = $request->getServerParams();
+
+        $data = [];
+        foreach ($fields as $field) {
+            if (!isset($params[$field])) {
+                return null;
+            }
+
+            $value = $params[$field];
+            if (!is_string($value) || !strlen($value)) {
+                return null;
+            }
+
+            $data[$field] = $value;
+        }
+
+        return $data;
+    }
+
+    /**
+     * Get values from the environment variables configured by `optional_fields`.
+     *
+     * @param \Psr\Http\Message\ServerRequestInterface $request The request that contains login information.
+     * @return array server params defined by optional_fields.
+     */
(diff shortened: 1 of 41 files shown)


src/Authenticator/EnvironmentAuthenticator.php
```diff
@@ -0,0 +1,127 @@
...
+     * - `fields` The fields to use to identify a user by.
+     * - `loginUrl` Login URL or an array of URLs.
+     * - `urlChecker` Url checker config.
+     *
+     * @var array
+     */
+    protected $_defaultConfig = [
+        'loginUrl' => null,
+        'urlChecker' => 'Authentication.Default',
+        'attributes' => [],
```
markstory: Why is this `attributes` and not `fields`?

src/Authenticator/EnvironmentAuthenticator.php
```diff
@@ -0,0 +1,127 @@
...
+     * @param \Psr\Http\Message\ServerRequestInterface $request The request that contains login information.
+     * @return \Authentication\Authenticator\ResultInterface
+     */
+    public function authenticate(ServerRequestInterface $request): ResultInterface
+    {
+        if (!$this->_checkUrl($request)) {
+            return $this->_buildLoginUrlErrorResult($request);
+        }
+        $data = $this->_getData($request);
+        if ($data === null || empty($data)) {
```
markstory: If you're using `empty()` you don't need to compare with `=== null` as empty() covers that.

src/Authenticator/EnvironmentAuthenticator.php
MolbioUnige: I've hesitated about the term. I chose `attributes` to stay close to the `Shibboleth` documentation. When you setup `Shibboleth` you choose what attributes you want, but you're right, it might be confusing.

src/Authenticator/EnvironmentAuthenticator.php
MolbioUnige: my bad :(

src/Authenticator/EnvironmentAuthenticator.php
MolbioUnige: Now, you made me think that what I did is not necessary. The `attributes` parameter (or renamed as `fields`)  should contain a list of environment variable names one would like to grab, without mapping names as well. Better have identifiers map values no?

MolbioUnige:
I revised the code a bit. Should I make a new commit or should I squash it to the previous one?
Sorry for asking, it's the first time a make a pull request.

ndm2:
Please squash, having rather small changes scattered over multiple commits is somewhat annoying when browsing the history and trying to understand changes. A more relevant commit message would also be nice 😉

src/Authenticator/EnvironmentAuthenticator.php
markstory: I'm not well versed with Shibboleth. But we'll you need a way to map the environment variables to the fields in your database schema that identify users.

For other parts of authentication we use `fields` to connect the form fields/request data to the username/password/token credential fields. Would something like 

```php
'fields' => [IdentifierInterface::CREDENTIAL_TOKEN => 'SHIBBOLETH_ID']
```
work? My question about `attributes` was trying to see if we need more than just the standard identifier/authenticator fields.

markstory:
> I revised the code a bit. Should I make a new commit or should I squash it to the previous one?

We can also squash on our side when merging :smile:

src/Authenticator/EnvironmentAuthenticator.php
MolbioUnige: The IdentityProvider (Shibboleth) provides [different user informations](https://en.wikipedia.org/wiki/Shibboleth_(software)#Attributes), nothing standardized in terms of names here. Some of the attributes can be used for identification, some for authorization.
I could grant access to my app only if a user has certain attributes. All swiss universities are using the same SSO. I could grant access to an identified user only if this user is member of my institute, or only if a user is member of certain universities, or only if it belongs to a certain department, or whatever. 

I have use case where a person must be identified the first time by it's email, and then by it's unique_id.
This application manages some of our students curriculum. Only certain students can access it. There is no registration form. Students are added to the database by a supervisor, and are identified the first time they connect by their private email, as it is not possible to know their SHIBBOLETH_ID before they connect for the first time.

Multiple emails can be linked to a single SHIBBOLETH_ID. The user can thus identify itself successfully to the IdentityProvider using different emails but would only be able to access my app with one of them, which is not satisfactory. I must save the SHIBBOLETH_ID after their first connection, and use that field for subsequent identification. So at one point, the Authenticator must pass more then one field to the identifier.

So depending the use case, `fields` could be restricted to only one token, but, in my cases, it would only be exceptional.

src/Authenticator/EnvironmentAuthenticator.php
```diff
@@ -0,0 +1,127 @@
...
+ * @since         2.10.0
+ * @license       https://opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Authentication\Authenticator;
+
+use Authentication\UrlChecker\UrlCheckerTrait;
+use Psr\Http\Message\ServerRequestInterface;
+
+/**
+ * Form Authenticator
```
ADmad:
```suggestion
 * Environment Authenticator
```

src/Authenticator/EnvironmentAuthenticator.php
```diff
@@ -0,0 +1,127 @@
...
+ *
+ * Authenticates an identity based on the POST data of the request.
+ */
+class EnvironmentAuthenticator extends AbstractAuthenticator
+{
+    use UrlCheckerTrait;
+
+    /**
+     * Default config for this object.
+     * - `fields` The fields to use to identify a user by.
```
ADmad:
```suggestion
     * - `attributes` <appropriate description here>
```

src/Authenticator/EnvironmentAuthenticator.php
```diff
@@ -0,0 +1,127 @@
...
+    /**
+     * Checks the fields to ensure they are supplied.
+     *
+     * @param \Psr\Http\Message\ServerRequestInterface $request The request that contains login information.
+     * @return array|null server params defined by fields.
+     */
+    protected function _getData(ServerRequestInterface $request): ?array
+    {
+        $attributes = $this->_config['attributes'];
+        /** @var array $params */
```
ADmad: Is it annotation really necessary? Return value of `getServerParams()` is always an array.

MolbioUnige:
Sorry I  made a mess with my commits

MolbioUnige:
I don't understand where the CI errors come from. I don't see anything if I click on `details`

MolbioUnige:
Should I start again from scratch? The PR I mean

ndm2:
It wasn't your fault:

> This is a scheduled Ubuntu-18.04 brownout. The Ubuntu-18.04 environment is deprecated and will be removed on April 1st, 2023. For more details, see https://github.com/actions/runner-images/issues/6002

src/Authenticator/EnvironmentAuthenticator.php
```diff
@@ -0,0 +1,156 @@
...
+     * - `fields` array of required fields to get from the environment
+     * - `optional_fields` array of optional fields to get from the environment
+     *
+     * @var array
+     */
+    protected $_defaultConfig = [
+        'loginUrl' => null,
+        'urlChecker' => 'Authentication.Default',
+        'fields' => [],
+        'optional_fields' => [],
```
markstory:
```suggestion
        'optionalFields' => [],
```

src/Authenticator/EnvironmentAuthenticator.php
```diff
@@ -0,0 +1,156 @@
...
+class EnvironmentAuthenticator extends AbstractAuthenticator
+{
+    use UrlCheckerTrait;
+
+    /**
+     * Default config for this object.
+     * - `loginUrl` Login URL or an array of URLs.
+     * - `urlChecker` Url checker config.
+     * - `fields` array of required fields to get from the environment
+     * - `optional_fields` array of optional fields to get from the environment
```
markstory:
```suggestion
     * - `optionalFields` array of optional fields to get from the environment
```

src/Authenticator/EnvironmentAuthenticator.php
```diff
@@ -0,0 +1,156 @@
...
+
+    /**
+     * Get values from the environment variables configured by `optional_fields`.
+     *
+     * @param \Psr\Http\Message\ServerRequestInterface $request The request that contains login information.
+     * @return array server params defined by optional_fields.
+     */
+    protected function _getOptionalData(ServerRequestInterface $request): array
+    {
+        $fields = $this->_config['optional_fields'];
```
markstory:
```suggestion
        $fields = $this->_config['optionalFields'];
```

PR #598, opened 2022-12-21
3.x: fix Uri class usage

LordSimal:
`Http\Uri` was dropped in https://github.com/cakephp/cakephp/pull/16859

But now since we don't have a wrapper class holding the `base` info the `DefaultUrlChecker` needed to be updated and use the config values instead. I don't know if this is the correct way right now but it was the first that came to my mind.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -74,6 +74,7 @@
         "test": "phpunit",
         "test-coverage": "phpunit --coverage-clover=clover.xml"
     },
+    "minimum-stability": "dev",
     "config": {
         "sort-packages": true,
         "allow-plugins": {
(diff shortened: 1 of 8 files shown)


ADmad:
`ServerRequest` has the `base` as an attribute so instead of the Uri we can pass the request instance to the `_getUrlFromRequest()` method and check if `$request->getAttribute('base')` is non-empty.

LordSimal:
So am I correct that the `base` attribute on the request has priority over the `App.base` config value?

ndm2:
I don't think there should be a fallback, if the server request object has no base, then either the app has none, or the request object was created manually (not via the factory) or has the attribute unset, both of which indicates possible intent.

ADmad:
The request's base attribute is set based on the App.base config, there's no need to check for it again and use it as fallback.

PR #600, opened 2022-12-28
attempt to implement support for cached keyset

swiffer:


composer.json
```diff
@@ -21,7 +21,8 @@
         "cakephp/cakephp": "^4.4",
         "cakephp/cakephp-codesniffer": "^4.0",
         "firebase/php-jwt": "^6.2",
-        "phpunit/phpunit": "^8.5 || ^9.3"
+        "phpunit/phpunit": "^8.5 || ^9.3",
+        "symfony/cache": "^5.4 || ^6.0"
```
markstory, line 25: Why do we need this dependency?

swiffer: see the discussion here:

https://github.com/cakephp/authentication/pull/535#discussion_r872972048

this was rebased / reopened as github closed it beeing stale for a while.

markstory: OK, so its the PSR6 vs PSR16 compatibility shims that we actually need. While I don't love the additional dependency I also don't relish maintaining adapter code for a low-frequency use feature. We'll should include the additional dependencies required in the documentation and raise an error with instructions to install the required dependencies if they don't exist.

src/Authenticator/JwtAuthenticator.php
```diff
@@ -149,7 +155,18 @@ protected function decodeToken(string $token): ?object
...
         $jsonWebKeySet = $this->getConfig('jwks');
         if ($jsonWebKeySet) {
-            $keySet = JWK::parseKeySet($jsonWebKeySet);
+            $jsonWebKeySetCache = $this->getConfig('jwksCache');
+            if ($jsonWebKeySetCache) {
+                $keySet = new CachedKeySet(
+                    $jsonWebKeySet,
+                    new Client(),
+                    new RequestFactory(),
+                    new Psr16Adapter(Cache::pool($jsonWebKeySetCache)),
```
ADmad, line 164: Given that `symfony/cache` is an optional dependency the existence of the `Psr16Adapter` class should be checked first and in case of failure show a message stating which package the dev should include in their app.

PR #608, opened 2023-03-15
Store only the original data in the impersonation session

### LINKED ISSUE #606: Impersonate issue with serialization for session
The method AuthenticationComponent::impersonate results with the error `Serialization of 'Closure' is not allowed`. 

This error is fixed by calling the method `getOriginalData` : `$identity = $this->getIdentity()->getOriginalData();`

Error trace:
```
session_write_close
CORE/src/Http/Session.php:381
Cake\Http\Session->close
CORE/src/Http/Server.php:93
Cake\Http\Server->run
ROOT/webroot/index.php:40
```

markstory:
Identity objects could contain closures by fetching the original data we can serialize into the session more consistently. It also ensures that when we next build an identity that a new decorator is created.

Fixes #606

### DIFF
--- a/src/Controller/Component/AuthenticationComponent.php
+++ b/src/Controller/Component/AuthenticationComponent.php
@@ -373,7 +373,7 @@ class AuthenticationComponent extends Component implements EventDispatcherInterf
         $result = $service->impersonate(
             $controller->getRequest(),
             $controller->getResponse(),
-            $identity,
+            $identity->getOriginalData(),
             $impersonated
         );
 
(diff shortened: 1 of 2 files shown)


src/Controller/Component/AuthenticationComponent.php
```diff
@@ -373,7 +373,7 @@ public function impersonate(ArrayAccess $impersonated)
         $result = $service->impersonate(
             $controller->getRequest(),
             $controller->getResponse(),
-            $identity,
+            $identity->getOriginalData(),
```
ADmad: Seems like we need to widen the type for `ImpersonationInterface::impersonate()`

markstory: Fixed this with some 'casting'.

PR #614, opened 2023-04-29
2.x => 3.x merge and fix tests

LordSimal:


LordSimal:
@ADmad I'd need your help with this named argument finder change. How do we handle this in the OrmResolver?

src/Authenticator/CookieAuthenticator.php
```diff
@@ -28,6 +28,7 @@
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
 use RuntimeException;
+use function Cake\Core\deprecationWarning;
```
ADmad: All deprecations can be simply dropped for 3.x.

ADmad:
> @ADmad I'd need your help with this named argument finder change. How do we handle this in the OrmResolver?

Just changing the [following line](https://github.com/cakephp/authentication/blob/3.x/src/Identifier/Resolver/OrmResolver.php#L66) from `$query->find($finder, $options);` to `$query->find($finder, ...$options);` should suffice.

ADmad:
I took care of the finder changes.

src/Authenticator/CookieAuthenticator.php
LordSimal: Right, forgot that 🙈

PR #637, opened 2023-09-14
adjust 3.x docs config

LordSimal:
Don't know if this 100% fixes the cake5 docs https://book.cakephp.org/authentication/3/en/index.html but lets see 😁


markstory:
This is a necessary step in getting the docs published :clap:

Issue #643, opened 2023-10-22
Reduce constraint for psr/http-message


swiffer:
Similar to what has been done here: https://github.com/cakephp/cakephp/pull/17360

Can we reduce this to widen compatibility with other packages?

dereuromark:
Make a PR and see what happens :)

swiffer:
@dereuromark, done - but the GitHub Actions Workflow will not install psr/http-message ^1.0 until CakePHP 5.0.2? is released and we are setting this as lower bound for cakephp/authentication i guess.

swiffer:
@dereuromark can you tag 3.0.3 (since CakePHP 5.0.2 is out now)

swiffer:
Sorry for asking again - could we have a 3.0.3 release tagged or is something pending @dereuromark ?

dereuromark:
I usually dont release them
@markstory did afaik

PR #644, opened 2023-10-22
Allow psr/http-message ^1.0 next to ^2.0

swiffer:


composer.json
```diff
@@ -13,7 +13,7 @@
         "cakephp/http": "^5.0",
         "laminas/laminas-diactoros": "^3.0",
         "psr/http-client": "^1.0",
-        "psr/http-message": "^2.0",
+        "psr/http-message": "^1.0 || ^2.0",
```
ADmad:
```suggestion
        "psr/http-message": "^1.1 || ^2.0",
```

swiffer:
@ADmad - done. see my comment here before merging:

https://github.com/cakephp/authentication/issues/643#issuecomment-1774161512

Issue #648, opened 2023-11-24
Impersonation for non-persistent authenticators


segy:
Hi 

we were planning to implement impersonation for a stateless authenticator and it looked logical to me, to implement `ImpersonationInterface`, so we can use auth component for that. But this interface extends `PersistenceInterface`. I don't see any logic in this inheritance. Could you please explain, if this is needed? If not, I can create PR for the necessary changes.

Thank you

ADmad:
How do you expect impersonation for stateless authentication to work?

segy:
we have this scenario (simplified) for JwtAuthenticator:
- we have user encoded in the token
- when we start the impersonation, the service will return a modified token containing also impersonation information
- when he stops, it returns the standard token

segy:
and the thing is, i couldn't find any reason for `ImpersonationInterface` to extend the `PersistenceInterface`. even the `SessionAuthenticator` (that is the only class implementing this) has a code `implements PersistenceInterface, ImpersonationInterface`, so the only change here would be to remove `extends PersistenceInterface`.

markstory:
> when we start the impersonation, the service will return a modified token containing also impersonation information
> when he stops, it returns the standard token

Where will you store the standard token if authentication is stateless?

segy:
nowhere. if we carry information about the impersonator, we can simply issue a new one when he decides to stop impersonating.

markstory:
> nowhere. if we carry information about the impersonator, we can simply issue a new one when he decides to stop impersonating.

So how would `stopImpersonating` and `isImpersonating` work? If you just want to manage the impersonation credentials yourself you should be able to do that already.

segy:
Of course I am able to do that. But sorry... I don't understand, what we are talking about :) If you insist, that the ImpersonationInterface extending PersistenceInterface is necessary and you don't want to change it, just close this.

segy:
but to answer your question: the `isImpersonating` obtains the request (and the token, which contains this information, is in the header). the `impersonate` and also `stopImpersonating` allows us to modify the response, so this logic can be handled in the authenticator.

ADmad:
> I don't understand, what we are talking about :)

Whenever someone opens an RFC we first try to understand the use case before we take a decision on whether to go ahead with the changes, hence the questions.

segy:
i added a proposed change, so you can consider.

markstory:
[...]

PR #650, opened 2023-11-29
Remove PersistenceInterface inheritance from ImpersonationInterface

segy:
solves https://github.com/cakephp/authentication/issues/648

### DIFF
--- a/src/Authenticator/ImpersonationInterface.php
+++ b/src/Authenticator/ImpersonationInterface.php
@@ -20,7 +20,7 @@ use ArrayAccess;
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
 
-interface ImpersonationInterface extends PersistenceInterface
+interface ImpersonationInterface
 {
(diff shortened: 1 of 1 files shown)


dereuromark:
Instead of changing the existing one, which is also somewhat BC breaking
Would it work to create your own `StatelessImpersationInterface extends ImpersonationInterface` instead?
And for your case just throw NotImplementedException instead in those methods of the PersistenceInterface?

dereuromark:
I do agree that 

    extends AbstractAuthenticator implements PersistenceInterface, ImpersonationInterface

is not clean, though.
As it duplicates the interface contract for the persistence part.

As such, maybe the change is fine after all - given that the changelog could outline the behavior change.

It seems to work because

    interface AuthenticationServiceInterface extends PersistenceInterface

so the now "removed" part of Persistence on Impersonation

    class AuthenticationService implements AuthenticationServiceInterface, ImpersonationInterface

is covered by that interface.

ADmad (approved):

PR #658, opened 2024-01-07
move wiki version table to readme

LordSimal:
I added it to the 3.x branch because its our current default branch

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -17,6 +17,14 @@ identification. Authorization is a [separate
 concern](https://en.wikipedia.org/wiki/Separation_of_concerns) that has been
 packaged into a separate [authorization plugin](https://github.com/cakephp/authorization).
 
+## Version map
+
+| Authentication plugin | branch | CakePHP core | PHP min |
+|-----------------------|--------|--------------|---------|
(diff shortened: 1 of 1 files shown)


dereuromark:
This is quite the opposite of what should be done
As keeping this in sync across Branches is annoying

https://github.com/FriendsOfCake/awesome-cakephp/blob/master/CONTRIBUTING.md#tips-for-creating-cakephp-plugins

Why not just adding a link to the table as we so far did in other repos?

LordSimal:
> As keeping this in sync across Branches is annoying

why should we need to keep this table in sync with other branches?
I'd say lets just start putting it in the current default branch and just add to it in future branches.

Otherwise I am also fine if we just put that version table in the generated docs even though that results in the same version based tables.

In the end I just want to get rid of the github wiki if its not used for anything really usefull.

dereuromark:
It is used for this very useful part IMO :)
As otherwise the older branches would always be lying, as in: they dont contain the table, but only the historical part of it.

version maps by design should be on a different dimension than normal docs and other things only relevant to each "version line"

PR #665, opened 2024-03-25
3.next: add ability to set custom container instance 

### LINKED ISSUE #646: Feature request: Make Authentication service available via DI in the Middleware
Related to https://github.com/cakephp/authorization/issues/265

Per [a recent community discussion](https://cakesf.slack.com/archives/C053DPNGT/p1699626768751309) it would be helpful for some apps to be able to obtain an instance of `AuthenticationServiceInterface` and/or the currently configured `AuthenticationService` from the service container.

Currently there is no way to add the `AuthenticationService` as a dependency to userland Services registered in the service container because the `ServerRequest` is added to the container _after_ the user's container registration code has run. This means that the only avenue for accessing the instantiated `AuthenticationService` is via a request attribute.

If the service were available via DI, that Service class could list it as a dependency when configuring the application service container and then call it directly from within the service. This provides greater type safety and less (direct) reliance on the HTTP request object.

jamisonbryant:
Refs: https://github.com/cakephp/authorization/pull/278 & https://github.com/cakephp/authentication/issues/646. Forward-ported to `3.next`.

### DIFF
--- a/src/Middleware/AuthenticationMiddleware.php
+++ b/src/Middleware/AuthenticationMiddleware.php
@@ -23,6 +23,7 @@ use Authentication\Authenticator\AuthenticationRequiredException;
 use Authentication\Authenticator\StatelessInterface;
 use Authentication\Authenticator\UnauthenticatedException;
 use Cake\Core\ContainerApplicationInterface;
+use Cake\Core\ContainerInterface;
 use Laminas\Diactoros\Response;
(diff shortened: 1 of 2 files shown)


ADmad:
@jamisonbryant I have updated the `3.next` branch and rebased your branch on top to avoid showing additional commits in this PR. You can do `git pull -f` locally to update.

PR #668, opened 2024-07-25
3.1.0

ADmad:
I think it's about time we release a new minor with DIC support.


ADmad:
This can be merged after https://github.com/cakephp/authentication/pull/670 is merged.

PR #670, opened 2024-07-26
Add convenience method to get the identifier.

ADmad:
If you just need that id/identifier then using this new method would be more convenient than doing `$this->Authentication->getIdentityData('id')` (which can throw an exception) or doing `$this->Authentication->getIdentity()?->getIdentifier()`.

### DIFF
--- a/src/Controller/Component/AuthenticationComponent.php
+++ b/src/Controller/Component/AuthenticationComponent.php
@@ -241,6 +241,16 @@ class AuthenticationComponent extends Component implements EventDispatcherInterf
     }
 
     /**
+     * Get the identifier (primary key) of the identity.
+     *
+     * @return array|string|int|null
+     */
+    public function getIdentifier(): array|string|int|null
+    {
+        return $this->getIdentity()?->getIdentifier();
+    }
+
+    /**
      * Returns the identity used in the authentication attempt.
      *
      * @return \Authentication\IdentityInterface|null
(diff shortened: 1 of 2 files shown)


src/Controller/Component/AuthenticationComponent.php
```diff
@@ -240,6 +240,16 @@ public function getResult(): ?ResultInterface
         return $this->getAuthenticationService()->getResult();
     }
 
+    /**
+     * Get the identifier (primary key) of the identity.
+     *
+     * @return array|string|int|null
+     */
+    public function getIdentifier(): array|string|int|null
```
dereuromark, line 248: The one not so nice thing about such a generic api is that array|scalar is always an issue for any app usage, in regards to developer or static analyzers.
Would be nice if there was a more concrete API to either get scalar or not, I always end up having to inline annotate everything that has such an API
- getScalarIdentifier()
- getArrayIdentifier() / getCompositeIdentifier()

etc as idea.

But I guess that is out of scope here. So no objections from me, just this remark.

markstory: The `array` part of this type is the problem imo. We've inherited that from the ORM though because we need to support composite primary keys :cry:

ADmad: Yeah the returns types are due to the possible pK values types. Addressing that would be a separate issue.

dereuromark: Well, there is an option to separate here as by my proposal, throwing exception in the other case.
Removing the need to delegate this check to the consumer.

But as mentioned, nothing that is important at this point. Just a though.

markstory: If we did narrow types here wouldn't we be left with partial support for composite primary keys? While the types are a bit annoying, I'd rather we have consistent support for composite primary keys over ergonomic types in one method.

PR #673, opened 2024-09-13
merge 2.x => 2.next

LordSimal:


ADmad:
I doubt we are are going to do a new Cake 4.x compatible minor. `2.next` should be just dropped.

LordSimal:
We actually do have changes in 2.next which are not yet released (like the [DIC support](https://github.com/cakephp/authentication/pull/647))

ADmad:
Wasn't are of that, okay.

PR #676, opened 2024-10-05
Fix PHP error when checking for impersonation without authentication.

ADmad:


markstory:
Tests are failing in an interesting way.

ADmad:
Ugghh.. Mocking issue.

PR #681, opened 2024-11-17
2.x update stan

LordSimal:


src/Authenticator/Result.php
```diff
@@ -50,7 +50,7 @@ class Result implements ResultInterface
     /**
      * Sets the result status, identity, and failure messages
      *
-     * @param null|array|\ArrayAccess $data The identity data
+     * @param null|array|\ArrayAccess|mixed $data The identity data
```
LordSimal, line 53: These `mixed` types needed to be added since we check for this and manually throw an exception. Otherwise phpstan would complain, that the `if()` cannot be reached.

ADmad: Seems like we forgot to add types to this class when updating for 5.x. I would just add proper types to the props/methods and bump the plugin version as required to minor/major.

psalm-baseline.xml
```diff
@@ -1,60 +1,120 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<files psalm-version="dev-master@">
+<files psalm-version="5.26.1@d747f6500b38ac4f7dfc5edbcae6e4b637d7add0">
+  <file src="src/AuthenticationService.php">
+    <PossiblyNullArgument>
+      <code><![CDATA[$provider]]></code>
+    </PossiblyNullArgument>
+    <RiskyTruthyFalsyComparison>
```
ADmad: I would simply disable this `RiskyTruthyFalsyComparison` in the main config file for all repos, it's pretty annoying.

src/Authenticator/Result.php
LordSimal: We did not miss it, this is 2.x and therefore CakePHP 4
Or would you not update stan of the 2.x version of this plugin?

dereuromark: I wouldnt. Only cake5 ones IMO.

PR #682, opened 2024-11-17
3.x update stan

LordSimal:


src/Authenticator/CookieAuthenticator.php
```diff
@@ -192,8 +192,7 @@ protected function _createToken(ArrayAccess|array $identity): string
 
         $usernameField = $this->getConfig('fields.username');
 
-        /** @var string */
-        return json_encode([$identity[$usernameField], $hash]);
+        return json_encode([$identity[$usernameField], $hash]) ?: '';
```
dereuromark: I just (string) cast it usually. But both works.

LordSimal: I'd prefer to prevent using the PHP casting system where possible 😅

src/AuthenticationService.php
```diff
@@ -492,8 +493,7 @@ public function isImpersonating(ServerRequestInterface $request): bool
      */
     protected function getImpersonationProvider(): ImpersonationInterface
     {
-        /** @var \Authentication\Authenticator\ImpersonationInterface $provider */
-        $provider = $this->getAuthenticationProvider();
+        $provider = $this->getAuthenticationProvider() ?? new stdClass();
```
ADmad: This is silly. No need to create a `stdClass` instance if `null` is returned. Instead change the next line to `if ($provider === null)`  and the code after as necessary.

src/Authenticator/CookieAuthenticator.php
ADmad: Does it stop complaining about the `false` return if you use the [JSON_THROW_ON_ERROR](https://www.php.net/manual/en/json.constants.php#constant.json-throw-on-error) flag?

We should do that anyway instead of silently returning an empty string.

src/AuthenticationService.php
LordSimal: I tried that but phpstan doesn't like having `get_class` getting passed a `null` 
Therefore I added a separate if with a separate exception.

src/Authenticator/CookieAuthenticator.php
LordSimal: It does stop complaining. thanks

PR #686, opened 2024-12-13
merge 2.x => 3.x

LordSimal:


ADmad:
The PR doesn't contain any actual changes. 3.x is already the current primary branch. So any new changes go there and are backported if required.

PR #694, opened 2025-03-14
correct the secretKey filename in getAuthenticationService function

dype35:
it the jwt.key file instead of jwt.pem


markstory:
Thank you 🎉

PR #696, opened 2025-03-20
Mark strings for translations

ByteHunter:
These strings can be marked for translations as they appear in HTTP responses (ie: API responses)


LordSimal:
`__d()` is only globally available, if a user has [this line ](https://github.com/cakephp/app/blob/5.x/config/bootstrap.php#L51) in their app active.

Therefore this needs to use the namespaced function verions.

LordSimal:
The main question now for me the fact, if this should go in 3.next or 3.x

Its not like this is a new feature...

tests/bootstrap.php
```diff
@@ -54,6 +55,14 @@
     ],
 ]);
 
+Cache::setConfig([
+    '_cake_translations_' => [
+        'engine' => 'File',
+        'prefix' => 'cake_core_',
+        'serialize' => true,
```
markstory:
```suggestion
        'engine' => 'Array',
```

No need to write to files during tests.

ByteHunter:
@LordSimal I've marked this for 3.next (mainly because was not sure if it fitted to 3.x). I welcome any sugestions/corrections on that, thanks 🙂

tests/bootstrap.php
LordSimal: Then we should do that in the [core testsuite](https://github.com/cakephp/cakephp/blob/5.x/tests/bootstrap.php#L92-L103) as well

ByteHunter:
@LordSimal I merged with the 3.x branch, is it acceptable or should I change that as it was?

markstory:
Thank you 🎉

ADmad:
The `cakephp/i18n` package is not a dependency of this lib https://github.com/cakephp/authentication/blob/6249660cd014d0df5e61a384833be5197b40caa2/composer.json#L25-L32

So these changes will cause an error when the lib is not used in a full Cake app. The usage of `__d()` need to be wrapped in `method_exists()` checks.

PR #700, opened 2025-04-16
Clarify multi column

dereuromark:
It is a bit hidden in general, maybe this makes it more clear for multi column as here the form field key would usually not be username, but sth more generic like "login".

### DIFF
--- a/docs/en/identifiers.rst
+++ b/docs/en/identifiers.rst
@@ -38,9 +38,9 @@ Configuration options:
 
 -  **fields**: The fields for the lookup. Default is
    ``['username' => 'username', 'password' => 'password']``. You can
-   also set the table field name to an array. For e.g. using
+   also set the lookup field name to an array. For e.g. using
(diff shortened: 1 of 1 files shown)


docs/en/identifiers.rst
```diff
@@ -38,9 +38,9 @@ Configuration options:
 
 -  **fields**: The fields for the lookup. Default is
    ``['username' => 'username', 'password' => 'password']``. You can
-   also set the ``username`` to an array. For e.g. using
-   ``['username' => ['username', 'email'], 'password' => 'password']``
-   will allow you to match value of either username or email columns.
+   also set the lookup field name to an array. For e.g. using
+   ``['login' => ['username', 'email'], 'password' => 'password']``
```
ADmad, line 42: You can't change the key. The code checks for `username` https://github.com/cakephp/authentication/blob/3.x/src/Identifier/PasswordIdentifier.php#L96

PR #703, opened 2025-04-17
Fix design flaw in auth process.

### LINKED ISSUE #701: Design flaw - can 5xx the application
There is quite the design flaw with the new auth via middleware that checks continuesly if it can log you in.
If the session is gone suddenly (I can also reproduce it by deleting the session content in the cookie), it will try to log you in based on e.g.

```php
// Load the authenticators. Session should be first.
            'fields' => $fields,
            'loginUrl' => Router::url([
                'prefix' => false,
                'plugin' => false,
                'controller' => 'Account',
                'action' => 'login',
            ]),
        ]);
        $service->loadAuthenticator('Authentication.Cookie', [
            'rememberMeField' => 'remember_me',
            'fields' => [
                'username' => 'email',
                'password' => 'password',
            ],
            'loginUrl' => Router::url([
                'prefix' => false,
                'plugin' => false,
                'controller' => 'Account',
                'action' => 'login',
            ]),
        ]);
        $service->loadAuthenticator('Tools.LoginLink', [
            'loginUrl' => Router::url([
                'prefix' => false,
                'plugin' => false,
                'controller' => 'Account',
                'action' => 'login',
            ]),
        ]);
```

NOW: It will log you in, so it returns the logged in user, but the session is not filled yet, and the request continues.
So the page and templates expect for those "protected" actions that the session can be asked for user id, role etc using getOrFail().
But since this is not persisted back into session, the cookie or login link user auth is still "transient".
This creates 5xx in the application.
Only after another F5 refresh the page is back in its original state.

How do you solve this?
Do you have a way to trigger a login persist here, or do we redirect away?
My feeling is that this should redirect to the login page which then would log you in and redirect back.
This way it should have the page loaded again properly.

dereuromark:
This seems to resolve https://github.com/cakephp/authentication/issues/701

But I am not sure if thats the correct approach.
Either way, the rest of the stack must only be gone through once the persisting has happened again, otherwise the following code can be in a 5xx state.

### DIFF
--- a/src/Middleware/AuthenticationMiddleware.php
+++ b/src/Middleware/AuthenticationMiddleware.php
@@ -85,6 +85,13 @@ class AuthenticationMiddleware implements MiddlewareInterface
 
         try {
             $result = $service->authenticate($request);
+            $authenticator = $service->getAuthenticationProvider();
+
+            if ($authenticator !== null && !$authenticator instanceof StatelessInterface) {
+                assert($result->getData() !== null);
+                $service->persistIdentity($request, new Response(), $result->getData());
+            }
+
         } catch (AuthenticationRequiredException $e) {
             $body = new Stream('php://memory', 'rw');
             $body->write($e->getBody());
@@ -104,16 +111,6 @@ class AuthenticationMiddleware implements MiddlewareInterface
 
         try {
             $response = $handler->handle($request);
-            $authenticator = $service->getAuthenticationProvider();
-
-            if ($authenticator !== null && !$authenticator instanceof StatelessInterface) {
-                /**
-                 * @psalm-suppress PossiblyNullArgument
-                 * @phpstan-ignore-next-line
-                 */
-                $return = $service->persistIdentity($request, $response, $result->getData());
-                $response = $return['response'];
-            }
         } catch (UnauthenticatedException $e) {
             $url = $service->getUnauthenticatedRedirectUrl($request);
             if ($url) {


src/Middleware/AuthenticationMiddleware.php
```diff
@@ -85,6 +85,13 @@ public function process(ServerRequestInterface $request, RequestHandlerInterface
 
         try {
             $result = $service->authenticate($request);
+            $authenticator = $service->getAuthenticationProvider();
+
+            if ($authenticator !== null && !$authenticator instanceof StatelessInterface) {
+                assert($result->getData() !== null);
+                $service->persistIdentity($request, new Response(), $result->getData());
```
ADmad, line 92: You are ignoring the return value of `persistIdentity()`.  When `CookieAuthenticator` is used the return value will contain the updated response instance with the header set for the cookie. So with your patch the cookie will never be set on the client.

ADmad: Nor you can return the response instance at this point as returning a response from the middleware will short circuit the response handling by the framework.

dereuromark: OK, how can we fix this up to adhere to that?

ADmad: There's nothing that can be done here. PSR-15's middleware implementation doesn't support passing an updated response instance down the queue, it can only pass down the request.

PR #705, opened 2025-04-18
Stan cleanup

ADmad:


src/AuthenticationService.php
```diff
@@ -212,14 +212,13 @@ public function authenticate(ServerRequestInterface $request): ResultInterface
      * @param \Psr\Http\Message\ServerRequestInterface $request The request.
      * @param \Psr\Http\Message\ResponseInterface $response The response.
      * @return array Return an array containing the request and response objects.
-     * @psalm-return array{request: \Psr\Http\Message\ServerRequestInterface, response: \Psr\Http\Message\ResponseInterface}
+     * @return array{request: \Psr\Http\Message\ServerRequestInterface, response: \Psr\Http\Message\ResponseInterface}
```
dereuromark, line 215: It might be better to put them on phpstan-return, as those are non trivial annotations IMO.
It is possible some or most IDEs los complete return type, not even array then anymore.

ADmad: Every major IDE and stan tools understand array shapes now.

ADmad: We already have such annotations without the `@phpstan-` prefix in core. E.g. https://github.com/cakephp/cakephp/blob/64789870f0b2b7b92de9ca85871c2da7193345b6/src/Http/Session.php#L74

dereuromark:
...

ADmad:
You commented on the wrong PR :)

PR #706, opened 2025-04-18
Add allowUnauthenticatedAll() convenience wrapper

### LINKED ISSUE #645: allowUnauthenticated() for all actions
Refs https://github.com/cakephp/authentication/issues/361

It would be nice to have a less hacky way of

    $this->Authentication->allowUnauthenticated([$this->request->getParam('action')]);
    
I can understand that not using any argument does not feel "safe" by default.
As such, maybe the proposed `'*'` string could be very clear on this.
Or we provide an own speaking method for it?

    $this->Authentication->allowUnauthenticatedAll()
    
All()/Any() etc

dereuromark:
Resolves https://github.com/cakephp/authentication/issues/645

setting the requireIdentity to false is not convenient for a specific controller

### DIFF
--- a/src/Controller/Component/AuthenticationComponent.php
+++ b/src/Controller/Component/AuthenticationComponent.php
@@ -193,6 +193,23 @@ class AuthenticationComponent extends Component implements EventDispatcherInterf
     /**
      * Set the list of actions that don't require an authentication identity to be present.
      *
+     * @return $this
+     */
+    public function allowUnauthenticatedAll()
+    {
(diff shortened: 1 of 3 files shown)


ADmad:
Give that the component already has a `requireIdentity` flag to disable identity check, just adding a convenience function which sets it to false would be a much easier solution.

```php
public function disableIdentityCheck(): void
{
    $this->setConfig('requireIdentity', false);
}

PR #707, opened 2025-04-18
RFC Explore storing user as non object.

dereuromark:
I would be open to feedback on how this could be accomplished.
Being able to store the user entity as array or better yet JSON is much more robust.

Any change in the object or its children (created/modified DateTime), related models, ..., role enum, ...) will all cause an immediate invalidation right now.
Storing them in their agnostic form and then rebuilding seems much more logical to me.

Currently it would need to get a transformer passed in though, as it knows too much about the ORM/model/entity.
Right now a "modelClass" defaulting to "Users" could suffice.

For me in a first test this seems to properly populate again:
```php
object(App\Model\Entity\User) id:1 {
  'first_name' => 'Admin'
  'role' => object(App\Model\Enum\UserRole) id:1 {
    name => 'Admin'
    value => 'admin'
  }
  'email' => 'admin@example.com'
  'email_verified_at' => object(Cake\I18n\DateTime) id:2 {
    'hasFixedNow' => false
    'time' => '2025-04-08 00:00:00.000000'
    'timezone' => 'UTC'
  }
  'remember_token' => null
  'company_id' => (int) 1
  'last_login' => object(Cake\I18n\DateTime) id:3 {
    'hasFixedNow' => false
    'time' => '2025-04-16 22:03:51.000000'
    'timezone' => 'UTC'
  }
  'created' => object(Cake\I18n\DateTime) id:4 {
    'hasFixedNow' => false
    'time' => '2025-04-08 00:00:00.000000'
    'timezone' => 'UTC'
  }
  'modified' => object(Cake\I18n\DateTime) id:5 {
    'hasFixedNow' => false
    'time' => '2025-04-09 10:41:11.000000'
    'timezone' => 'UTC'
  }
  'phone' => null
  'last_name' => 'Tester'
```

I put my idea behind a feature flag serialize true/false
this way it could be made BC

### DIFF
--- a/src/Authenticator/SessionAuthenticator.php
+++ b/src/Authenticator/SessionAuthenticator.php
@@ -15,10 +15,12 @@ declare(strict_types=1);
  */
 namespace Authentication\Authenticator;
 
+use App\Model\Entity\User;
 use ArrayAccess;
 use ArrayObject;
 use Authentication\Identifier\AbstractIdentifier;
 use Cake\Http\Exception\UnauthorizedException;
+use Cake\ORM\TableRegistry;
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
 
@@ -59,6 +61,8 @@ class SessionAuthenticator extends AbstractAuthenticator implements PersistenceI
         /** @var \Cake\Http\Session $session */
         $session = $request->getAttribute('session');
         $user = $session->read($sessionKey);
+        $user = json_decode($user, true);
+        $user = TableRegistry::getTableLocator()->get('Users')->newEntity($user, ['validate' => false]);
 
         if (empty($user)) {
             return new Result(null, Result::FAILURE_IDENTITY_NOT_FOUND);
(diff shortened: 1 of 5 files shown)


dereuromark:
Possible alternative that could be cleaner:
https://github.com/dereuromark/cakephp-tinyauth/pull/153

ADmad:
> Currently it would need to get a transformer passed in though, as it knows too much about the ORM/model/entity.

We obviously can't just stick an ORM call into the session authenticator.

Check out how the CookieAuthenticator functions. It stores only the username and a hashed token in the cookie and then for authentication passes those values to the identifier collection (the TokenIdentifier should be loaded for the auth service) which takes care of returning the identity (usually the User entity).

You can create a new SessionIdentifier class which uses the OrmResolver to get the table instance and recreates the entity.

dereuromark:
Do you think we need the token here too? Or would the ID suffice?
See linked TinyAuth PR.

How is session spoofing etc currently preventing with the user data array?
Don't think there is much difference here, is there?

dereuromark:
And if we add this, what would be a good name for this?
- UserIdSessionAuthenticator
- IdSessionAuthenticator
- PrimaryKeySessionAuthenticator

jamisonbryant:
`PrimaryKeySessionAuthenticator` sounds good because there’s already a `getPrimaryKey()` method on Table classes, which handles composite PKs.

I’m less a fan of “User” as there may be some tables that don’t use that nomenclature, e.g. “Clients” or even OAuth application ID.

PR #708, opened 2025-04-18
Allow unauthenticated all

dereuromark:
Replaces https://github.com/cakephp/authentication/pull/706

### DIFF
--- a/src/Controller/Component/AuthenticationComponent.php
+++ b/src/Controller/Component/AuthenticationComponent.php
@@ -191,6 +191,17 @@ class AuthenticationComponent extends Component implements EventDispatcherInterf
     }
 
     /**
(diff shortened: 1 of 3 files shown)


ADmad:
This should target 3.next

dereuromark:
Dont think we have one such branch yet, maybe this can become it.

src/Controller/Component/AuthenticationComponent.php
```diff
@@ -190,6 +190,17 @@ protected function doIdentityCheck(): void
         }
     }
 
+    /**
+     * Disables the identity check for this controller and as all its actions.
+     * They then don't require an authentication identity to be present.
```
markstory, line 195:
```suggestion
     * Disables identity checks for the current action.
```

PR #710, opened 2025-04-18
Add PrimaryKeySessionAuthenticator

dereuromark:
Replaces https://github.com/cakephp/authentication/pull/707

Ports https://github.com/dereuromark/cakephp-tinyauth/pull/153 as direct class into this plugin.

```php
    $service->loadAuthenticator('Authentication.PrimaryKeySession');
        'tokenField' => 'id', // lookup for DB table
        'dataField' => 'key', // incoming data
        'resolver' => 'Authentication.Orm',
    ]);
```

### DIFF
--- a/docs/en/authenticators.rst
+++ b/docs/en/authenticators.rst
@@ -37,12 +37,22 @@ It also helps to avoid session invalidation.
 Session itself stores the entity object including nested objects like DateTime or enums.
 With only the ID stored, the invalidation due to objects being modified will also dissolve.
 
-Make sure to match this with a Token identifier with ``id`` keys:
+Make sure to match this with a Token identifier with ``key``/``id`` keys::
 
     $service->loadIdentifier('Authentication.Token', [
-        'tokenField' => 'id',
-        'dataField' => 'id',
+        'tokenField' => 'id', // lookup for DB table
+        'dataField' => 'key', // incoming data
         'resolver' => 'Authentication.Orm',
+    ]);
+
+    $service->loadAuthenticator('Authentication.PrimaryKeySession', [
+        'urlChecker' => 'Authentication.CakeRouter',
+        'loginUrl' => [
+            'prefix' => false,
+            'plugin' => false,
+            'controller' => 'Users',
+            'action' => 'login',
+        ],
     ]);
 
 Form
--- a/docs/en/authenticators.rst
+++ b/docs/en/authenticators.rst
@@ -26,6 +26,25 @@ Configuration options:
    identifier in your user storage. Defaults to ``username``. This option is
    used when the ``identify`` option is set to true.
 
+PrimaryKeySession
+=================
+
+This is an improved version of Session that will only store the primary key.
+This way the data will always be fetched fresh from the DB and issues like
+having to update the identity when updating account data should be gone.
+
+It also helps to avoid session invalidation.
+Session itself stores the entity object including nested objects like DateTime or enums.
+With only the ID stored, the invalidation due to objects being modified will also dissolve.
+
+Make sure to match this with a Token identifier with ``id`` keys:
+
+    $service->loadIdentifier('Authentication.Token', [
+        'tokenField' => 'id',
+        'dataField' => 'id',
+        'resolver' => 'Authentication.Orm',
+    ]);
+
 Form
 ====
 
--- a/docs/en/authenticators.rst
+++ b/docs/en/authenticators.rst
@@ -39,13 +39,14 @@ With only the ID stored, the invalidation due to objects being modified will als
 
 Make sure to match this with a Token identifier with ``key``/``id`` keys::
 
-    $service->loadIdentifier('Authentication.Token', [
-        'tokenField' => 'id', // lookup for DB table
-        'dataField' => 'key', // incoming data
-        'resolver' => 'Authentication.Orm',
-    ]);
-
     $service->loadAuthenticator('Authentication.PrimaryKeySession', [
+        'identifier' => [
+            'Authentication.Token' => [
+                'tokenField' => 'id', // lookup for resolver and DB table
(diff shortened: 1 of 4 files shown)


docs/en/authenticators.rst
```diff
@@ -26,6 +26,25 @@ Configuration options:
...
+
+This is an improved version of Session that will only store the primary key.
+This way the data will always be fetched fresh from the DB and issues like
+having to update the identity when updating account data should be gone.
+
+It also helps to avoid session invalidation.
+Session itself stores the entity object including nested objects like DateTime or enums.
+With only the ID stored, the invalidation due to objects being modified will also dissolve.
+
+Make sure to match this with a Token identifier with ``key``/``id`` keys:
```
markstory:
```suggestion
Make sure to match this with a Token identifier with ``key``/``id`` keys::
```

docs/en/authenticators.rst
```diff
@@ -26,6 +26,25 @@ Configuration options:
...
+
+    $service->loadIdentifier('Authentication.Token', [
+        'tokenField' => 'id', // lookup for DB table
+        'dataField' => 'key', // incoming data
+        'resolver' => 'Authentication.Orm',
+    ]);
```
markstory: Shouldn't we also give an example of loading the `PrimaryKeySession` authenticator?

tests/TestCase/Authenticator/PrimaryKeySessionAuthenticatorTest.php
```diff
@@ -0,0 +1,441 @@
...
+        $response = new Response();
+        $authenticator = new PrimaryKeySessionAuthenticator($this->identifiers);
+
+        $data = new ArrayObject(['id' => 1]);
+
+        $this->sessionMock
+            ->expects($this->exactly(2))
+            ->method('check')
+            ->with(
+                ...static::withConsecutive(['Auth'], ['Auth']),
```
markstory, line 209: Why `static`?

src/Authenticator/PrimaryKeySessionAuthenticator.php
```diff
@@ -0,0 +1,110 @@
...
+class PrimaryKeySessionAuthenticator extends SessionAuthenticator
+{
+    /**
+     * @param \Authentication\Identifier\IdentifierInterface $identifier
+     * @param array<string, mixed> $config
+     */
+    public function __construct(IdentifierInterface $identifier, array $config = [])
+    {
+        $config += [
+            'identifierKey' => 'key',
```
markstory, line 24: If you're inheriting from `SessionAuthenticator` how does `identifierKey` interact with `fields`?

What is `key` here? I don't see that as a field in the `auth_users` table that your tests use. Is it primarily to be aligned with the identifier configuration?

docs/en/authenticators.rst
```diff
@@ -26,6 +26,25 @@ Configuration options:
...
+having to update the identity when updating account data should be gone.
+
+It also helps to avoid session invalidation.
+Session itself stores the entity object including nested objects like DateTime or enums.
+With only the ID stored, the invalidation due to objects being modified will also dissolve.
+
+Make sure to match this with a Token identifier with ``key``/``id`` keys:
+
+    $service->loadIdentifier('Authentication.Token', [
+        'tokenField' => 'id', // lookup for DB table
```
markstory: This feels like it could be risky when combined with other authenticators.

dereuromark: Token field? thats the DB lookup field internally, that should be fine
It basically builds the conditions for the ORM resolver:

    return $this->getResolver()->find($conditions);

The dataField is the one coming from the authenticator and possibly matching another identifier

src/Authenticator/PrimaryKeySessionAuthenticator.php
dereuromark: key here maps to the dataField of the identifier, this is how they pass in the data.

tests/TestCase/Authenticator/PrimaryKeySessionAuthenticatorTest.php
dereuromark: This is copy and paste from the other Session test.

docs/en/authenticators.rst
dereuromark: None of the others had actually examples, but I added it now.

src/Authenticator/PrimaryKeySessionAuthenticator.php
dereuromark: fields is unused afaik.

docs/en/authenticators.rst
ADmad: > This feels like it could be risky when combined with other authenticators.

I have similar concerns. While our default configs might not have any config overlap, we should ensure there isn't a possibility of opening a vulnerability even with custom config and potentially end up with something similar to the algo config vulnerability that were found in various JWT libs.

I think we should first work of #711 to directly associate identifiers with authenticators and add this new authenticator only after that.

dereuromark: https://github.com/cakephp/authentication/pull/712 is open
But IMO those don't collide.
The deprecation in parallel together with new docs will make sure people move over to a safer approach.
While it should still work out for everyone so far even if on the deprecated approach.

dereuromark:
Anything missing? Or should we continue here?

dereuromark:
Should we close this, if there is no interest further?

LordSimal:
I think its a good idea and should be added to the next version of the auth plugin

dereuromark:
I am successfully using it via TinyAuth for a month now, seems to work perfectly, even with a short cache layer in between (redis).

dereuromark:
Do you want to make a release here @markstory

PR #714, opened 2025-04-20
allow unauthenticatedRedirect as array cake style.

dereuromark:
Seemed weird to me that in normal Cake context in a Cake plugin the URL cannot be defined in normal array style
if array, Router::url() is usually also used in other parts, and always allowed as alternative to strings.

### DIFF
--- a/src/AuthenticationService.php
+++ b/src/AuthenticationService.php
@@ -373,7 +373,7 @@ class AuthenticationService implements AuthenticationServiceInterface, Impersona
         if ($target === null) {
             return null;
         }
-        if (is_array($target)) {
+        if (is_array($target) && class_exists(Router::class)) {
             $target = Router::url($target);
         }
         if ($param === null) {
--- a/src/AuthenticationService.php
+++ b/src/AuthenticationService.php
@@ -26,6 +26,7 @@ use Authentication\Authenticator\StatelessInterface;
 use Authentication\Identifier\IdentifierCollection;
 use Authentication\Identifier\IdentifierInterface;
 use Cake\Core\InstanceConfigTrait;
+use Cake\Routing\Router;
 use InvalidArgumentException;
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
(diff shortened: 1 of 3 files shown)


ADmad:
> ...Seemed weird to me that in normal Cake context in a Cake plugin the URL cannot be defined in normal array style

That's cause this packages isn't a cake plugin. It can be used standalone. It doesn't have `cakephp/cakephp` as a dependency.

dereuromark:
Ohh
Well, thats very theoretical, in all seriousness, is anyone actually using it outside the Cake world? :)

README also says

> [PSR7](https://www.php-fig.org/psr/psr-7/) Middleware authentication stack for the CakePHP framework.

dereuromark:
It does have a cakephp/http dependency already
Just not the other packages

There is already a bit of dependency issues I guess:
The component uses

        return Router::normalize($logoutRedirect);
        
But that would then be outside the current dependencies.
I guess a "semi optional" component then.

That said: I am also fine if we close this and keep the "plugin" a "not plugin" :)

src/AuthenticationService.php
```diff
@@ -372,6 +373,9 @@ public function getUnauthenticatedRedirectUrl(ServerRequestInterface $request):
         if ($target === null) {
             return null;
         }
+        if (is_array($target)) {
```
dereuromark: We could add a class exists check?

ADmad: Yes

ADmad:
> The component uses

Cause components are a Cake only thing and AuthComponent can't be used outside a cake app :)

dereuromark:
Yeah, makes sense.
Check added.

LordSimal:
The only real core dependency this plugin contains are related to exceptions and cookies.

image

I only know of chronos, which can be used standalone. But most other stuff is at least somewhat cakephp subpackge dependent.

But in the end I agree with dereuromark that it seems very unlikely, that this package is being used outside of the cakephp ecosystem.

dereuromark:
Well, for now nothing has to change (yet).
The "real plugin" discussion can be reopened in the future or when we major the plugin finally :)

ADmad:
> The only real core dependency this plugin contains are related to exceptions and cookies.

ADmad:
> The only real core dependency this plugin contains are related to exceptions and cookies.

`cakephp/http` is added as a dependency

PR #718, opened 2025-04-24
Alert about wrong configuration.

dereuromark:
I changed from username to login, but apparently, one authenticator wasnt adjusted.
this was hard to debug without a clear feedback.

### DIFF
--- a/src/Authenticator/CookieAuthenticator.php
+++ b/src/Authenticator/CookieAuthenticator.php
@@ -134,9 +134,14 @@ class CookieAuthenticator extends AbstractAuthenticator implements PersistenceIn
         $usernameField = $this->getConfig('fields.username');
         $passwordField = $this->getConfig('fields.password');
 
-        $salt = $this->getConfig('salt', '');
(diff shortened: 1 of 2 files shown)


src/Authenticator/CookieAuthenticator.php
```diff
@@ -134,9 +134,14 @@ protected function _createPlainToken(ArrayAccess|array $identity): string
         $usernameField = $this->getConfig('fields.username');
         $passwordField = $this->getConfig('fields.password');
 
-        $salt = $this->getConfig('salt', '');
+        if ($identity[$usernameField] === null || $identity[$passwordField] === null) {
+            throw new InvalidArgumentException(
```
markstory, line 138: Wouldn't this create a 500 for the application? Could we use an exception that will render a 40x response instead?

dereuromark: It should, it is a developer error.
It can only run into this if you misconfigured them afaik.

As otherwise it shouldnt even go into this authenticator that far.
They would at bare minimum be empty strings, never null. Null means you set them to impossible values IMO.

PR #721, opened 2025-05-02
Fix token auth prefix removal.

### LINKED ISSUE #720: str_ireplace replaces $prefix if $prefix is embedded in token string now
> My Token Auth stopped working with this change because I am using a token with the word 'token' imbedded in it e.g. "auth-token-136791ca15f1f8f30db633a1f395a2bb6af690f169349e6b23f4ae8f5ac3bd24"
> 
> 
> 
> so the change stripped out "token"
> 
> 
> 
> ```php
> 
> function stripTokenPrefixOld(string $token, string $prefix): string
> 
> {
> 
>     return str_ireplace($prefix . ' ', '', $token);
> 
> }
> 
> 
> 
> function stripTokenPrefixNew(string $token, string $prefix): string
> 
> {
> 
>     return trim(str_ireplace($prefix, '', $token));
> 
> }
> 
> 
> 
> $test = ['Token auth-token-1234567', 'Token token-34567', 'Token 1234566777'];
> 
> 
> 
> foreach ($test as $token) {
> 
>     echo "old: " . stripTokenPrefixOld($token, 'Token') . PHP_EOL;
> 
>     echo "new: " . stripTokenPrefixNew($token, 'Token') . PHP_EOL;
> 
> }
> 
> ```
> 
> 
> 
> ## Output
> 
> ```
> 
> old: auth-token-1234567
> 
> new: auth--1234567
> 
> old: token-34567
> 
> new: -34567
> 
> old: 1234566777
> 
> new: 1234566777
> 
> ```
> 
> 
> 
> Not sure if using preg_replace or just reverting is the best option:
> 
> 
> 
> ```
> 
>  return preg_replace('/^' . preg_quote($prefix, '/') . '\s+/i', '', $token);
> 
> ```
> 
> 
> 
> 
> 
>  

 _Originally posted by @toggenation in [8597b11](https://github.com/cakephp/authentication/commit/8597b1157a1dbbd03e35a5c3c1389056ba79d924#r156012612)_

dereuromark:
Resolves https://github.com/cakephp/authentication/issues/720

Does this work?

However, the test case doesnt quite seem right, it fails.
Do you have an idea for a correct test case?

### DIFF
--- a/tests/TestCase/Authenticator/TokenAuthenticatorTest.php
+++ b/tests/TestCase/Authenticator/TokenAuthenticatorTest.php
@@ -150,6 +150,16 @@ class TokenAuthenticatorTest extends TestCase
         $result = $tokenAuth->authenticate($requestWithHeaders);
         $this->assertInstanceOf(Result::class, $result);
         $this->assertSame(Result::FAILURE_IDENTITY_NOT_FOUND, $result->getStatus());
+
+        // should not modify in between
+        $requestWithHeaders = $this->request->withAddedHeader('X-Dipper-Auth', 'auth-token-13');
+        $tokenAuth = new TokenAuthenticator($this->identifiers, [
+            'header' => 'X-Dipper-Auth',
+            'tokenPrefix' => 'token_',
+        ]);
+        $result = $tokenAuth->authenticate($requestWithHeaders);
+        $this->assertInstanceOf(Result::class, $result);
+        $this->assertSame(Result::SUCCESS, $result->getStatus());
     }
 
     /**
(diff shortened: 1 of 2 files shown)


tests/TestCase/Authenticator/TokenAuthenticatorTest.php
```diff
@@ -150,6 +150,16 @@ public function testTokenPrefix()
         $result = $tokenAuth->authenticate($requestWithHeaders);
         $this->assertInstanceOf(Result::class, $result);
         $this->assertSame(Result::FAILURE_IDENTITY_NOT_FOUND, $result->getStatus());
+
+        // should not modify in between
+        $requestWithHeaders = $this->request->withAddedHeader('X-Dipper-Auth', 'auth-token-13');
```
markstory:
```suggestion
        $requestWithHeaders = $this->request->withAddedHeader('X-Dipper-Auth', 'token_auth-token-13');
```

Don't you also need the prefix on your token?

PR #722, opened 2025-05-14
Error: Supplied key param cannot be coerced into a public key

joaopatrocinio:
The secretKey in loadAuthenticator should be the public key (jwt.pem).

Using the private key I get this error:

image

While on this topic, should this property be called secretKey? It causes some confusion to whether its the public or private key.


dereuromark:
We should probably use publicKey as name, and deprecate the old one. This could be done as a follow up PR in 3.next.

PR #729, opened 2025-07-21
Dispatch a Authentication.authenticate event on successful authentication.

ADmad:
Subscribing to this event would allow the app to do to various like updating last login time, keep track of failed login attempts etc.

### DIFF
--- a/src/AuthenticationService.php
+++ b/src/AuthenticationService.php
@@ -23,9 +23,12 @@ use Authentication\Authenticator\ImpersonationInterface;
 use Authentication\Authenticator\PersistenceInterface;
 use Authentication\Authenticator\ResultInterface;
 use Authentication\Authenticator\StatelessInterface;
+use Authentication\Event\AuthenticateEvent;
 use Authentication\Identifier\IdentifierCollection;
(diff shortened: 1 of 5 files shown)


src/AuthenticationService.php
```diff
@@ -190,7 +194,12 @@ public function authenticate(ServerRequestInterface $request): ResultInterface
         $result = null;
         /** @var \Authentication\Authenticator\AuthenticatorInterface $authenticator */
         foreach ($this->authenticators() as $authenticator) {
-            $result = $authenticator->authenticate($request);
+            $result = $this->dispatchAuthenticateEvent(
```
markstory, line 197: Is it ok if this trigger on every request? Perhaps we should include that in the documentation?

ADmad: Hmm.. I overlooked the fact that this would trigger even when reading the user record from session. For the use cases I mentioned, the event is only needed for a login form POST. So triggering an event only from FormAuthenticate would make more sense.

ADmad:
I'll revisit this in the future, closing for now.

PR #737, opened 2025-08-26
Fix up identifier defaulting.

### LINKED ISSUE #735: [Improvement] Set default Identifiers based on the default configuration
Extended from https://github.com/cakephp/authentication/issues/733 , expanded into it's own issue as the previous is addressed. 

Identifiers are now to be set per Authenticator. 
But some Authenticators ask for an Identifier when they may not need one. For example, Form shouldn't need one if it's [username, password] as those are the default fields https://github.com/cakephp/authentication/blob/3.x/src/Authenticator/FormAuthenticator.php#L44 , so IMO the default Identifier should be `Authentication.Password`.

This feels like it should be enough
```php
    public function getAuthenticationService(ServerRequestInterface $request): AuthenticationServiceInterface
    {
        $service = new AuthenticationService([
            'unauthenticatedRedirect' => [
                'prefix' => false,
                'plugin' => null,
                'controller' => 'Users',
                'action' => 'login',
            ],
            'queryParam' => 'redirect',
        ]);

        // Load the authenticators. Session should be first.
        $service->loadAuthenticator('Authentication.Form');

        return $service;
    }
```

Rather than needing to do this:
```php
    public function getAuthenticationService(ServerRequestInterface $request): AuthenticationServiceInterface
    {
        $service = new AuthenticationService([
            'unauthenticatedRedirect' => [
                'prefix' => false,
                'plugin' => null,
                'controller' => 'Users',
                'action' => 'login',
            ],
            'queryParam' => 'redirect',
        ]);

        // Load the authenticators. Session should be first.
            'identifier' => 'Authentication.Password',
        ]);

        return $service;
    }
```

dereuromark:
Resolves https://github.com/cakephp/authentication/issues/735

  This makes the authentication setup much simpler. Developers can now use:
```php
  $service->loadAuthenticator('Authentication.Form');
  $service->loadAuthenticator('Authentication.HttpBasic');
  $service->loadAuthenticator('Authentication.Cookie');
  $service->loadAuthenticator('Authentication.Token');
  $service->loadAuthenticator('Authentication.Jwt');
```
  All without needing to explicitly configure identifiers, as sensible defaults are provided.

### DIFF
--- a/src/Authenticator/FormAuthenticator.php
+++ b/src/Authenticator/FormAuthenticator.php
@@ -17,6 +17,8 @@ declare(strict_types=1);
 namespace Authentication\Authenticator;
 
 use Authentication\Identifier\AbstractIdentifier;
+use Authentication\Identifier\IdentifierCollection;
+use Authentication\Identifier\IdentifierInterface;
 use Authentication\UrlChecker\UrlCheckerTrait;
 use Cake\Routing\Router;
 use Psr\Http\Message\ServerRequestInterface;
@@ -46,6 +48,22 @@ class FormAuthenticator extends AbstractAuthenticator
             AbstractIdentifier::CREDENTIAL_PASSWORD => 'password',
         ],
     ];
+
+    /**
+     * Constructor
+     *
+     * @param \Authentication\Identifier\IdentifierInterface $identifier Identifier or identifiers collection.
+     * @param array<string, mixed> $config Configuration settings.
+     */
+    public function __construct(IdentifierInterface $identifier, array $config = [])
+    {
+        // If no identifier is configured, set up a default Password identifier
+        if ($identifier instanceof IdentifierCollection && $identifier->isEmpty()) {
+            $identifier = new IdentifierCollection(['Authentication.Password']);
+        }
+
+        parent::__construct($identifier, $config);
+    }
 
     /**
      * Checks the fields to ensure they are supplied.
--- a/src/Authenticator/FormAuthenticator.php
+++ b/src/Authenticator/FormAuthenticator.php
@@ -59,7 +59,14 @@ class FormAuthenticator extends AbstractAuthenticator
     {
         // If no identifier is configured, set up a default Password identifier
         if ($identifier instanceof IdentifierCollection && $identifier->isEmpty()) {
-            $identifier = new IdentifierCollection(['Authentication.Password']);
+            // Pass the authenticator's fields configuration to the identifier
+            $identifierConfig = [];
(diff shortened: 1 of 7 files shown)


src/Authenticator/FormAuthenticator.php
```diff
@@ -47,6 +49,22 @@ class FormAuthenticator extends AbstractAuthenticator
...
+     * Constructor
+     *
+     * @param \Authentication\Identifier\IdentifierInterface $identifier Identifier or identifiers collection.
+     * @param array<string, mixed> $config Configuration settings.
+     */
+    public function __construct(IdentifierInterface $identifier, array $config = [])
+    {
+        // If no identifier is configured, set up a default Password identifier
+        if ($identifier instanceof IdentifierCollection && $identifier->isEmpty()) {
+            $identifier = new IdentifierCollection(['Authentication.Password']);
```
ADmad: If we are setting a default identifier then I would expect it to be auto configured to as per the authenticator config. For e.g. if the `fields.username` config is set to `email` for the authenticator, then the same should be passed to the identifier too. Currently we have this redundancy where we have to configure it for both the authenticator and the identifier.

dereuromark: I added a commit, is that what you had in mind?

src/Authenticator/JwtAuthenticator.php
```diff
@@ -56,7 +57,13 @@ class JwtAuthenticator extends TokenAuthenticator
...
     public function __construct(IdentifierInterface $identifier, array $config = [])
     {
-        parent::__construct($identifier, $config);
+        // Override parent's default - JWT should use JwtSubject identifier
+        if ($identifier instanceof IdentifierCollection && $identifier->isEmpty()) {
+            $identifier = new IdentifierCollection(['Authentication.JwtSubject']);
+        }
+
+        // Call TokenAuthenticator's constructor but skip its default
+        AbstractAuthenticator::__construct($identifier, $config);
```
markstory, line 66: Should this be `parent::__construct()`?

dereuromark: Then it would trigger the Token ones default I guess.

markstory: Hrm. If we need different behavior perhaps we should undo the inheritance and duplicate some code. We can do that separately too.

markstory (approved):
Looks good to me.

dereuromark:
ping @ADmad

ADmad (approved):

PR #738, opened 2025-09-01
Update password-hashers.rst

txj:
Fix problems with examples when using the preferred 3.3.x configuration pattern.

Fixes 2 problems:
1) fallback authentication didn't work (typo in config)
2) login action failed with:
[Cake\Core\Exception\CakeException] Unknown object `Password`. in vendor/cakephp/cakephp/src/Core/ObjectRegistry.php on line 238

### DIFF
--- a/docs/en/password-hashers.rst
+++ b/docs/en/password-hashers.rst
@@ -47,7 +47,8 @@ fallback hasher as follows::
             'passwordHasher' => [
                 'className' => 'Authentication.Fallback',
                 'hashers' => [
-                    'Authentication.Default' => [
+                    'Authentication.Default',
+                    [
                         'className' => 'Authentication.Legacy',
                         'hashType' => 'md5',
                         'salt' => false, // turn off default usage of salt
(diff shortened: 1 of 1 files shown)


docs/en/password-hashers.rst
```diff
@@ -47,7 +47,8 @@ fallback hasher as follows::
             'passwordHasher' => [
                 'className' => 'Authentication.Fallback',
                 'hashers' => [
-                    'Authentication.Default' => [
+                    'Authentication.Default',
```
dereuromark, line 50: are you sure about this?

txj: Yes, with the current config both the Default and Legacy hashers do not work, I cannot login with either a bcrypt or sha1 encrypted password. With the fix both work correctly.

ADmad: @dereuromark https://github.com/cakephp/authentication/blob/a3154b58b817991d6d925845ea72dfaeeb483d40/tests/TestCase/PasswordHasher/FallbackPasswordHasherTest.php#L35

PR #739, opened 2025-09-06
Don't populate IdentifierCollection errors with empty nested arrays

ADmad:
Before this fix using `$result->getErrors()` would return `['Password' => []]` when using form authenticate with a non existent user.


ADmad:
In the future we also need to fix this inconsistency of errors returned by `Result::getErrors()`. 

If an authenticator directly sets the result's errors you would get a numerically indexed array like `[0 => 'foo']` but if the an identifier returns errors then the `IdentifierCollection` nests them using the identifier name as key and `Result::getErrors()` would return something like `['IdentifierName' => [0 => 'foo']]`.

PR #744, opened 2025-10-24
Set the redirect query param only for GET requests.

ADmad:
Currently if the session expires for a logged in user and they click on a (post) link for "Delete" action, after login they get redirected to the delete action with a GET request, resulting in an exception since that action is only accessible with POST/DELETE request.


dereuromark (approved):
Makes sense

dereuromark:
I adjusted the docs in a follow up PR.

Issue #716, opened 2025-04-20
New major


dereuromark:
We want to make a new major soon were we would then remove the deprecations we added:

- IdentifierCollection (https://github.com/cakephp/authentication/pull/712)

I would also like to see the following fixed up:

- [x] CakeRouterUrl urlChecker (maybe called "Cake") auto defaults if we detect CakePHP context instead of "Default" ( https://github.com/cakephp/authentication/pull/714/commits/5607c7b70d86d5ff0ee25e3722591f54e73a9ebd ), or "Array vs String"
- [x] single URL by default, most apps only ever need a single app
       The array|string magic is not needed anymore then
       It internally accepts both string or array style as always in Cake
- [x] A MultiUrl checker can be added to specifically support multiple loginUrls
- [x] Constants for username/password shoud be at their respective classes instead of abstract, there are several ones that need either different naming or none of those pairs.

In general the next major could be more cake context bound and have it as a real dependency.
makes things much easier and the plugin most likely has never been used in an external context.

ADmad:
> In general the next major could be more cake context bound and have it as a real dependency.

I am fine with that. I highly doubt this plugin/lib is used independently without CakePHP, so we can just drop any Cake agnostic stuff.

LordSimal:
Everything I found "officially" dependent on the auth plugin on packagist is definitely cakephp related, so making the auth plugins framework agnostic is not needed.
https://packagist.org/packages/cakephp/authentication/dependents?order_by=downloads

ADmad:
Packagist won't show private project usage but again I highly doubt there are any.

LordSimal:
Sure, but the only place where we indireclty say it is "framework agnostic" is the github description with
> Authentication plugin for CakePHP. Can also be used in PSR7 based applications.

So I'd say the few users who are acutally using this outside of the cakephp framework will be able to find another solution.

dereuromark:
I adjusted the major branch as per your feedback:

- Default (old Cake)
- Generic (old Default)

Hardcoded default, no more auto detect

Adjusted the migration guide accordingly.

dereuromark:
@ADmad Are we ready for the release?

dereuromark:
I prepared the release notes - high level.

PR #748, opened 2025-11-06
4.x major: Remove deprecations, fix up MultiChecker.

dereuromark:
Prepare https://github.com/cakephp/authentication/issues/716


dereuromark:
For some reason 4.x is now protected already, even though just created.
I have the ready branch at "4.x-squashed", maybe someone with access can just force push into 4.x?

ADmad:
> For some reason 4.x is now protected already, even though just created.

The protection rule is set for `*.x`, that's way. I have temporarily changed it to `3.x`, so you should be able to force push to `4.x` now.

src/Authenticator/EnvironmentAuthenticator.php
```diff
@@ -45,6 +47,21 @@ class EnvironmentAuthenticator extends AbstractAuthenticator
...
+     * Constructor
+     *
+     * @param \Authentication\Identifier\IdentifierInterface|null $identifier Identifier instance.
+     * @param array<string, mixed> $config Configuration settings.
+     */
+    public function __construct(?IdentifierInterface $identifier, array $config = [])
+    {
+        if ($identifier === null) {
+            $identifier = IdentifierFactory::create('Authentication.Callback');
+        }
```
ADmad:
```suggestion
        $identifier ??= IdentifierFactory::create('Authentication.Callback');
```

src/Authenticator/JwtAuthenticator.php
```diff
@@ -55,14 +55,14 @@ class JwtAuthenticator extends TokenAuthenticator
...
      */
-    public function __construct(IdentifierInterface $identifier, array $config = [])
+    public function __construct(?IdentifierInterface $identifier, array $config = [])
     {
         // Override parent's default - JWT should use JwtSubject identifier
-        if ($identifier instanceof IdentifierCollection && $identifier->isEmpty()) {
-            $identifier = new IdentifierCollection(['Authentication.JwtSubject']);
+        if ($identifier === null) {
+            $identifier = IdentifierFactory::create('Authentication.JwtSubject');
         }
```
ADmad:
```suggestion
        $identifier ??= IdentifierFactory::create('Authentication.JwtSubject');
```

src/Authenticator/TokenAuthenticator.php
```diff
@@ -40,14 +40,14 @@ class TokenAuthenticator extends AbstractAuthenticator implements StatelessInter
...
      */
-    public function __construct(IdentifierInterface $identifier, array $config = [])
+    public function __construct(?IdentifierInterface $identifier, array $config = [])
     {
         // If no identifier is configured, set up a default Token identifier
-        if ($identifier instanceof IdentifierCollection && $identifier->isEmpty()) {
-            $identifier = new IdentifierCollection(['Authentication.Token']);
+        if ($identifier === null) {
+            $identifier = IdentifierFactory::create('Authentication.Token');
         }
```
ADmad:
```suggestion
        $identifier ??= IdentifierFactory::create('Authentication.Token');
```

dereuromark:
Auth code looks pretty clean and cake-user-friendly now, doesnt it?

dereuromark:
@ADmad How do we merge our regression fix into this?
Do we want to keep the "cleaner" constructor of 4.x? Or do we move to the lazy defaulting?

ADmad:
Doing it in `getIdentifier()` does have the benefit that the objects chain gets initialized only when needed.

src/Authenticator/JwtAuthenticator.php
```diff
@@ -71,6 +67,22 @@ public function __construct(?IdentifierInterface $identifier, array $config = []
...
+     *
+     * @return \Authentication\Identifier\IdentifierInterface
+     */
+    public function getIdentifier(): IdentifierInterface
+    {
+        if ($this->_identifier === null) {
+            $this->_identifier = IdentifierFactory::create('Authentication.JwtSubject');
+        }
+
+        return $this->_identifier;
```
ADmad:
```suggestion
        return $this->_identifier ??= IdentifierFactory::create('Authentication.JwtSubject');
```

src/Authenticator/TokenAuthenticator.php
```diff
@@ -38,17 +38,19 @@ class TokenAuthenticator extends AbstractAuthenticator implements StatelessInter
...
+    public function getIdentifier(): IdentifierInterface
     {
-        // If no identifier is configured, set up a default Token identifier
-        $identifier ??= IdentifierFactory::create('Authentication.Token');
+        if ($this->_identifier === null) {
+            $this->_identifier = IdentifierFactory::create('Authentication.Token');
+        }
 
-        parent::__construct($identifier, $config);
+        return $this->_identifier;
```
ADmad:
```suggestion
        return $this->_identifier ??= IdentifierFactory::create('Authentication.Token');
```

markstory (commented):
Looking good. I think the impacts of these breaking changes should be a reasonable amount of work in a small section for most applications. I don't think we need to worry about rector here.  I like how the authenticator + identifier relationship has ended up working.

dereuromark:
Just to be clear:

Config needs to change from:
```php
  $identifier = [
      'Authentication.Token' => [
          'tokenField' => 'id',
          'dataField' => 'key',
          'resolver' => [
              'finder' => 'auth',
          ],
      ],
  ];
```
  To:
```php
  $identifier = [
      'className' => 'Authentication.Token',
      'tokenField' => 'id',
      'dataField' => 'key',
      'resolver' => [
          'finder' => 'auth',
      ],
  ];
```

We could shim this (allow for the old array one), but then it would be possible to also insert multiple array key/value pairs or alike.

ADmad:
I think just clarifying the change in a migration guide should suffice

dereuromark:
Some things are auto-upgradable: https://github.com/cakephp/upgrade/pull/370

Shall we start the release process on this major?

ADmad:
> Shall we start the release process on this major?

Since we are doing a new major I think we can make this a "proper" CakePHP plugin, i.e. add `cakephp/cakephp` to composer's `require` config. I highly doubt anyone is using the auth plugin outside of Cake.

If we do so then some housekeeping that can be done is, rename `GenericUrlChecker` to `StringUrlChecker` (and update it's docblock description). Similarly change/cleanup any other code which was meant for framework agnostic usage (can't think of anything off the top of my head though).

dereuromark:
Technically, https://github.com/cakephp/authorization/blob/459cd4d752b3b93ff49791b33ae42a04b82cceb4/composer.json#L27 also only has one package

Couldnt we do the housekeeping for it anyways? and document it as such?

ADmad:
Yes we can do the same for Authorization too.

dereuromark:
This PR is only for review, we dont merge this one, correct? As we directly release on 4.x branch

LordSimal:
Correct, 4.x will stay as it is and the new release will be done on that branch

dereuromark:
@ADmad merged 3.x and did the cleanup.
If there is nothing else left feel free to release/tag.

ADmad:
> If there is nothing else left feel free to release/tag.

@dereuromark Nothing comes to mind. I think we are ready to release 4.0.0

dereuromark:
@ADmad I will make a CakePHP Authentication 3.3.5 release first.

dereuromark:
https://github.com/cakephp/authentication/releases/tag/4.0.0

PR #750, opened 2025-11-20
Rename Plugin to AuthenticationPlugin

ADmad:


dereuromark:
Shouldn't this be in 4.x major?
People can load plugins using the fcqn

Alternative is deprecation and extending New one

ADmad:
> Alternative is deprecation and extending New one

That's a better option, I'll update

dereuromark (approved):

Issue #751, opened 2025-11-20
[Security] Add redirect loop protection to AuthenticationService


dereuromark:
# [Security] Add redirect loop protection to AuthenticationService

## Summary

The `AuthenticationService::getLoginRedirect()` method is vulnerable to redirect loop attacks where malicious actors or misconfigured bots can create deeply nested redirect chains that waste server resources and potentially enable security exploits.

## Current Behavior

When an unauthenticated user tries to access a protected page, the authentication middleware redirects them to the login page with a `redirect` query parameter containing the original URL:

```
GET /protected/page → 302 Redirect to /login?redirect=/protected/page
```

However, the `getLoginRedirect()` method (lines 433-461 in `src/AuthenticationService.php`) only validates that the redirect URL:
- Is not empty
- Doesn't contain a host or scheme (prevents external redirects)
- Starts with `/`

**It does NOT check for:**
- ❌ Recursive redirect parameters
- ❌ Multiple encoding levels
- ❌ Redirecting back to authentication-related pages

## Vulnerability Example

A bot or attacker can create URLs like:
```
/login?redirect=/login?redirect=/login?redirect=/protected/page
```

Which when URL-encoded becomes:
```
/login?redirect=%2Flogin%3Fredirect%3D%252Flogin%253Fredirect%253D%25252Fprotected%25252Fpage
```

Each time the system processes this, it creates another level of nesting, potentially leading to:
1. **Resource exhaustion** - Server wastes CPU parsing deeply nested URLs
2. **Log pollution** - Malformed URLs flood access logs
3. **SEO damage** - Search engines index login pages with redirect loops
4. **Potential exploits** - Could be combined with other vulnerabilities

## Real-World Evidence

From production access logs (GPTBot crawling a CakePHP site):
```
74.7.243.239 - - [20/Nov/2025:13:23:46 +0000] "GET /login?redirect=%2Flogin%3Fredirect%3D%252Flogin%253Fredirect%253D%25252Flogin%25253Fredirect%25253D%2525252Flogin%2525253Fredirect%2525253D%252525252Flogin%252525253Fredirect%252525253D%25252525252Fcommunity%25252525252Fstories..." 200 8802 "-" "Mozilla/5.0 AppleWebKit/537.36 (KHTML, like Gecko; compatible; GPTBot/1.3; +https://openai.com/gptbot)"
```

This shows **6-7 levels of nested redirects** being actively exploited by AI crawlers.

## Proposed Solution

[...]

PR #752, opened 2025-11-20
Add optional redirect loop protection to AuthenticationService

dereuromark:
## Summary

This PR adds optional redirect validation to `AuthenticationService::getLoginRedirect()` to prevent redirect loop attacks that have been observed in production environments.

## Problem

The current implementation validates that redirect URLs are relative (not external) but does NOT check for:
- ❌ Nested redirect parameters
- ❌ Multiple encoding levels  
- ❌ Redirects to authentication pages (causing loops)
- ❌ Excessively long URLs (DOS prevention)

**Real-world evidence:** Production logs show bots (especially GPTBot) creating 6-7 levels of nested redirects, wasting server resources and potentially enabling security exploits.

## Solution

Adds configurable `redirectValidation` option with:
- **Opt-in design**: Disabled by default for backward compatibility
- **Four validation checks**: Redirect depth, encoding levels, URL length, blocked patterns
- **Extensible**: Protected `validateRedirect()` method can be overridden
- **Well-tested**: 8 new test cases, all existing tests pass (312 total)
- **Fully documented**: New documentation page with security considerations

## Configuration Example

```php
$service->setConfig([
    'queryParam' => 'redirect',
    'redirectValidation' => [
        'enabled' => true,
        'maxDepth' => 1,
        'maxEncodingLevels' => 1,
        'maxLength' => 2000,
        'blockedPatterns' => ['#/login#i', '#/logout#i'],
    ],
]);
```

## Changes

- ✅ Add `validateRedirect()` protected method to `AuthenticationService`
- ✅ Add `redirectValidation` configuration with sensible defaults
- ✅ Update `getLoginRedirect()` to call validation
- ✅ Add 8 comprehensive test cases (all pass)
- ✅ Add detailed documentation (`docs/en/redirect-validation.rst`)
- ✅ Maintain 100% backward compatibility (disabled by default)
- ✅ All existing tests pass (312 tests, 926 assertions)
- ✅ Code style checks pass

## Validation Logic

1. **Redirect depth**: Count `redirect=` occurrences in decoded URL
2. **Encoding level**: Count `%25` (percent-encoding of %)
3. **URL length**: Check total character count
4. **Blocked patterns**: Match against configured regex patterns

If validation fails, returns `null` instead of invalid URL.

## Security Impact

**Before:**
- ⚠️ Nested redirects like `/login?redirect=/login?redirect=/login...`
- ⚠️ Double-encoded URLs like `%252F`
- ⚠️ Redirects to `/logout` causing loops
- ⚠️ Excessively long URLs (potential DOS)

**After (when enabled):**
- ✅ All nested redirects blocked
- ✅ Deep encoding detected and rejected
- ✅ Auth page loops prevented
- ✅ URL length limited

## References

- Issue: #751
- OWASP: [Unvalidated Redirects and Forwards](https://owasp.org/www-community/attacks/Unvalidated_Redirects_and_Forwards)
- CWE-601: [URL Redirection to Untrusted Site](https://cwe.mitre.org/data/definitions/601.html)

## Testing

```bash
# Run new tests
vendor/bin/phpunit --filter testGetLoginRedirect

# Run full test suite
vendor/bin/phpunit

# Check code style
vendor/bin/phpcs --standard=vendor/cakephp/cakephp-codesniffer/CakePHP src/
```

## Checklist

- [x] Tests pass
- [x] Code style checks pass  
- [x] Documentation added
- [x] Backward compatible (opt-in)
- [x] Security considerations documented

---

**Ready for review!** Happy to make any adjustments based on feedback.

### DIFF
--- a/src/AuthenticationService.php
+++ b/src/AuthenticationService.php
@@ -93,10 +93,6 @@ class AuthenticationService implements AuthenticationServiceInterface, Impersona
      *     'maxDepth' => 1,                // Max nested "redirect=" parameters (default: 1)
      *     'maxEncodingLevels' => 1,       // Max percent-encoding levels (default: 1)
      *     'maxLength' => 2000,            // Max URL length in characters (default: 2000)
-     *     'blockedPatterns' => [          // Regex patterns to reject (default: auth pages)
-     *         '#/login#i',
-     *         '#/logout#i',
-     *     ],
      * ]
      * ```
      *
@@ -126,11 +122,6 @@ class AuthenticationService implements AuthenticationServiceInterface, Impersona
             'maxDepth' => 1,
             'maxEncodingLevels' => 1,
             'maxLength' => 2000,
-            'blockedPatterns' => [
-                '#/login#i',
-                '#/logout#i',
-                '#/register#i',
-            ],
         ],
     ];
 
@@ -524,13 +515,6 @@ class AuthenticationService implements AuthenticationServiceInterface, Impersona
         // Check URL length to prevent DOS attacks
         if (strlen($redirect) > $config['maxLength']) {
             return null;
-        }
-
-        // Check against blocked patterns
-        foreach ($config['blockedPatterns'] as $pattern) {
-            if (preg_match($pattern, $decodedUrl)) {
-                return null;
-            }
         }
 
         return $redirect;
--- a/src/AuthenticationService.php
+++ b/src/AuthenticationService.php
@@ -503,7 +503,7 @@ class AuthenticationService implements AuthenticationServiceInterface, Impersona
         $config = $this->getConfig('redirectValidation');
 
         // If validation is disabled, return the URL as-is (backward compatibility)
-        if (empty($config['enabled'])) {
+        if (!$config['enabled']) {
             return $redirect;
         }
 
@@ -511,13 +511,13 @@ class AuthenticationService implements AuthenticationServiceInterface, Impersona
 
         // Check for nested redirect parameters
         $redirectCount = substr_count($decodedUrl, 'redirect=');
-        if ($redirectCount > $config['maxDepth']) {
+        if ($redirectCount >= $config['maxDepth']) {
             return null;
         }
 
         // Check for multiple encoding levels (e.g., %25 = percent-encoded %)
         $encodingCount = substr_count($redirect, '%25');
-        if ($encodingCount > $config['maxEncodingLevels']) {
+        if ($encodingCount >= $config['maxEncodingLevels']) {
             return null;
         }
 
--- a/src/AuthenticationService.php
+++ b/src/AuthenticationService.php
@@ -83,6 +83,22 @@ class AuthenticationService implements AuthenticationServiceInterface, Impersona
      *    AuthenticationComponent::allowUnauthenticated()
      * - `queryParam` - The name of the query string parameter containing the previously blocked URL
      *   in case of unauthenticated redirect, or null to disable appending the denied URL.
+     * - `redirectValidation` - Configuration for validating redirect URLs to prevent loops. See below.
+     *
+     * ### Redirect Validation Configuration:
+     *
+     * ```
+     * 'redirectValidation' => [
+     *     'enabled' => true,              // Enable validation (default: false for BC)
+     *     'maxDepth' => 1,                // Max nested "redirect=" parameters (default: 1)
+     *     'maxEncodingLevels' => 1,       // Max percent-encoding levels (default: 1)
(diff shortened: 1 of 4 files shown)


src/AuthenticationService.php
```diff
@@ -457,7 +484,56 @@ public function getLoginRedirect(ServerRequestInterface $request): ?string
...
+     *
+     * @param string $redirect The redirect URL to validate
+     * @return string|null The validated URL or null if invalid
+     */
+    protected function validateRedirect(string $redirect): ?string
+    {
+        $config = $this->getConfig('redirectValidation');
+
+        // If validation is disabled, return the URL as-is (backward compatibility)
+        if (empty($config['enabled'])) {
```
ADmad:
```suggestion
        if (!$config['enabled']) {
```

dereuromark:
Thanks for the quick review @ADmad and @Copilot!

I've addressed all the feedback in commit 1ae39c2:

✅ **Fixed comparison operators** (>= instead of >)
- Changed `redirectCount > maxDepth` to `>=`
- Changed `encodingCount > maxEncodingLevels` to `>=`
- This correctly blocks when the count meets or exceeds the threshold

✅ **Replaced empty() with !**
- Changed `empty($config['enabled'])` to `!$config['enabled']` for clearer intent

All tests still pass (312 tests, 926 assertions). Ready for another review!

src/AuthenticationService.php
```diff
@@ -105,6 +121,17 @@ class AuthenticationService implements AuthenticationServiceInterface, Impersona
...
+        'redirectValidation' => [
+            'enabled' => false, // Disabled by default for backward compatibility
+            'maxDepth' => 1,
+            'maxEncodingLevels' => 1,
+            'maxLength' => 2000,
+            'blockedPatterns' => [
+                '#/login#i',
+                '#/logout#i',
+                '#/register#i',
+            ],
```
markstory: This is a lot of complexity to handle URL validation.

markstory (commented):
This whole thing seems quite complex for the problem it is trying to solve.

src/AuthenticationService.php
dereuromark: I am all in for simpler detection of course, what parts should we drop from the default handling here?
the patterns?

dereuromark: The validateRedirect() is quite simple itself, though.

ADmad: The `blockedPatterns` could be dropped.

dereuromark:
Thanks for the feedback @markstory and @ADmad!

I've simplified the implementation in commit 591ec19:

## Removed
❌ **blockedPatterns configuration** - Removed entirely
❌ **Pattern-based validation logic** - Removed from core implementation
❌ **2 pattern-based tests** - Removed

## What Remains (Core Security Features)
✅ **Nested redirect detection** - Blocks `/login?redirect=/login?redirect=...`
✅ **Deep encoding detection** - Blocks double-encoded URLs like `%252F`
✅ **URL length limits** - Prevents DOS via long URLs

## Custom Pattern Blocking
Applications that need pattern-based blocking (e.g., blocking redirects to `/login`, `/logout`) can extend `AuthenticationService` and override `validateRedirect()`. I've updated the documentation with an example.

---

**Result:** Much simpler implementation focused on the core redirect loop vulnerability.

- **310 tests pass** (920 assertions)
- **Code style checks pass**
- **Documentation updated** with custom validation example

Ready for another review!

dereuromark:
Gonna test it on the server before we tag any release.

dereuromark:
So far so good.

Only one more case where it can run into loops. when working together with authorization - and UnauthorizedHandler.
It seems there is one important part missing that gets triggered when a user logged in has not sufficient rights, when you there try to send a user to the action they wanted to go.

This seems to solve it, but its not really too nice.
```php
// SafeRedirectHandler extends plugin one
	public function handle(Exception $exception, ServerRequestInterface $request, array $options = []): ResponseInterface {
		// If user is already logged in but not authorized, redirect to fallback
		// instead of login page (which would cause a loop)
		$identity = $request->getAttribute('identity');
		if ($identity) {
			$message = $options['unauthorizedMessage'] ?? __('You are not authorized to access that location.');
			if ($message) {
				/** @var \Cake\Http\ServerRequest $request */
				$session = $request->getSession();
				$session->write('Flash.flash', [
					[
						'message' => $message,
						'key' => 'flash',
						'element' => 'flash/error',
						'params' => [],
					],
				]);
			}

			$fallbackUrl = Router::url(['prefix' => false, 'plugin' => false, 'controller' => 'Account', 'action' => 'index']);

			return (new Response())
				->withHeader('Location', $fallbackUrl)
				->withStatus(302);
		}
```

PR #755, opened 2025-11-24
Fix loadIdentifier called after loadAuthenticator losing resolver config

### LINKED ISSUE #754: [Regression 3.3.3] OrmResolver config (userModel/finder) is ignored when nested under 'resolver' in PasswordIdentifier
I'm reporting a regression found in the cakephp/authentication plugin, specifically version 3.3.3. Version 3.3.2 works as expected.

The configuration for the OrmResolver is correctly passed to AuthenticationService::loadIdentifier(), but the nested keys ('userModel', 'finder') are stripped or lost when the PasswordIdentifier is initialized, causing the OrmResolver to fall back to its default model, 'Users', instead of the intended one, e.g., 'Members'.

This prevents any non-default user model from being authenticated.

The following configuration, designed to use the 'Members' model:

```
$service->loadIdentifier('Authentication.Password', [
    'resolver' => [
        'userModel' => 'Members', // This configuration is lost in 3.3.3
        'finder' => 'all',
    ],
    'fields' => ['username' => 'email', 'password' => 'password'],
]);
```

Output from AuthenticationService::loadIdentifier() (Configuration is correct):

```
'Authentication.Password'
[
    'resolver' => [
        'userModel' => 'Members', // Correctly present here
        'finder' => 'all',
    ],
    // ...
]
```

Output from OrmResolver::__construct() (Configuration is lost):

```
// Contents of $config array received by OrmResolver::__construct()
[
    'className' => 'Authentication.Orm', // Only this key remains
]
```

CakePHP version: 5.2.9
Authentication version: 3.3.3 (not working)
Authentication version: 3.3.2 (working)

dereuromark:
## Summary

Fixes #754

When `loadAuthenticator()` was called before `loadIdentifier()`, the resolver config (including `userModel` and `finder`) would be lost because:

1. `loadAuthenticator()` triggers `authenticators()` → `identifiers()` which creates an empty `IdentifierCollection`
2. The authenticator (e.g., `FormAuthenticator`) would see the empty collection and create a **new** default `PasswordIdentifier` with default resolver config
3. Later calls to `loadIdentifier()` would add to the **service's** collection, but the authenticator already had its own **separate** collection

### DIFF
--- a/src/Authenticator/FormAuthenticator.php
+++ b/src/Authenticator/FormAuthenticator.php
@@ -50,26 +50,23 @@ class FormAuthenticator extends AbstractAuthenticator
     ];
 
     /**
-     * Constructor
+     * Gets the identifier, loading a default Password identifier if none configured.
      *
-     * @param \Authentication\Identifier\IdentifierInterface $identifier Identifier or identifiers collection.
-     * @param array<string, mixed> $config Configuration settings.
+     * This is done lazily to allow loadIdentifier() to be called after loadAuthenticator().
+     *
+     * @return \Authentication\Identifier\IdentifierInterface
      */
-    public function __construct(IdentifierInterface $identifier, array $config = [])
+    public function getIdentifier(): IdentifierInterface
     {
-        // If no identifier is configured, set up a default Password identifier
-        if ($identifier instanceof IdentifierCollection && $identifier->isEmpty()) {
-            // Pass the authenticator's fields configuration to the identifier
+        if ($this->_identifier instanceof IdentifierCollection && $this->_identifier->isEmpty()) {
             $identifierConfig = [];
-            if (isset($config['fields'])) {
-                $identifierConfig['fields'] = $config['fields'];
+            if ($this->getConfig('fields')) {
+                $identifierConfig['fields'] = $this->getConfig('fields');
             }
-            $identifier = new IdentifierCollection([
-                'Authentication.Password' => $identifierConfig,
-            ]);
+            $this->_identifier->load('Authentication.Password', $identifierConfig);
         }
 
-        parent::__construct($identifier, $config);
+        return $this->_identifier;
     }
 
     /**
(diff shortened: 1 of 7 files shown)


dereuromark:
The fix is quite heavy, as it would need to move the defaulting out of the constructor.
We have 3 options:
- document the loading order
- fix the bc break
- revert defaulting and only apply to new major

src/Authenticator/FormAuthenticator.php
```diff
@@ -50,26 +49,43 @@ class FormAuthenticator extends AbstractAuthenticator
...
+     * to find POST data that is used to find a matching record in the `config.userModel`. Will return false if
+     * there is no post data, either username or password is missing, or if the scope conditions have not been met.
      *
-     * @param \Authentication\Identifier\IdentifierInterface $identifier Identifier or identifiers collection.
-     * @param array<string, mixed> $config Configuration settings.
+     * @param \Psr\Http\Message\ServerRequestInterface $request The request that contains login information.
+     * @return \Authentication\Authenticator\ResultInterface
      */
-    public function __construct(IdentifierInterface $identifier, array $config = [])
+    public function authenticate(ServerRequestInterface $request): ResultInterface
```
markstory: Is there a reason we can't use `setIdentifier()`/`getIdentifier()`? It seems it could be a simpler path forward.

If when we need the identifier, and the collection is empty, authenticators could create their default. We could even have a hook method on `AbstractAuthenticator` if you wanted to get fancy.

ADmad: > If when we need the identifier, and the collection is empty, authenticators could create their default.

That's what's being done below isn't it? Just that the `_identifier` property is used directly instead of using `setIdentifer()`/`getIdentifier()`.

dereuromark:
We need to squash merge this, and then for merging to 4.x we probably want to go back to cleaner constructor way.

src/Authenticator/FormAuthenticator.php
```diff
@@ -68,6 +66,19 @@ public function authenticate(ServerRequestInterface $request): ResultInterface
             $this->_identifier->load('Authentication.Password', $identifierConfig);
         }
 
+        return $this->_identifier;
+    }
+
+    /**
+     * Authenticates the identity contained in a request. Will use the `config.userModel`, and `config.fields`
+     * to find POST data that is used to find a matching record in the `config.userModel`. Will return false if
+     * there is no post data, either username or password is missing, or if the scope conditions have not been met.
```
ADmad:
```suggestion
     * Authenticates the identity contained in a request.
     * 
     * Will use the `config.userModel`, and `config.fields`
     * to find POST data that is used to find a matching record in the `config.userModel`. Will return false if
     * there is no post data, either username or password is missing, or if the scope conditions have not been met.
```

src/Authenticator/FormAuthenticator.php
ADmad: After the updates done by dereuromark I see now what you meant :)

dereuromark: Do you want to make an alternative PR or adjust this one?
Or all good now?

ADmad (approved):

PR #763, opened 2026-01-13
Deprecate SessionAuthenticator identify option

dereuromark:
## Summary

- Deprecate the `identify` option in SessionAuthenticator
- Simplify the Quick Start example to not pass unnecessary identifier to SessionAuthenticator

### Why deprecate `identify`?

When `identify => true`:
- The default `fields` config only includes `username`, so only the username is passed to the identifier
- The `PasswordIdentifier::identify()` method skips password verification when no password credential is provided
- Even if you add `password` to fields, it fails because sessions store hashed passwords, but `password_verify(hash, hash)` doesn't work

This makes the option ineffective for:
- Detecting password changes
- Remotely invalidating sessions when passwords are updated

Users should use `PrimaryKeySessionAuthenticator` instead if they need to fetch fresh user data from the database on each request.

### Quick Start example improvement

The SessionAuthenticator doesn't need an identifier by default - it just uses session data directly as identity. The example was unnecessarily passing `$passwordIdentifier` to it.

### DIFF
--- a/src/Authenticator/SessionAuthenticator.php
+++ b/src/Authenticator/SessionAuthenticator.php
@@ -21,6 +21,7 @@ use Authentication\Identifier\AbstractIdentifier;
 use Cake\Http\Exception\UnauthorizedException;
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\ServerRequestInterface;
+use function Cake\Core\deprecationWarning;
 
 /**
  * Session Authenticator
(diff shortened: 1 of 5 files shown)


src/Authenticator/SessionAuthenticator.php
```diff
@@ -31,9 +32,10 @@ class SessionAuthenticator extends AbstractAuthenticator implements PersistenceI
      * Default config for this object.
      * - `fields` The fields to use to verify a user by.
      * - `sessionKey` Session key.
-     * - `identify` Whether to identify user data stored in a session. This is
-     *   useful if you want to remotely end sessions that have a different password stored,
-     *   or if your identification logic needs additional conditions before a user can login.
+     * - `identify` Whether to identify user data stored in a session.
+     *   Deprecated: This option only verifies that the username exists in the database,
+     *   it does not verify passwords. Use `PrimaryKeySessionAuthenticator` instead if you
+     *   need to fetch fresh user data from the database on each request.
```
markstory: Why call out that passwords aren't verified? The PrimaryKeySessionAuthenticator isn't checking passwords either.

dereuromark: identify sounded like some verification.
but we can drop that sentence of course.

dereuromark: The main reason:

The identify option only verifies that the username exists in the database - it does not verify passwords as the documentation previously suggested.
So this was more to rectify the wrong docs from before.

Removing that sentence now ,though.

PR #764, opened 2026-01-14
Add default TokenIdentifier for PrimaryKeySessionAuthenticator

dereuromark:
## Summary

`PrimaryKeySessionAuthenticator` now provides a default `TokenIdentifier` that works out of the box, eliminating the need for explicit configuration in most cases.

### The Problem

Currently, using `PrimaryKeySessionAuthenticator` requires understanding the internal field mapping between the authenticator and identifier:

```php
$service->loadAuthenticator('Authentication.PrimaryKeySession', [
    'identifier' => [
        'className' => 'Authentication.Token',
        'tokenField' => 'id',
        'dataField' => 'key',
    ],
]);
```

The field names don't align intuitively:
- `identifierKey` (authenticator) ↔ `dataField` (identifier)
- `idField` (authenticator) ↔ `tokenField` (identifier)

### The Solution (Option 1 - Implemented)

Override `getIdentifier()` to lazily create a default `TokenIdentifier` with matching configuration:

```php
// Now works without any configuration!
$service->loadAuthenticator('Authentication.PrimaryKeySession');
```

Custom configuration is still fully supported when needed.

### Alternative Approaches Considered

**Option 2: Align the naming**

Rename the config keys to use consistent terminology across authenticator and identifier. This would improve discoverability but still require explicit configuration.

**Option 3: Change defaults on PrimaryKeySession side**

Change `identifierKey` default from `'key'` to `'id'`, reducing the configuration needed. However, this still requires understanding which fields to configure.

### Why Option 1?

- Zero configuration for the common case (looking up users by `id`)
- Follows the existing pattern in `SessionAuthenticator` (which provides a default `PasswordIdentifier`)
- Custom identifiers still work exactly as before
- The `idField` and `identifierKey` config options automatically propagate to the default identifier

## Test Plan

- [x] Existing tests still pass
- [x] New test: `testAuthenticateSuccessWithDefaultIdentifier` - verifies authentication works without explicit identifier
- [x] New test: `testGetIdentifierReturnsDefaultWhenNotConfigured` - verifies default TokenIdentifier is created
- [x] New test: `testGetIdentifierUsesCustomConfig` - verifies custom `idField`/`identifierKey` propagates correctly

### DIFF
--- a/docs/en/authenticators.rst
+++ b/docs/en/authenticators.rst
@@ -37,22 +37,33 @@ It also helps to avoid session invalidation.
 Session itself stores the entity object including nested objects like DateTime or enums.
 With only the ID stored, the invalidation due to objects being modified will also dissolve.
 
-Make sure to match this with a Token identifier with ``key``/``id`` keys::
+A default ``TokenIdentifier`` is provided that looks up users by their ``id`` field,
+so minimal configuration is required::
+
+    $service->loadAuthenticator('Authentication.PrimaryKeySession');
+
+Configuration options:
+
+-  **idField**: The field in the database table to look up. Default is ``id``.
+-  **identifierKey**: The key used to store/retrieve the primary key from session data.
+   Default is ``key``.
+
+For custom lookup fields, the ``idField`` and ``identifierKey`` options propagate
+to the default identifier automatically::
+
+    $service->loadAuthenticator('Authentication.PrimaryKeySession', [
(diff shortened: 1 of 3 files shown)


ADmad:
LGTM. The docs would also have to be updated to mention the default identifier config.

dereuromark:
@ADmad Done

PR #787, opened 2026-04-16
Add redirectAfterLogin() helper

dereuromark:
## Summary

This adds a small controller-level helper to make the safe post-login redirect path the obvious one:

```php
return $this->Authentication->redirectAfterLogin('/dashboard');
```

Internally this builds on the existing safe redirect parsing already present in the plugin via `AuthenticationService::getLoginRedirect()` and `AuthenticationComponent::getLoginRedirect()`.

## Why

The plugin already does the hard part correctly:

- `AuthenticationService::getLoginRedirect()` rejects redirect targets with a `scheme` or `host`
- it normalizes local paths
- it applies optional redirect-loop validation

The gap is ergonomics.

In real applications, controller code often ends up doing this instead:

```php
$redirect = $this->request->getQuery('redirect', '/dashboard');
return $this->redirect($redirect);
```

or this:

```php
$redirect = $this->Authentication->getLoginRedirect('/dashboard') ?? '/dashboard';
return $this->redirect($redirect);
```

The second version is safe, but it is a two-step pattern that is easy to bypass or forget. A first-class helper makes the secure path shorter and more discoverable.

## What this adds

This PR adds:

```php
AuthenticationComponent::redirectAfterLogin(array|string|null $default = '/')
```

Behavior:

1. Resolve the validated local redirect target using existing logic: `getLoginRedirect($default)`
2. Fallback to `$default` when no valid redirect is available
3. Return the controller redirect response directly

Example:

```php
if ($result->isValid()) {
    return $this->Authentication->redirectAfterLogin('/dashboard');
}
```

We could release this as a new minor.

### DIFF
--- a/src/Controller/Component/AuthenticationComponent.php
+++ b/src/Controller/Component/AuthenticationComponent.php
@@ -379,15 +379,12 @@ class AuthenticationComponent extends Component implements EventDispatcherInterf
      * controller's redirect method so applications can use the plugin's
      * existing safe redirect parsing without manually reading query params.
      *
-     * @param array|string|null $default Default URL to use when no valid login redirect is available.
+     * @param array|string $default Default URL to use when no valid login redirect is available.
      * @return \Cake\Http\Response|null
      */
-    public function redirectAfterLogin(array|string|null $default = '/'): ?Response
+    public function redirectAfterLogin(array|string $default = '/'): ?Response
     {
         $target = $this->getLoginRedirect($default) ?? $default;
-        if ($target === null) {
-            return null;
-        }
 
         return $this->getController()->redirect($target);
     }
--- a/src/Controller/Component/AuthenticationComponent.php
+++ b/src/Controller/Component/AuthenticationComponent.php
@@ -29,6 +29,7 @@ use Authentication\IdentityInterface;
 use Cake\Controller\Component;
 use Cake\Event\EventDispatcherInterface;
 use Cake\Event\EventDispatcherTrait;
+use Cake\Http\Response;
 use Cake\Routing\Router;
 use Cake\Utility\Hash;
 use Exception;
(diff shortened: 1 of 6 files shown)


src/Controller/Component/AuthenticationComponent.php
```diff
@@ -370,6 +371,27 @@ public function getLoginRedirect(array|string|null $default = null): ?string
...
+     * target from the current request when available.
+     *
+     * This is a convenience wrapper around `getLoginRedirect()` plus the
+     * controller's redirect method so applications can use the plugin's
+     * existing safe redirect parsing without manually reading query params.
+     *
+     * @param array|string|null $default Default URL to use when no valid login redirect is available.
+     * @return \Cake\Http\Response|null
+     */
+    public function redirectAfterLogin(array|string|null $default = '/'): ?Response
```
markstory: Why would one use `null`?

tests/TestCase/Controller/Component/AuthenticationComponentTest.php
```diff
@@ -397,6 +397,84 @@ public function testGetLoginRedirect(): void
...
+            ->withAttribute('identity', $this->identity)
+            ->withAttribute('authentication', $this->service)
+            ->withQueryParams(['redirect' => '//evil.example/phish']);
+
+        $controller = new Controller($request);
+        $registry = new ComponentRegistry($controller);
+        $component = new AuthenticationComponent($registry);
+
+        $response = $component->redirectAfterLogin($url);
+        $this->assertSame('/dashboard', $response?->getHeaderLine('Location'));
```
markstory, line 475: If we're going to have a `null` parameter lets have a test for it. From my read, it looks like a scenario that shouldn't exist.

src/Controller/Component/AuthenticationComponent.php
ADmad: I agree, don't see the need for a `null` argument or return value.

dereuromark:
Yeah, sounds right.
I adjusted it.

PR #788, opened 2026-04-28
Add AuthenticationComponent::replaceIdentity()

dereuromark:
## Summary

Adds `AuthenticationComponent::replaceIdentity()` for swapping the in-request identity attribute without going through `clearIdentity()` / `persistIdentity()`.

## Motivation

A common pattern in apps is to enrich the active user with eager-loaded associations or computed flags in `AppController::beforeFilter()`. Doing this with `setIdentity()` has a surprising side effect: `setIdentity()` calls `clearIdentity()` first, and the service's `clearIdentity()` actively calls `stopImpersonating()` on impersonation-aware authenticators (`AuthenticationService.php` lines 192-200). So a pattern like

```php
if ($user && !$user->some_association) {
    $reloaded = $this->Users->get($user->id, finder: 'fullProfile');
    $this->Authentication->setIdentity($reloaded);
}
```

silently ends impersonation on every request where the association was missing.

`replaceIdentity()` solves this by writing only to the request attribute - no session writes, no `clearIdentity` chain, no impersonation interaction. It mirrors what `withAttribute('identity', new Identity($data))` does, but routes through the service so the configured `identityClass` is honored.

## Changes

- `AuthenticationComponent::replaceIdentity(ArrayAccess|array $identity)` - new method
- `AuthenticationServiceInterface::buildIdentity()` - lifted to the interface so the component can call it without a concrete-type cast (the method already existed on `AuthenticationService`)
- 3 unit tests: array data, identity instance, and the impersonation-survival case

### DIFF
--- a/docs/en/authentication-component.md
+++ b/docs/en/authentication-component.md
@@ -102,6 +102,60 @@ The result returned will contain an array like this:
 > context you're working in you'll have to use these instances from now on if you
 > want to continue to work with the modified response and request objects.
 
+## Replacing the current identity
+
+Use `setIdentity()` to change which user is logged in (e.g. after registration
+or social-login first-touch). It clears all persisted identity data and writes
+the new identity through every persisting authenticator:
+
+```php
+$this->Authentication->setIdentity($user);
+```
+
+> [!WARNING]
+> `setIdentity()` ends an active impersonation session by default, because it
+> goes through `clearIdentity()` first, which calls `stopImpersonating()` on
+> impersonation-aware authenticators. See the two methods below for the
+> non-default cases.
+
+### Refresh the active identity for the current request only
+
+When you only need to swap the in-request identity (for example to attach
+eager-loaded associations or computed flags in `beforeFilter()`) without
+touching the session or persistence, use `replaceIdentity()`:
+
+```php
+// AppController::beforeFilter()
+$identity = $this->Authentication->getIdentity();
+if ($identity && !$identity->some_association) {
+    $reloaded = $this->fetchTable('Users')
+        ->get($identity->getIdentifier(), finder: 'fullProfile');
+    $this->Authentication->replaceIdentity($reloaded);
+}
+```
+
+This rewrites only the request attribute. The session is not touched, so an
+active impersonation is preserved and no privilege-escalation side effects
+(like session rotation) occur.
+
+### Persist a refreshed identity while impersonating
+
+If the refresh has to survive into subsequent requests but you still want to
+keep an active impersonation alive, pass `preserveImpersonation: true` to
+`setIdentity()`:
+
+```php
+$this->Authentication->setIdentity($reloaded, preserveImpersonation: true);
+```
+
+The new identity is persisted into the session as usual, but the
+impersonation slot (`AuthImpersonate`) and the active authenticator are left
+intact. Note that this also skips the session rotation that the default
+`setIdentity()` flow performs - it is a refresh, not a privilege transition,
+so do not use it for login or role changes.
+
+See [User Impersonation](impersonation.md) for the broader context.
+
 ## Configure Automatic Identity Checks
 
 By default `AuthenticationComponent` will automatically enforce an identity to
--- a/docs/en/authentication-component.md
+++ b/docs/en/authentication-component.md
@@ -134,7 +134,7 @@ if ($identity && !$identity->some_association) {
 }
 ```
 
-This rewrites only the request attribute. The session is not touched, so an
+This rewrites only the request attribute. The session is not modified, so an
 active impersonation is preserved and no privilege-escalation side effects
 (like session rotation) occur.
 
--- a/docs/en/authentication-component.md
+++ b/docs/en/authentication-component.md
@@ -115,8 +115,8 @@ $this->Authentication->setIdentity($user);
 > [!WARNING]
 > `setIdentity()` ends an active impersonation session by default, because it
 > goes through `clearIdentity()` first, which calls `stopImpersonating()` on
-> impersonation-aware authenticators. See the two methods below for the
-> non-default cases.
+> impersonation-aware authenticators. Use `replaceIdentity()` below for the
+> request-only refresh case.
 
 ### Refresh the active identity for the current request only
 
@@ -137,22 +137,6 @@ if ($identity && !$identity->some_association) {
 This rewrites only the request attribute. The session is not modified, so an
 active impersonation is preserved and no privilege-escalation side effects
 (like session rotation) occur.
-
-### Persist a refreshed identity while impersonating
-
-If the refresh has to survive into subsequent requests but you still want to
-keep an active impersonation alive, pass `preserveImpersonation: true` to
-`setIdentity()`:
-
-```php
-$this->Authentication->setIdentity($reloaded, preserveImpersonation: true);
-```
-
-The new identity is persisted into the session as usual, but the
-impersonation slot (`AuthImpersonate`) and the active authenticator are left
-intact. Note that this also skips the session rotation that the default
-`setIdentity()` flow performs - it is a refresh, not a privilege transition,
-so do not use it for login or role changes.
 
 See [User Impersonation](impersonation.md) for the broader context.
 
--- a/src/AuthenticationServiceInterface.php
+++ b/src/AuthenticationServiceInterface.php
@@ -16,12 +16,14 @@ declare(strict_types=1);
  */
 namespace Authentication;
 
-use ArrayAccess;
 use Authentication\Authenticator\AuthenticatorInterface;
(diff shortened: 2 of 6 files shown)


src/AuthenticationServiceInterface.php
```diff
@@ -69,6 +70,17 @@ public function getResult(): ?ResultInterface;
...
+    /**
+     * Build an identity object from raw identity data.
+     *
+     * If the supplied data is already an `IdentityInterface`, it is returned
+     * unchanged.
+     *
+     * @param \ArrayAccess<string, mixed>|array<string, mixed> $identityData Identity data.
+     * @return \Authentication\IdentityInterface
+     */
+    public function buildIdentity(ArrayAccess|array $identityData): IdentityInterface;
```
dereuromark: We will have to method annotate this for BC.

LordSimal: Since this is a interface change this can't go into either 4.x or 4.next

ADmad:
The fact that `AuthenticationComponent::setIdentity()` calls `AuthenticationService::clearIdentity()` has bitten me too.

Your proposed solution of `replaceIdentity()` which only updates the request attribute can also be problematic as it doesn't persist the changes. I have a use case where the identity needs to be checked and updated only on specific actions and once it's updated the updates should persist for the subsequent requests.

Maybe we need an option for `AuthenticationComponent::setIdentity()` or a new method which just skips the `AuthenticationService::clearIdentity()` call?  I haven't checked but would preserve any current impersonation?

----------------------------------

Another related issue with using `AuthenticationComponent::setIdentity()` is that it causes problems when you are also using the `Authorization` plugin. During normal flow the `AuthorizationMiddleware` decorates the identity object so that the identity implements the `Authorization\IdentityInterface`.

If you use `AuthenticationComponent::setIdentity()` that decoration is lost and if call any code which expects a `Authorization\IdentityInterface` instance it will generate an error. So we need a `AuthenticationComponent::setIdentity()` too (currently there's only `AuthenticationComponent::buildIdentity()` and then you have to update the request attribute yourself).

src/AuthenticationServiceInterface.php
dereuromark: Yes thus my sentence here - to make it BC

dereuromark:
Pushed an update that addresses both points.

**On the interface BC break (per @LordSimal):**
Reverted the addition of `buildIdentity()` on `AuthenticationServiceInterface` and replaced it with a `@method` docblock annotation on the interface, so the call site stays type-aware without breaking third-party implementers.

**On your "needs to persist for subsequent requests" point:**
You're right that the original `replaceIdentity()` only swaps the in-request attribute and is not enough for the case where the refresh has to survive into the next request - it doesn't rewrite the session, and (as I found while testing) `SessionAuthenticator::persistIdentity()` deliberately won't overwrite an existing `Auth` slot, so just skipping `clearIdentity()` doesn't actually persist either.

So I went with option A: a new flag on `setIdentity()`:

```php
$this->Authentication->setIdentity($reloaded, preserveImpersonation: true);
```

Implementation: `AuthenticationService::clearIdentity()` gets an optional third parameter `bool $stopImpersonation = true`. The interface signature is untouched (PHP allows adding optional params on the implementation), so external `AuthenticationServiceInterface` implementers remain compatible. When set to `false` the impersonation slot survives, but each authenticator's `clearIdentity()` is still called so `persistIdentity()` can write the new identity into the session cleanly. The `_successfulAuthenticator` is also preserved in this path so same-request `isImpersonating()` calls keep working.

The two methods now have distinct intents:
- `replaceIdentity($identity)` - request-only swap, no session write (cheap, ideal for `beforeFilter` enrichment).
- `setIdentity($identity, preserveImpersonation: true)` - persist into session **and** keep impersonation alive (your use case).

New tests cover both the impersonation-preserving path and a regression test confirming the default `setIdentity()` flow still ends impersonation.

**On the `Authorization\IdentityInterface` decoration loss:**
Agreed that's a real adjacent papercut, but I'd prefer to handle it in a separate PR/issue so this one stays focused. Happy to open one if there isn't already.

 Why I deferred it                                                                                                                                                                                         
                                                                         
  There are three possible fixes and none are 1-line:                                                                                                                                                       
   
  1. Make Authentication aware of Authorization. Have AuthenticationComponent::setIdentity() look up an Authorization\IdentityDecorator if the authorization service is on the request and re-wrap. Cleanest
   from the user's POV but creates a soft dependency on the authorization plugin — not great for a plugin that's supposed to work standalone.
  2. Event hook. Dispatch an Authentication.identityReplaced event after writing the new identity attribute. The authorization plugin (or app code) listens and re-decorates. Decoupled, but requires       
  changes in both plugins, and listeners need access to the authorization service.                                                                                                                          
  3. Document the workaround. Tell users they need to manually re-decorate after setIdentity()/replaceIdentity():
  ```php
  $this->Authentication->setIdentity($reloaded);                                                                                                                                                            
  if ($this->request->getAttribute('authorization')) {                   
      $identity = new IdentityDecorator(                                                                                                                                                                    
          $this->request->getAttribute('authorization'),                 
          $this->request->getAttribute('identity'),                                                                                                                                                         
      );                                                                 
      $this->setRequest($this->request->withAttribute('identity', $identity));                                                                                                                              
  }  
  ```                                                                                                                                                                                                       
  3. Cheap to ship, but every consumer has to remember it.

dereuromark:
Anything left?

docs/en/authentication-component.md
```diff
@@ -102,6 +102,60 @@ The result returned will contain an array like this:
...
+// AppController::beforeFilter()
+$identity = $this->Authentication->getIdentity();
+if ($identity && !$identity->some_association) {
+    $reloaded = $this->fetchTable('Users')
+        ->get($identity->getIdentifier(), finder: 'fullProfile');
+    $this->Authentication->replaceIdentity($reloaded);
+}
+```
+
+This rewrites only the request attribute. The session is not touched, so an
```
markstory:
```suggestion
This rewrites only the request attribute. The session is not modified, so an
```

tests/TestCase/Controller/Component/AuthenticationComponentTest.php
```diff
@@ -280,6 +280,173 @@ public function testSetIdentityOverwrite(): void
...
+
+        $component->impersonate($impersonated);
+
+        $reloaded = new ArrayObject(['username' => 'larry', 'profile' => 'loaded']);
+        $component->setIdentity($reloaded);
+
+        $this->assertNull(
+            $controller->getRequest()->getSession()->read('AuthImpersonate'),
+            'Default setIdentity() must end an active impersonation.',
+        );
```
markstory, line 401:
```suggestion
        $component->setIdentity($reloaded);

        $this->assertNull(
            $controller->getRequest()->getSession()->read('AuthImpersonate'),
            'Default setIdentity() must end an active impersonation.',
        );
        $this->assertFalse($component->isImpersonating());
```

markstory (approved):

src/Controller/Component/AuthenticationComponent.php
```diff
@@ -299,15 +300,34 @@ public function getIdentityData(string $path): mixed
...
+     * impersonation.
+     *
      * @param \ArrayAccess|array $identity Identity data to persist.
+     * @param bool $preserveImpersonation Whether to keep an active
+     *  impersonation alive while replacing the identity. Defaults to false
+     *  (existing behavior).
      * @return $this
      */
-    public function setIdentity(ArrayAccess|array $identity)
+    public function setIdentity(ArrayAccess|array $identity, bool $preserveImpersonation = false)
```
ADmad: This change in signature means a new major will be required.

dereuromark: Its an optional arg.
Thus I thought a minor suffices.
Especially since extending Auth component to overwrite these methods is unusual here, no?

ADmad: However small, there is a possibility that this public method could be overridden. We need to maintain BC promise in minor and this signature change would break the promise.

dereuromark: We can sure do v5, but I find it a bit over the top on this specific issue.

dereuromark: In that case an alternative could be a new method then for now.

setIdentityAllowingReplace()
setIdentityPreservingImpersonation()

etc

LordSimal: This PR already adds another method to fix the initial issue via an opt-in adjustment for the users.

As long as its documented we can keep BC here - aka revert this change back to how was - and do this cleanup in the next major.

LordSimal (approved):

LordSimal:
Can confirm that this works to fix my problem with impersonation being forgotten if I swap to this new method.

PR #789, opened 2026-04-28
docs: warn that setIdentity()/clearIdentity() end impersonation

dereuromark:
## Summary

- Adds a 3rd "Limitations" bullet to `docs/en/impersonation.md` explaining that `setIdentity()` and `clearIdentity()` (and therefore `logout()`) silently end impersonation, with a recommended workaround for the common "refresh active user" use case.
- Adds a "Replacing the current identity" section to `docs/en/authentication-component.md` that documents `setIdentity()` (it wasn't documented in this file before) and links the impersonation warning.

## Why

The interaction is real and surprising:

`AuthenticationComponent::setIdentity()` calls the service's `clearIdentity()` first, and `AuthenticationService::clearIdentity()` actively calls `stopImpersonating()` on any `ImpersonationInterface` authenticator that returns `true` for `isImpersonating()` ([AuthenticationService.php#L194](https://github.com/cakephp/authentication/blob/4.x/src/AuthenticationService.php#L194)).

So an app pattern like

```php
// AppController::beforeFilter()
if ($user && !$user->some_association) {
    $reloaded = $this->Users->get($user->id, finder: 'fullProfile');
    $this->Authentication->setIdentity($reloaded);
}
```

silently ends impersonation on every request where the association was missing. The recommended workaround (write to the request attribute directly) is small but isn't currently documented.

## Notes

- Documentation only - no code changes.
- A separate draft PR (#788) proposes a `replaceIdentity()` convenience method that codifies the workaround as first-class API. This docs PR stands on its own and is useful regardless of whether #788 lands; the docs example just uses the manual `withAttribute()` form.

### DIFF
--- a/docs/en/authentication-component.md
+++ b/docs/en/authentication-component.md
@@ -102,6 +102,24 @@ The result returned will contain an array like this:
 > context you're working in you'll have to use these instances from now on if you
 > want to continue to work with the modified response and request objects.
 
+## Replacing the current identity
+
+Use `setIdentity()` to change which user is logged in (e.g. after registration
+or social-login first-touch). It clears all persisted identity data and writes
+the new identity through every persisting authenticator:
+
+```php
+$this->Authentication->setIdentity($user);
+```
+
+> [!WARNING]
+> `setIdentity()` ends an active impersonation session because it goes through
+> `clearIdentity()` first, which calls `stopImpersonating()` on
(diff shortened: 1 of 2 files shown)


dereuromark:
With https://github.com/cakephp/authentication/pull/788 accepted, it would superseed this PR, so draft for now again.

PR #802, opened 2026-06-26
Update AuthenticationComponent.php

bancer:
This ensures that remember_me cookie is cleared after the user that already has this cookie logs in without remember_me checked.


markstory:
A new test case would be nice to have. While the current tests don't fail, what failed before that passes now?

bancer:
I've added tests.

bancer:
@markstory Is there anything that stops this from being merged?

bancer:
TODO: apply the same changes in 3.x and 4.x

PR #808, opened 2026-07-17
Return null from getIdentificationProvider() for identifier-less authenticators

dereuromark:
Fixes #179.

## Problem

On 4.x an authenticator's identifier is optional. `AuthenticatorCollection::create()` passes `null` when no `identifier` is configured, and the session authenticator (default `identify` => false) authenticates purely from stored session data, so it legitimately has no identifier.

`AbstractAuthenticator::getIdentifier()` throws in that case. `AuthenticationService::getIdentificationProvider()` called it unconditionally, so the whole request blew up whenever such an authenticator was the successful one:

```
Identifier is required for `Authentication\Authenticator\SessionAuthenticator`. Please provide an identifier instance.
```

This surfaces for any consumer that reads the identification provider after a session based login (the reporter hit it via the TinyAuth DebugKit panel).

## Fix

- Add `MissingIdentifierException` (extends `RuntimeException` for backwards compatibility) thrown by `AbstractAuthenticator::getIdentifier()` when no identifier is available.
- `getIdentificationProvider()` catches exactly that type and returns `null`.

Catching the specific type rather than a bare `RuntimeException`:

- preserves lazy identifier creation in subclasses (`FormAuthenticator`, `CookieAuthenticator`, `JwtAuthenticator`, `TokenAuthenticator`, `PrimaryKeySessionAuthenticator`, ...) whose overridden `getIdentifier()` builds a default identifier on first call, and
- does not mask genuine identifier configuration failures, which still throw a plain `RuntimeException` from `IdentifierFactory`.

## Backwards compatibility

`MissingIdentifierException extends RuntimeException`, so existing `catch (RuntimeException)` blocks and the message are unchanged.

### DIFF
--- a/src/Authenticator/MissingIdentifierException.php
+++ b/src/Authenticator/MissingIdentifierException.php
@@ -1,0 +1,31 @@
+<?php
+declare(strict_types=1);
+
+/**
+ * CakePHP(tm) : Rapid Development Framework (https://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ *
+ * Licensed under The MIT License
+ * For full copyright and license information, please see the LICENSE.txt
+ * Redistributions of files must retain the above copyright notice.
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ * @link          https://cakephp.org CakePHP(tm) Project
+ * @since         4.0.0
+ * @license       https://opensource.org/licenses/mit-license.php MIT License
+ */
+namespace Authentication\Authenticator;
+
+use RuntimeException;
+
+/**
+ * Thrown when an authenticator is asked for its identifier but none was
+ * configured (and none could be created lazily).
+ *
+ * Extends `RuntimeException` for backwards compatibility with existing catch
+ * blocks; it lets callers distinguish the "no identifier available" case from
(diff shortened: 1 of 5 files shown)


src/Authenticator/MissingIdentifierException.php
```diff
@@ -0,0 +1,31 @@
...
+ * CakePHP(tm) : Rapid Development Framework (https://cakephp.org)
+ * Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ *
+ * Licensed under The MIT License
+ * For full copyright and license information, please see the LICENSE.txt
+ * Redistributions of files must retain the above copyright notice.
+ *
+ * @copyright     Copyright (c) Cake Software Foundation, Inc. (https://cakefoundation.org)
+ * @link          https://cakephp.org CakePHP(tm) Project
+ * @since         4.0.0
```
ADmad: 4.2 or 4.3?

ADmad (approved):
LGTM besides the version nitpick.

src/Authenticator/MissingIdentifierException.php
dereuromark: Good catch - 4.2.1 is already out, so this is 4.3.0. Fixed.

dereuromark: Correction: this is a bugfix, so next patch 4.2.2. Set to that.

